PR #209, opened 2019-08-13
Miovision zero pad tmc

rickl4:


radumas:
This should go into `automate_miovision_aggregation` first.

PR #214, opened 2019-08-20
Here api

rickl4:
Should this be merged with master as well?


radumas:
Actually what's in master is more up to date.

PR #215, opened 2019-08-20
Wys api

rickl4:


wys/api/python/wys_api.py
```diff
@@ -0,0 +1,351 @@
...
+                        WHERE B.address IS NULL
+                        
+                        UNION 
+                        
+                        SELECT a.api_id, a.address, a.sign_name, a.dir, start_date, lat, lng FROM daily_intersections A
+                        LEFT JOIN locations B ON A.api_id = B.api_id
+                        AND B.lat = A.lat
+                        WHERE B.lat IS NULL
+                        
+                        SELECT a.api_id, a.address, a.sign_name, a.dir, start_date, lat, lng FROM daily_intersections A
```
radumas: This seems duplicated and also there's no UNION between this SELECT and the one before it.

wys/api/python/wys_api.py
```diff
@@ -0,0 +1,351 @@
...
+                else:
+                    direction=None
+                temp=[location, address, name, direction, start_date, lat, long]
+                loc_table.append(temp)
+                signs_iterator=signs_list
+                
+            execute_values(cur, 'INSERT INTO daily_intersections (api_id, address, sign_name, dir, start_date, lat, lng) VALUES %s', loc_table)
+            string="""
+                        WITH locations AS (
+                        SELECT api_id, address, sign_name, dir, lat, lng, max(start_date)
```
radumas: max(start_date) will be labelled as `max`, but that's not called anywhere else in this SQL. should this be `DISTINCT` INSTEAD?

PR #219, opened 2019-08-21
Miovision api

rickl4:


radumas:
We should update the `volumes/README.md` to reflect miovision data being a part of our volume data

radumas:
I realized that setting up partitioned tables was something that we should be doing for the API data.

Issue #223, opened 2019-09-10
Automate update of Big Data RDS from Red Light Camera backend


KatiRG:
Query to replicate the relevant signal asset info from Red Light Camera database into our database. Should be stored in a permanent schema (via some Airflow process).

KatiRG:
@aharpalaniTO Do you have the query used to make the RLC table?

KatiRG:
@aharpalaniTO I have this process working and dumping to test table `vz_safety_programs_staging.rlc_test`. Can you check this table and let me know when I can write to the real `rlc` table?

aharpalaniTO:
The result sets are identical. Query below returns 0 rows.

```
(TABLE vz_safety_programs_staging.rlc_test EXCEPT TABLE vz_safety_programs_staging.rlc)
UNION ALL
(TABLE vz_safety_programs_staging.rlc_test EXCEPT TABLE vz_safety_programs_staging.rlc)
```

Is the code for the process ingesting data from the RLC feed to the RDS on Github?

KatiRG:
branch `airflow`, `dag` directory that calls tasks in `assets` directory

Issue #221, opened 2019-08-28
Automate update of Big Data RDS from Signal View backend


KatiRG:
Set up using airflow.

Store Signal View datasets in new folder `assets`.  

Store airflow dag files in new folder `dags`.

(This issue was originally in `bdit_vz_programs/issues/7`).

KatiRG:
@aharpalaniTO I have created a VIEW in my local RDS:
```
CREATE VIEW signals_cart AS
	SELECT 
	'Traffic Signals' AS ASSET_TYPE,
	a.ID AS PX,
	a.STREETNAME AS MAIN_STREET,
	a.MidblockRoute AS MIDBLOCK_ROUTE,
	a.SIDE1ROUTEF AS SIDE1_STREET,
	a.SIDE2ROUTE AS SIDE2_STREET,
	b.LATITUDE,
	b.LONGITUDE,
	b.ACTIVATIONDATE AS ACTIVATION_DATE,
	NULL AS DETAILS
	FROM sgmaingeneral a
	INNER JOIN public.sgpxgenmaingeneral b ON a.SGMAINGENERALOID = b.SGMAINGENERALOID
	WHERE a.ID::int < 5000 AND b.ACTIVATIONDATE IS NOT NULL AND b.REMOVALDATE IS NULL
	UNION ALL
	SELECT 
	'Pedestrian Crossovers' AS ASSET_TYPE,
	a.ID AS PX,
	a.STREETNAME AS MAIN_STREET,
	a.MidblockRoute AS MIDBLOCK_ROUTE,
	a.SIDE1ROUTEF AS SIDE1_STREET,
	a.SIDE2ROUTE AS SIDE2_STREET,
	b.LATITUDE,
	b.LONGITUDE,
	b.ACTIVATIONDATE AS ACTIVATION_DATE,
	NULL AS DETAILS
	FROM public.sgmaingeneral a
	INNER JOIN public.sgpxgenmaingeneral b ON a.SGMAINGENERALOID = a.SGMAINGENERALOID
	WHERE a.ID::int >= 5000 AND a.ID::int < 7000 AND b.ACTIVATIONDATE IS NOT NULL AND b.REMOVALDATE IS NULL
	UNION ALL
	SELECT
	'Audible Pedestrian Signals' AS ASSET_TYPE,
	a.ID AS PX,
	a.STREETNAME AS MAIN_STREET,
	a.MidblockRoute AS MIDBLOCK_ROUTE,
[...]

PR #226, opened 2019-09-25
Airflow

KatiRG:

- Pipeline for pulling traffic signal data from Oracle database and storing them into pre-existing view called `signals_cart` in local PostgreSQL server 

## Issue(s) this solves:
- #223  
- #221  

- two python scripts in `dags` folder
- postgres queries in bash scripts stored in `assets` folder that are called by the python scripts


dags/rlc.py
```diff
@@ -0,0 +1,117 @@
...
+
+  local_table='vz_safety_programs_staging.rlc'
+
+  url = "https://secure.toronto.ca/opendata/cart/red_light_cameras.json"
+  return_json = requests.get(url).json()
+  rows = [list(feature.values()) for feature in return_json]
+  insert = 'INSERT INTO {0} VALUES %s'.format(local_table)
+  with conn:
+    with conn.cursor() as cur:
+      execute_values(cur, insert, rows)
```
radumas, line 85: this assumes variables are in a consistent order in json, which is incorrect and why this DAG was failing

radumas (changes requested):
highlighting the bug in the RLC dag, haven't reviewed the rest of this PR

Issue #238, opened 2019-11-21
Zero filling SQL in `aggregate_15_min_tmc()` is not filling in zero volume for missing datetime_bin


jovenc:
> The `aggregate_15_min_tmc()` function performs zero-filling by cross-joining a table containing all possible movements `intersection_movements` to create a table with all possible times and movements. Through a join with volumes, the query checks if there are no counts at that time, intersection, movement, and classification the is no volume, and fills the gap with a 0-volume bin.

Above is what the function should do based on README.md but right now it is only creating zero volume where datetime_bin exist in the table after 15min bin aggregation and attempt to insert zero into the table. Hence, there isn't really any zero volume 15min bin created and `miovision_api.volumes_tmc_zeroes` table which should have a record of any zero volume bin created is empty.

NOTE: Zero filling is only applicable to rows where classification_uid IN (1, 2, 6). Looking solely at 3 days worth of data, zero filling is only really necessary for about 1 to 5 datetime_bin for classification_uid =2 each day.

radumas:
The problem is that it's using the `datetime_bin` values from the `bin` temporary table, but those depend on whether there was sufficient data for each 15-minute bin for each intersection. The solution I think is to do an actual `CROSS JOIN`, replacing this code 

https://github.com/CityofToronto/bdit_data-sources/blob/miovision_api_bugfix/volumes/miovision/sql/function-aggregate-volumes_15min_tmc.sql#L86-L92

with 

```sql
                SELECT m.*, datetime_bin 
		FROM miovision_api.intersection_movements m
		CROSS JOIN generate_series(start_date, end_date - interval '15 minutes', INTERVAL '15 minutes') AS dt(datetime_bin)
		WHERE classification_uid IN (1,2,6) ;
```

jovenc:
I tried adapting the above method and using the query below but it is still not filling up missing datetime_bin for me. I played around with different kind of joins (FULL OUTER JOIN, LEFT OUTER JOIN) but they all gave me the same result?

```
WITH zero_padding_movements AS (
		/*Cross product of legal movement for cars, bikes, and peds and the bins to aggregate*/
		SELECT m.*, datetime_bin 
		FROM miovision_api.intersection_movements m
		CROSS JOIN generate_series(start_date - interval '1 hour', end_date - interval '1 hour 15 minutes', INTERVAL '15 minutes') AS dt(datetime_bin)
		WHERE classification_uid IN (1,2,6) 
		)
[...]

Issue #237, opened 2019-11-20
volumes_15min_tmc is missing data from 23:15-23:45


radumas:
Raised here https://github.com/CityofToronto/bdit_data-sources/issues/233#issuecomment-555115410

These should be filled in when processing the subsequent day, but don't. Potentially the fix to #220 might have over-corrected somehow, but nothing obvious is jumping out from the SQL script.

Issue #236, opened 2019-11-20
Optimize Miovision API Materialized Views


radumas:
They're really slow to refresh.

Maybe they should work some other way...

Issue #233, opened 2019-11-04
Miovision API Puller bug fix


jovenc:
In `miovision_api_bugfix` branch, the script fixed can be found at `volumes/miovision/api/intersection_tmc.py`

## Issues solved/changes made so far:

- #220 resolved, no more duplicated 15min data
- Changed start_date to start_time where necessary to avoid confusion
- Edited `pull_date` function to pull the desired intersection and for clarity
- Edited `process_data` function
- Separated `refresh_views` into a new function
- Created/moved around the `try...except:` block for better clarity
- Parameterized query

## Remaining issues:

- [x] #222 still exists but the error only pops out a step before the script is done aka after all the pulling, inserting and processing data is done. Not sure what is wrong. The script works fine before today.

```
(api) jchew@ip-10-160-2-198:~/miovision_api/volumes/miovision/api$ pipenv run python intersection_tmc.py run-api --intersection 5
04 Nov 2019 11:02:45            DEBUG    Start
04 Nov 2019 11:02:45            DEBUG    Connected to DB
04 Nov 2019 11:02:45            INFO    Pulling from 2019-11-03 00:00:00-04:00 to 2019-11-04 00:00:00-05:00
04 Nov 2019 11:02:45            DEBUG    [(5, 'f4cfdbdc-afea-4a8a-adc4-bb64a3e8e4af', 'Front / Bathurst', Decimal('43.640752'), Decimal('-79.401359'), 'Front', 'Bathurst', 13468126, 297, '0101000020E610000023BDA8DDAFD953C05D6C5A2904D24540', None, None, None, True)]
04 Nov 2019 11:02:45            INFO    Front / Bathurst     2019-11-03 00:00:00-04:00
04 Nov 2019 11:02:51            INFO    Completed data pulling for 2019-11-03 00:00:00-04:00
04 Nov 2019 11:03:03            INFO    Inserted into volumes and updated log
04 Nov 2019 11:03:13            INFO    NOTICE:  Invalid movements more than 1000, QC check raw data

04 Nov 2019 11:03:34            INFO    Aggregated to 15 minute bins
04 Nov 2019 11:03:34            INFO    Completed data processing for 2019-11-03 00:00:00-04:00
04 Nov 2019 11:04:05            INFO    missing_dates_query done
04 Nov 2019 11:04:05            INFO    report_dates done
04 Nov 2019 11:04:05            INFO    Front / Bathurst     2019-11-04 00:00:00-04:00
[...]

PR #239, opened 2019-11-22
Miovision api bugfix

radumas:

- Fixes data bugs in aggregating the data
- Improves the Python
- Removes outdated Materialized Views for King Street Pilot

## Issue(s) this solves:

- #219 
- #222 
- #237 
- #220 
- #238 
- #233 
- #239 
- #236 

 @jovenc :

- The `datetime_bin` for pulling into the tables `miovision_api.volumes` , `miovision_api.volumes_15min_tmc` and `miovision_api.volumes_15min` was modified to account for the definition of one day data in this case (23:00 previous day to 23:00 current day). Also made sure the datetime_bin does not overlap.
- Generated series so that zero volumes can be added to 'missing' datetime_bin without any volume data.
- Moved refresh materialized view into a new function named `refresh_views` which has not been tested.
- Fixed Json error by fixing/decoding the format of the error message returned when an error occurs.
- Did some reorganizing/renaming and added comments for clarification purposes.
- Unique constraints added to relevant tables.
- Parameterized query.
- Updated Miovision README with figures to explain TMC & ATR as well as the `datetime_bin` for a day.


volumes/miovision/api/intersection_tmc.py
```diff
@@ -183,157 +188,139 @@ def get_pedestrian(table, start_date, end_iteration_time, intersection_id1, inte
...
 
+        end_iteration_time+=time_delta
+        start_time+=time_delta
+        if start_time>=end_time:
+            break
+    logger.info('Done')
+        
 if __name__ == '__main__':
     cli()
+    run_api(start_date, end_date, path, intersection, pull)
```
radumas: This line should be unnecessary

Issue #253, opened 2020-01-07
Migrate here_api to run on airflow


radumas:

radumas:
Got an error when trying to export the RDS connection as an environment variable.

```
[2020-01-07 14:51:26,043] {bash_operator.py:124} INFO - Output:
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO - Traceback (most recent call last):
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -   File "/etc/airflow/data_scripts/here/traffic/here_api.py", line 214, in <module>
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -     main()
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -   File "/etc/airflow/data_scripts/here/traffic/here_api.py", line 211, in main
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -     cli(obj={})
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -   File "/etc/airflow/data_scripts/.venv/lib/python3.5/site-packages/click/core.py", line 764, in __call__
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -     return self.main(*args, **kwargs)
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -   File "/etc/airflow/data_scripts/.venv/lib/python3.5/site-packages/click/core.py", line 696, in main
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -     _verify_python3_env()
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -   File "/etc/airflow/data_scripts/.venv/lib/python3.5/site-packages/click/_unicodefun.py", line 124, in _verify_python3_env
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO -     ' mitigation steps.' + extra
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO - RuntimeError: Click will abort further execution because Python 3 was configured to use ASCII as encoding for the environment. Consult https://click.palletsprojects.com/en/7.x/python3/ for mitigation steps.
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO - 
[2020-01-07 14:51:26,642] {bash_operator.py:128} INFO - This system supports the C.UTF-8 locale which is recommended.
[...]

PR #269, opened 2020-02-28
Here airflow

radumas:

- Creates a here dag to automate pulling here data
- Modifies the here_api utility to not rely on sending emails and get credentials from the environment. 

## Issue(s) this solves:

- #253 

- Nothing


radumas:
Already merged in #258

Issue #301, opened 2020-05-25
CurbTO — Automate pulling in installed interventions


chmnata:
- [ ] Modify gccview script to pull from external page

- [ ] Create airflow dag to run every evening

PR #313, opened 2020-07-30
Allow null geometry to pass through

chmnata:

- Allow features with no geometry to be inserted

## Issue(s) this solves:

- #301


gis/gccview/pull_interventions.py
```diff
@@ -77,7 +77,10 @@ def send_tempdata(output_table, insert_column, return_json):
     fields = return_json['fields']
     trials = [[field['name'], field['type']] for field in fields]
     for feature in features:
-        geom = feature['geometry']
+        try:
+            geom = feature['geometry']
```
radumas, line 81: `geom = feature.get('geometry')` is a simpler solution

chmnata: changed to get

Issue #315, opened 2020-08-04
Google Business Data: Pull popular times and wait times data


radumas:
Commit code to the `google-business` branch tagging this issue in commit messages with `#[issuenumber]` and then open a pull request

PR #321, opened 2020-08-20
Google business

webgisgeek:

- let Raph to review
- 

## Issue(s) this solves:

- #315

-


radumas (changes requested):
Some comments on improving Python and Markdown.

Per our conversation there is outstanding analysis of the data to perform but it is not a priority at the moment.

Issue #330, opened 2020-09-21
Miovision api script can fail silently


radumas:
There needs to be a `sys.exit(1)` in this exception block so that Airflow will realize that the script has not exited successfully. Right now this block catches any exception and makes it seem like the program ran successfully, which is wrong.

https://github.com/CityofToronto/bdit_data-sources/blob/a22e407b09534dc01da99e15893725c2917f0248/volumes/miovision/api/intersection_tmc.py#L84-L85

For example, log output when changing permission on the `miovision_api` schema meant the script failed early.

```
[2020-09-19 03:00:13,250] {bash_operator.py:124} INFO - Output:
[2020-09-19 03:00:13,526] {bash_operator.py:128} INFO - 72    	INFO    Pulling from 2020-09-18 00:00:00-04:00 to 2020-09-19 00:00:00-04:00
[2020-09-19 03:00:13,527] {bash_operator.py:128} INFO - 77    	CRITICAL    Traceback (most recent call last):
[2020-09-19 03:00:13,527] {bash_operator.py:128} INFO -   File "/etc/airflow/dev_scripts/volumes/miovision/api/intersection_tmc.py", line 75, in run_api
[2020-09-19 03:00:13,527] {bash_operator.py:128} INFO -     pull_data(conn, start_time, end_time, intersection, path, pull, key, dupes)
[2020-09-19 03:00:13,527] {bash_operator.py:128} INFO -   File "/etc/airflow/dev_scripts/volumes/miovision/api/intersection_tmc.py", line 287, in pull_data
[2020-09-19 03:00:13,527] {bash_operator.py:128} INFO -     cur.execute(string2, (start_time,))
[2020-09-19 03:00:13,527] {bash_operator.py:128} INFO - psycopg2.errors.InsufficientPrivilege: permission denied for schema miovision_api
[2020-09-19 03:00:13,527] {bash_operator.py:128} INFO - LINE 1: SELECT * FROM miovision_api.intersections
[2020-09-19 03:00:13,527] {bash_operator.py:128} INFO -                       ^
[2020-09-19 03:00:13,527] {bash_operator.py:128} INFO - 
[2020-09-19 03:00:13,528] {bash_operator.py:128} INFO - 
[2020-09-19 03:00:13,566] {bash_operator.py:132} INFO - Command exited with return code 0
```

cczhu:
[...]

Issue #329, opened 2020-09-16
Update Miovision API script to deal with classification "MotorizedVehicle"


radumas:
This is not being dealt with in the current Python script and results in `classification_uid = 0`.

https://github.com/CityofToronto/bdit_data-sources/blob/a22e407b09534dc01da99e15893725c2917f0248/volumes/miovision/api/intersection_tmc.py#L121-L136

- [x] So a new classification needs to be created in the script and in the `classification` table

- [x] All records after 2019-11-07 with `classification_uid = 0` should get updated to the new number

- [x] Honestly the script should fail if the classification is unknown 

- [ ] New `intersection_movements` need to be created for the new classification, this could be done similarly to point 2 under [Adding Intersections](https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/miovision#adding-intersections)

cczhu:
Should we raise `ValueError` rather than `return '0'` in `get_classification` to solve that third point?

radumas:
Ya that seems like an elegant solution

radumas:
Added created new `intersection_movements` to task list

cczhu:
> Added created new `intersection_movements` to task list

The link currently goes to the TPS challenge 😹

radumas:
lol I need to actually copy the URL from the address bar, not just click on the 🔗 button

cczhu:
Cool, sounds like I can do

```sql
INSERT INTO miovision_api.intersection_movements
      ( intersection_uid, classification_uid, leg, movement_uid )
SELECT intersection_uid,
       '9'::integer AS classification_uid,
       leg,
	   movement_uid
FROM miovision_api.intersection_movements 
WHERE classification_uid = 1
```

to update the valid movements. Is it an issue that `miovision_api.intersection_movements` will no longer be sorted?

Issue #323, opened 2020-08-26
Miovision - Light Autos - late night data drops


aharpalaniTO:
There appears to be an issue in which data for light autos appears to be truncated for certain intersections and days. Specifically, there appears to be data for cyclists, pedestrians, etc., but light auto volumes are erroneously zero.  While for some intersections this doesn't appear to be a problem at all, for other it's fairly significant (e.g. intersection_uid = 2).

High-level identification of the problem:
```
-- expectation is that if things are normal, this should return next to no results (with possible exception of road closures where peds are still permitted)
WITH ind AS (
SELECT intersection_uid, datetime_bin
FROM miovision_api.volumes_15min
WHERE EXTRACT(HOUR FROM datetime_bin) >= 22
GROUP BY intersection_uid, datetime_bin
HAVING SUM(CASE WHEN classification_uid = 1 THEN volume ELSE NULL END) = 0 AND SUM(volume) > 0 -- where light volumes are 0 in any 15-minute bin, but other modes have volumes
	)
	
	SELECT intersection_uid, intersection_name, date_trunc('month', datetime_bin) AS mth, COUNT(1) as cnt -- maximum value here is ~360 (4 bins/hr * 3 hrs * 30 days)
	FROM ind
	INNER JOIN miovision_api.intersections USING (intersection_uid)
	GROUP BY intersection_uid, intersection_name, date_trunc('month', datetime_bin)
	ORDER BY intersection_uid, date_trunc('month', datetime_bin)
```

Crude confirmation that this issue also exists in `miovision_api.volumes`:
```
-- expectation is that if things are normal, the distribution should be entirely max_hr = 23, with a few scattered exceptions for when cameras go down midday

WITH ind AS (
SELECT intersection_uid, datetime_bin::date AS dt, MAX(EXTRACT(HOUR FROM datetime_bin)) AS max_hr
FROM miovision_api.volumes
WHERE classification_uid = 1
	GROUP BY intersection_uid, datetime_bin::date
)

SELECT intersection_uid, max_hr, COUNT(1)
FROM ind
GROUP BY intersection_uid, max_hr
ORDER BY intersection_uid, max_hr
```

Possible causes:
1, something weird is happening in transform from raw data pull to `miovision_api.volumes`
2. API script is hitting some sort of max row constraint (similar to an issue we had with Blip)
3. Issue on Miovision's end (the API doesn't have all the data available)

radumas:
Should've spotted that earlier
[...]

PR #332, opened 2020-09-21
Update Miovision API pull script to adapt to recent changes

cczhu:

- Changes the time interval used in sequential API calls in `intersection_tmc.py` to 6 hours to avoid late data gaps.
- Have `intersection_tmc.py` exit with code 1 if any exceptions is encountered when pulling raw data from the API with `pull_data`.
- Assign vehicle class `9` to `MotorizedVehicle`. Have `get_classification` raise an exception if it encounters a class that hasn't been defined in the function.
- Remove a number of trailing whitespaces.
- Adds development notebooks that confirms the use of a 6-hour time interval for API pulls, and checks that we can successfully import raw data into Postgres.

## Issue(s) this solves:

- #323
- #329
- #330

- Examination of the changes to `intersection_tmc.py` to make sure I haven't made any careless mistakes.


volumes/miovision/api/intersection_tmc.py
```diff
@@ -327,23 +340,20 @@ def pull_data(conn, start_time, end_time, intersection, path, pull, key, dupes):
...
         except psycopg2.Error as exc:
             logger.exception(exc)
             sys.exit(1)
-        
+
         if pull_data:
             logger.info('Skipping aggregating and processing volume data')
         else:
-            process_data(conn, start_time, end_iteration_time)
+            process_data(conn, c_start_t, c_end_t)
```
radumas, line 354: Per our discussion, `process_data()` should get moved out of the `time_delta` loop and take `start_time, end_time` as parameters

radumas (changes requested):
Nice changes to the README, whitespace, and great addition of notebooks.

One change wrt `process_data`

Also would be good if notebook folder was referenced in README, and I guess notebook folder should get its own README to briefly touch on what each notebook is about?

volumes/miovision/api/intersection_tmc.py
cczhu: Moved code block out of the loop. Haven't made the requisite changes to `miovision_api.refresh_gapsize_lookup`, but as discussed we can save that for another PR.

radumas (approved):
Awesome!

Issue #338, opened 2020-10-06
Bug in Miovision gap finding process for setting data to `NULL`


radumas:
`BETWEEN` is an inclusive operator.

`a BETWEEN x AND y`

is equivalent to

`a >= x AND a <= y`

https://github.com/CityofToronto/bdit_data-sources/blob/263464d1ffdeff12bb57a4aae1f6bfb044332402/volumes/miovision/sql/function-aggregate-volumes_15min_tmc.sql#L37-L38

cczhu:
Resolved by #344.

PR #344, opened 2020-10-14
Fix Miovision aggregation to properly use 1 hour pads for short gaps

cczhu:

- Fixes a bug in the Miovision data processing pipeline where the 15-minute aggregation script would pad out 1 hour 15 minutes around short gaps rather than 1 hour.

## Issue(s) this solves:

- #338

- Cursory glance at the code.


radumas (approved):
Sublime

PR #346, opened 2020-11-02
Fix Miovision API Puller Bugs

### LINKED ISSUE #339: Miovision API logging cosmetic issue
Right now the logs say `Completed data pulling from s to s` because somewhere in the script a timestamp is not being converted properly to a string:

image

This is not at all high priority but I should fix it.

### LINKED ISSUE #345: Port timebin fix to `master`
A potentially serious problem was fixed in [a commit ](https://github.com/CityofToronto/bdit_data-sources/commit/d9c042b18acf4ee52660aab2a69c03b7a05caa0a) in `miov_update_postproc`. Since we're still working on how to perform proper postprocessing an entire data refresh, that fix was never pushed to production.

In retrospect not committing that fix separately to master was a foolish thing to do. We should make a separate fix in master (this shouldn't result in any merge conflicts) so that we're not downloading an additional 6 hours of data every day.

In `intersection_tmc.py`, change:

```
    for i in range(math.ceil((end_time - start_time) / dt)):
```

to

```
    for i in range(math.ceil((end_time - start_time) / dt) - 1):
```

cczhu:

- Stop the API from pulling one extra bin of data beyond the endpoint of the datetime range (cherry-picked from [d9c042](https://github.com/CityofToronto/bdit_data-sources/commit/d9c042b18acf4ee52660aab2a69c03b7a05caa0a)).
- Fix a timestamp display issue in the logging.

## Issue(s) this solves:

- Closes #339 
- Closes #345

- Glance at the source code (and then push to productionization for tomorrow's run).

### DIFF
--- a/volumes/miovision/api/intersection_tmc.py
+++ b/volumes/miovision/api/intersection_tmc.py
@@ -370,8 +370,9 @@ def pull_data(conn, start_time, end_time, intersection, path, pull, key, dupes):
             table.extend(table_veh)
             table.extend(table_ped)
 
-        logger.info('Completed data pulling from {0:s} to {1:s}'
-                    .format(c_start_t, c_end_t))
+        logger.info('Completed data pulling from %s to %s'
+                    %(c_start_t, c_end_t))
+
         try:
             insert_data(conn, c_start_t, c_end_t, table, dupes)
         except psycopg2.Error as exc:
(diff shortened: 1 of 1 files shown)


volumes/miovision/api/intersection_tmc.py
```diff
@@ -370,8 +370,9 @@ def pull_data(conn, start_time, end_time, intersection, path, pull, key, dupes):
             table.extend(table_veh)
             table.extend(table_ped)
 
-        logger.info('Completed data pulling from {0:s} to {1:s}'
-                    .format(c_start_t, c_end_t))
+        logger.info('Completed data pulling from %s to %s'
```
radumas, line 373: FYI you can append variable to be formatted into the string without calling a formatting function
https://docs.python.org/3/howto/logging.html#logging-variable-data

radumas (approved):

Issue #348, opened 2020-11-04
Collisions data pipeline


cczhu:
Updated version of [bdit_vz_analysis/#8](https://github.com/Toronto-Big-Data-Innovation-Team/bdit_vz_analysis/issues/8).

We don't currently have a collisions.acc table that's refreshed on a regular basis from the MOVE server. This will be required for network screening.

Tasks:

Tasks:
- [x] set up a new data source, `collisions`.
- [x] set up a script that copies ACC.dat from Flashcrow to our Postgres. We can't run an Airflow process yet because Flashcrow and the EC2 are in different AWS regions.
- [x] augment the process with some basic checks that the new data isn't corrupted or whatnot (eg. number of NaNs in a column doesn't exceed some fraction, number of rows doesn't decrease below some threshold, etc.)
- [x] document the dataset.

cczhu:
Self-consistency checks:

- Ensure that `wc ACC.dat` returns the same number of rows
- Ensure that the number of unique `accnb` and `accyear` combinations in `collisions.acc` is the same as the number of `collision_no`.
- Ensure that the number of rows in `collisions.events` is the same as the number in in `collisions.collision_no`.
- Ensure that the number of unique `collision_no` in `collisions.involved` is the same as the number of lines in `collisions.collision_no`.
- Ensure there are no NaNs in `collision_no` in either `collisions.events` or `collisions.involved`.

cczhu:
First draft complete - launching a PR.

Issue #347, opened 2020-11-03
Get rid of unnecessary legacy collision tables


cczhu:
As part of standardizing the collision data pipeline, we need to get rid of all the deprecated collisions tables still in the Postgres (this should also free up a few GB of space, which is quite welcome). Unfortunately they have many dependencies that we'll have to handle.

**czhu.involved2019**
- aharpal.ksi_2018
- czhu.collision_events_for_ase
- czhu.collision_events_safetyzones_forase_20
- czhu.collision_events_safetyzones_forase_50
- czhu.collision_events_srz_old
- czhu.collision_events_for_tps
- czhu.temp_collision_events_srz

**traffic.acc**
?

**collisions.acc_20200304**
- czhu.collision_no
- czhu.collision_events
- czhu.collision_involved
- czhu.temp_collision_events_for_ase
- czhu.temp_collision_events_safetyzones_forase
- vz_analysis.mbkscreen_20200504_midblkonly
- vz_analysis.mbkscreen_20200219_pedcollision_test 
- vz_analysis.mbs20200701_collisions 
- czhu.dv_ksi 
- natalie.neighbourhood_prop_ksi 
- natalie.atp_neighbourhood_score
- natalie.neighbourhood_prop_ksi
- rsaunders.ksi15

**collisions.acc_20200116**
- czhu.collision_no_torstar 
- czhu.collision_involved_torstar
- czhu.collision_events_torstar
- vz_analysis.mbkscreen_20200219_pedcollision
- vz_analysis.mbkscreen_20200219_pedcollision_mblkonly
- collision_events_torstar
- rsaunders.ksitruckdata
- rsaunders.ksitrkpedcyc

Tasks:
- [x] Contact Aakash, Natalie and Racheal to ask them to point their materialized views to the most recent `collisions.acc` or their derived tables.
- [x] Double-check with Aakash that all remaining copies should be kept.
- [x] Merge [PR #32](https://github.com/Toronto-Big-Data-Innovation-Team/bdit_data_requests/pull/32) of `bdit_data_requests` (now a year old!) and delete TPS and ASE materialized views.
- [x] Figure out which legacy and test `vz_analysis` tables should be deleted.
- [x] Delete the tables above that no longer have dependencies.

cczhu:
Deleted legacy `events` and `involved` tables in `collisions`. Archived child materialized views in `vz_challenge` before deleting.

cczhu:
[...]

PR #349, opened 2020-11-05
Add first draft of collisions pipeline (#348).

cczhu:

- Adds table and view definitions for collisions dataset on Postgres.
- Adds a Python script, `pull_acc_script.py`, that manually retrieves `ACC.dat` from the MOVE server and uploads it to Postgres. This will become deprecated once the RDS gets moved to Canada East, but most of the Postgres components of the script should still be useful.
- Cleaned RDS of old versions of ACC, and added legacy suffixes to those tables that still have parent views in production (see #347 for progress).

## Issue(s) this solves:

- #348

- Check of the SQL scripts to make sure I'm not doing anything silly.
- Check resulting table structures on Postgres (under `collisions` schema) to ensure everything makes sense.
- Quick read of Readme.md to make sure I'm not missing anything.

- Should we get rid of `collision_factors.collision_id` and `collision_factors.collision_id_date`?

### DIFF
--- a/collisions/GettingStarted.md
+++ b/collisions/GettingStarted.md
@@ -1,0 +1,292 @@
+# Getting Started with Collision Data
+
+Welcome to the tutorial on query collision data! This tutorial assumes you are
+familiar with Postgres and PostGIS. If you'd like resources on those, please see
+the Data & Analytics Postgres onboarding documentation [here](https://www.notion.so/bditto/PostgreSQL-Exercises-322493ab085b442f96bfdb77b039cfca).
+
+## Layout
+
+`collisions.acc` is a direct mirror of the table on the MOVE server, and has not
+been processed to eg. convert category codes into text. It could be used for
+analysis, but is typically only examined for diagnostic purposes.
+
+`collisions.events` is a table of all data related to the collision event, such
+as its location and time of day, as opposed to data related to involved
+individuals, such as injury or manoeuver. `collisions.involved` includes all
+involved individual data - "involved data" for short.
+
+## Joining Events and Involved
+
+THe `events` and `involved` tables are often joined to get a single table, where
+each row represents an individual involved in a crash, but will contain data at
+both event and involved levels.
+
+To join the two tables together, we use
+
+```sql
+SELECT *
+FROM collisions.events
+LEFT JOIN collisions.involved USING (collision_no);
+```
+
+This will return *every* record available, so typically we also append a `WHERE`
+clause to restrict the query by time and place. This query is popular enough
+that it has been turned into a view, `collisions.alldata`. For example, to
+return all involved from January 2015, one could write:
+
+```sql
+SELECT *
+FROM collisions.alldata
+WHERE accdate >= '2015-01-01' AND accdate < '2015-02-01';
+```
+
+## Counting Events and Involved
+
+One common query is for the total number of collision events, or the total
+number of individuals involved in collisions, aggregated to the month, or year.
+If we, for example, wanted the number of collision events and involved from
+2015-2019 inclusive, we'd do:
+
+```sql
+SELECT accyear,
+       -- Number of collision **events**.
+       COUNT(DISTINCT collision_no) n_collisions,
+       -- Number of people **involved**.
+       COUNT(*) n_involved
+FROM collisions.alldata
+WHERE accyear BETWEEN 2015 AND 2019
+GROUP BY accyear
+ORDER BY 1;
+```
+
+As of 2021-07-05, the output looks like:
+
+accyear | n_collisions | n_involved
+-- | -- | --
+2015 | 50865 | 124226
+2016 | 55637 | 108283
+2017 | 58917 | 105669
+2018 | 62401 | 108103
+2019 | 64367 | 108316
+
+but be aware that due to the ever-refreshing nature of collisions mentioned in
+`Readme.md`, these numbers will surely change by small amounts with time.
+
+The `data_source` column lists whether the collision event comes from TPS or CRC
+(or cannot be deduced from the `ACCNB`). `validation_userid` gives the name of
+Data & Analytics staff member who most recently validated the involved
+(sometimes only some individals from a collision event are validated). If we
+want the number of events, involved and validated involved from the data
+subdivided by year and data source, we'd do:
+
+```sql
+SELECT accyear,
+       data_source,
+       -- Number of collision **events**.
+       COUNT(DISTINCT collision_no) n_collisions,
+       -- Number of people **involved**.
+       COUNT(*) n_involved,
+       -- Number of people **involved** whose data has been validated.
+	   COUNT(*) FILTER (WHERE validation_userid IS NOT NULL) n_valid_involved
+FROM collisions.alldata
+WHERE accyear BETWEEN 2015 AND 2019
+GROUP BY accyear, data_source
+ORDER BY 1, 2;
+```
+
+As of 2021-07-05, the output looks like:
+
+accyear | data_source | n_collisions | n_involved | n_valid_involved
+-- | -- | -- | -- | --
+2015 | CRC | 36059 | 89329 | 36490
+2015 | TPS | 14806 | 34897 | 28899
+2016 | CRC | 44106 | 80651 | 31763
+2016 | TPS | 11507 | 27601 | 22301
+2016 | NULL | 24 | 31 | 0
+2017 | CRC | 48978 | 81514 | 23135
+2017 | TPS | 9923 | 24127 | 20272
+2017 | NULL | 16 | 28 | 2
+2018 | CRC | 52697 | 84471 | 18595
+2018 | TPS | 9692 | 23611 | 17641
+2018 | NULL | 12 | 21 | 5
+2019 | CRC | 55132 | 85975 | 10663
+2019 | TPS | 9203 | 22291 | 5317
+2019 | NULL | 32 | 50 | 7
+
+We see that the majority of collisions come from the CRC (since they handle
+minor collisions). Collisions from TPS are much more likely to be validated
+(because they're much more likely to involved killed or seriously injured
+individuals, which are prioritized for validation). The number of validated
+collisions goes down with year (since it takes time to validate).
+
+## Summing Involved
+
+Data request clients will often ask for data aggregated up to the collision
+event level, but also ask for involved-level data. One way to satisfy these
+requests (though be sure to confirm with the client that this is acceptable to
+them!) is to sum up the number of involved of a particular category (see eg.
+[this data
+request](https://github.com/Toronto-Big-Data-Innovation-Team/bdit_data_requests/pull/140/)).
+If we, for example, wanted the number of killed or seriously injured (KSI) per
+collisions on 2015 involving at least one KSI, and at least one person under the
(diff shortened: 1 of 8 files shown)


cczhu:
Oooo, shell injection issues. @radumas is there a better way to run shell commands from a Python script that also requires user input? Is there a sensible way to sanitize user inputs (eg. remove pipes).

(Do we care, since this'll only be used by me for a few months?)

radumas:
I used subprocess. Dunno if CodeFactor would scream about that
https://github.com/CityofToronto/bdit_data-sources/blob/master/here/traffic/here_api.py#L163-L187

cczhu:
The comment to `collisions.acc` currently says

> Raw collision database, from Flashcrow /data/replicator/flashcrow-CRASH/dat/ACC.dat. Refreshed on %s.

Fix this Charles 😠

cczhu:
Jesse's query for determining which collisions are from the CRC and which from TPS MVA:

```sql
SELECT EXTRACT(YEAR FROM "ACCDATE") accyear,
       EXTRACT(MONTH FROM "ACCDATE") accmonth,
	   COUNT(DISTINCT "ACCNB") n_collisions,
	   COUNT(DISTINCT "ACCNB") FILTER (WHERE LENGTH("ACCNB") = 10) n_collisions_tps,
	   COUNT(DISTINCT "ACCNB") FILTER (WHERE LENGTH("ACCNB") = 9) n_collisions_crc,
	   COUNT(DISTINCT "ACCNB") FILTER (WHERE "INJURY" IN ('3', '4')) n_ksi
FROM collisions.acc
WHERE "ACCDATE" >= '2019-01-01' AND "ACCDATE" < '2021-03-19'
GROUP BY EXTRACT(YEAR FROM "ACCDATE"), EXTRACT(MONTH FROM "ACCDATE")
ORDER BY 1, 2
```

cczhu:
Also this is from Jim Millington (e-mail to Jesse in the week of 2021-03-15), and should be documented:

> Based on the convention (and the size limit of the field) the accnb can only be 10 digits long. From police we have gotten 9 digit accnbs in the past, which leaves only 1 digit for the year from police. They have so many because all offences in the Ontario law book are in the same database, including murder. We are filtered to get only the Toronto traffic records. So all TPS records for 2021 will start with 1. (and this is why you need to consider the year along with accnb to make a unique collision come 2023, since the scheme was created in 2013. I warned Jim S, but he said we didn't care because he would be retired by then.)
The ones that start 10xxxxxxxx will eventually have numbers like 18xxxxxxxx
Also based on the convention, CRCs have a range assigned to them at the beginning of the year for their records alone that number 8000000, so they end up being 9 digits long starting with a 2 digit year, I would expect all CRC records for 2021 to start with 218.
And remember that the acc_archive table contains records that couldn't load into acc because of missing cords or out of bounds or private property

cczhu:
To do:

- [ ] `impact_type` is an event level variable, I think. Double-check this and move it as needed.
- [ ] Include a `px` variable.
- [ ] Scan through [ACC Legacy Schema](https://www.notion.so/bditto/5cf7a4b3ee7d40de8557ac77c1cd2e69?v=56499d5d41e04f2097750ca3267f44bc) and see if there's any useful columns not already in `events` and `involved`.
- [ ] Figure out a way of identifying CRC vs. TPS data sources (see comment above).

cczhu:
Also fix this in `collision_factors.vehtype`:

image

cczhu:
@radumas points out we should also have in `collisions.events` a geometry table:

```
ST_SetSRID(ST_MakePoint(longitude, latitude), 4326)
```

and an associated spatial index.

cczhu:
`impact_type` is an event-level value, but is currently stuck in `involved`. We should move it.

cczhu:
In a meeting with the MOVE team ([notes](https://www.notion.so/bditto/MOVE-x-BDITTO-Collision-Filtering-Standard-23dbb390f9804206842816596dca9327)) it was revealed that ACC's lon-lat values are in [SRID 4269](https://epsg.io/4269)! This probably leads to very minor offsets in the collision locations, but we should implement that to be consistent with the MOVE platform (which converts these lon-lats to 4326).

cczhu:
Can we make a view of `collisions.events` joined with `collision.involved` which runs really fast if you filter further, or does the view act like a CTE?

cczhu:
Performed a quick check to see what error we obtain by setting ACC long and lat as EPSG:4326 rather than EPSG:4269, as mentioned above:

```sql
WITH step_1 AS (
	-- Produce WGS84 geoms directly from lat-lon, and WGS84 geoms from WGS83 geoms.
	SELECT ST_SetSRID(ST_MakePoint("LONGITUDE", "LATITUDE"), 4326) geom_wgs84,
	       ST_TRANSFORM(ST_SetSRID(ST_MakePoint("LONGITUDE", "LATITUDE"), 4269), 4326) geom_wgs83conv
	FROM collisions.acc
-- Extract longitude and latitude from each.
), step_2 AS (
	SELECT ST_X(geom_wgs84) long_wgs84,
		   ST_Y(geom_wgs84) lat_wgs84,
		   ST_X(geom_wgs83conv) long_wgs83conv,
		   ST_Y(geom_wgs83conv) lat_wgs83conv
	FROM step_1
)
SELECT *
FROM step_2
WHERE (2 * ABS(long_wgs84 - long_wgs83conv) / (long_wgs84 + long_wgs83conv + 1e-30) > 1e-14)
	OR (2 * ABS(lat_wgs84 - lat_wgs83conv) / (lat_wgs84 + lat_wgs83conv + 1e-30) > 1e-14);
```

This returns no rows, indicating that we can simply use the raw (offset) long and lat without first converting to 4269.

cczhu:
Collision pipeline is *finally* working again 🎉 

image

cczhu:
Some basic QCing:

## Collision Data Sources

```sql
WITH step_1 AS (
	SELECT "ACCNB" accnb,
	       "ACCDATE" accdate,
		   CASE
			   WHEN "ACCNB"::bigint >= 1000000000 THEN 'TPS'
			   WHEN "ACCNB"::bigint >= 100000000 THEN 'CRC'
			   ELSE NULL
		   END data_source
	FROM collisions.acc
	WHERE "ACCDATE" BETWEEN '2019-01-01' AND CURRENT_DATE
)
SELECT EXTRACT(year FROM accdate) accyear,
       data_source,
       COUNT(accnb) n_collisions
FROM step_1
GROUP BY data_source, EXTRACT(year FROM accdate)
ORDER BY 1, 2;
```

accyear | data_source | n_collisions
-- | -- | --
2019 | CRC | 83509
2019 | TPS | 22277
2019 | NULL | 50
2020 | CRC | 22523
2020 | TPS | 1
2020 | NULL | 15765
2021 | CRC | 86
2021 | TPS | 2359

That's a ton of NULL collisions in 2020. The reason for this is likely because `"ACCNB"::bigint` shaves off leading zeros, which we naturally get when we pass the decadal mark (i.e. ACCNBs for 2020 collisions will look like 0000xxxxxx). What if we correct for that?

```sql
WITH step_1 AS (
	SELECT "ACCNB" accnb,
	       "ACCDATE" accdate,
		   CASE
			   WHEN SUBSTRING("ACCNB", 1, 4) = '0000' AND NOT "ACCDATE" BETWEEN '2020-01-01' AND '2020-12-31' THEN 'Wut'
			   WHEN SUBSTRING("ACCNB", 1, 4) = '0000' AND "ACCDATE" BETWEEN '2020-01-01' AND '2020-12-31' THEN 'TPS'
			   WHEN "ACCNB"::bigint >= 1000000000 THEN 'TPS'
			   WHEN "ACCNB"::bigint >= 100000000 THEN 'CRC'
			   ELSE NULL
		   END data_source
	FROM collisions.acc
	WHERE "ACCDATE" BETWEEN '2014-01-01' AND CURRENT_DATE
)
SELECT EXTRACT(year FROM accdate) accyear,
       data_source,
	   COUNT(accnb) n_collisions
FROM step_1
GROUP BY data_source, EXTRACT(year FROM accdate)
ORDER BY 1, 2;
```

accyear | data_source | n_collisions
-- | -- | --
2014 | CRC | 84020
2014 | TPS | 39967
2015 | CRC | 89376
2015 | TPS | 34897
2016 | CRC | 80782
2016 | TPS | 27601
2016 | NULL | 31
2017 | CRC | 81556
2017 | TPS | 24127
2017 | Wut | 2
2017 | NULL | 26
2018 | CRC | 84523
2018 | TPS | 23610
2018 | NULL | 21
2019 | CRC | 83509
2019 | TPS | 22277
2019 | Wut | 3
2019 | NULL | 47
2020 | CRC | 22523
2020 | TPS | 5687
2020 | NULL | 10079
2021 | CRC | 86
2021 | TPS | 2359

That's only somewhat better!

(Aside - we also have 5 collisions that have 2020 ACCNBs but ACCDATEs that are from 2017 or 2019. I've made a note in [Known Issues with ACC](https://www.notion.so/bditto/f1a9ea1e3cd54cf0982d7b3b51b3f2f0?v=6f7bba77c90c4f2c940042e958c08583).)

cczhu:
@candu has investigated this issue on MOVE's end, and refreshed `ACC.dat` - see CityofToronto/bdit_flashcrow#975. Following the refresh, pulled `ACC.dat` into our database.

Now, in our database, running

```sql
SELECT COUNT(*)
FROM collisions.acc WHERE "ACCDATE" BETWEEN '2020-01-01' AND '2020-12-31'
```

returns 62147, a **much** more reasonable number, but also one that's *larger* than the 49119 from Oracle. Breaking this down:

```sql
-- Postgres
SELECT EXTRACT(MONTH from "ACCDATE") accmonth,
       COUNT(*) n_involved
FROM collisions.acc WHERE "ACCDATE" BETWEEN '2020-01-01' AND '2020-12-31'
GROUP BY EXTRACT(MONTH from "ACCDATE")
ORDER BY 1;
```

image


```sql
-- Oracle
SELECT EXTRACT(MONTH FROM accdate) accmonth,
       COUNT(*) n_involved
FROM TRAFFIC.ACC
WHERE accdate BETWEEN TO_DATE('2020-01-01', 'YYYY-MM-DD') AND TO_DATE('2020-12-31', 'YYYY-MM-DD')
GROUP BY EXTRACT(MONTH FROM accdate)
ORDER BY 1
```

image

I'm willing to believe the Postgres numbers much more than the Oracle ones, seeing as the lifting of the first lockdown in June 2020 would have generated more collisions, not less. This may be an issue of the Oracle DB not being periodically refreshed, something that I'll need to take up with @jwrcoleman when he returns from vacation.

cczhu:
Added documentation on getting started with querying collision data. PR now fully ready for review.

KatiRG:
A few comments from reviewing the README:

- statement "TPS and CRC send collision records": how do they actually send the data to us and where is it sent to? Is data still first going to the Oracle database (somehow), from which MOVE then reads and then our schema then reads from MOVE?
- same questions for when "TPS and CRC will also send changes to existing collision records".
- statement "pull_acc_script.py requires: click, psycopg2": can you briefly add what they are needed for

Otherwise it is very clear!

cczhu:
Thanks for the comments, @KatiRG . I've updated the Readme to address the comments.

aharpalaniTO (approved):

Issue #351, opened 2020-11-06
VZ Map - Add 2020 School Safety Zones google sheet to airflow process


KatiRG:
A new sheet for 2020 School Safety Zones is ready to be imported into the VZ Map:

The link to the 2020 sheet is https://docs.google.com/spreadsheets/d/1pJipqKLSuAoYvxiUXHHhdSwTalrag5cbTGxBl1kDSsg/edit#gid=0.

PR #354, opened 2020-11-10
Add 2020 School Safety Zone Google Sheet to dag

KatiRG:
Reads in Google Sheet for 2020 School Safety Zones in the existing scripts that already read 2018 and 2019 data sheets.

## Issue(s) this solves:

- #351 

The additions are in:
- `dags/vz_google_sheets.py`: this is where the 2020 task has been added (`task3`). Also `catchup=False` was added to the `dag` argument list
- `vision_zero/schools.py`: here the specs for the 2020 google sheet were added to the `sheets` object, and then the call to pull from 2020 was added at the end of the script (`pull_from_sheet(con, service, 2020)`). **Note**: the api key is still being read from `'/home/jchew/bdit_data-sources/vision_zero/key.json'`.


dags/Untitled.ipynb
```diff
@@ -0,0 +1,6 @@
+{
+ "cells": [],
+ "metadata": {},
+ "nbformat": 4,
+ "nbformat_minor": 4
+}
```
radumas, line 6: This should be deleted

dags/vz_google_sheets.py
```diff
@@ -42,7 +42,7 @@ def task_fail_slack_alert(context):
 #to read the python script for pulling data from google sheet and putting it into tables in postgres
 try:
   sys.path.append('/etc/airflow/data_scripts/vision_zero/')
-  from schools import pull_from_sheet
+  from schools2020 import pull_from_sheet
```
radumas, line 45: I think this is going to fail, since the schools file didn't get renamed

vision_zero/schools.py
```diff
@@ -31,25 +32,35 @@
```
radumas, line 32: While these lines are being run. They don't actually affect how the airflow script is being run, since that doesn't run the part under `if __name__ == '__main__':`. The DAG is actually already pulling credentials from Airflow's configurations. Just move these lines pulling from Joven's home folder to the `main` block so importing the script in Airflow won't fail if Joven's home folder gets deleted.

https://github.com/CityofToronto/bdit_data-sources/blob/0b77909b07ed0c8c119c93fd8a4253fdad3954ee/vision_zero/schools.py#L28-L31

radumas (changes requested):
Were the approved speed limit reductions supposed to be included?

Thanks for raising the issue of credentials. I dug in a little and  it looks like this was already solved in the `pull_wys` workflow. The Google credentials are being managed by airflow, and a `service` object is created in the DAG and passed to the Python function.

https://github.com/CityofToronto/bdit_data-sources/blob/231517f742b9ad9be697f42ec465d25adb81082e/dags/pull_wys.py#L52-L55

dags/Untitled.ipynb
KatiRG: whoops
done

dags/vz_google_sheets.py
KatiRG: good catch, thanks!

vision_zero/schools.py
KatiRG: done

KatiRG:
> 
> 
> Were the approved speed limit reductions supposed to be included?
> 

Approved speed limit reductions are handled by another workflow. Kenny is managing a shapefile that contains both approved and installed SLRs that he periodically emails to me to import into the database.

vision_zero/schools.py
```diff
@@ -120,7 +115,17 @@ def pull_from_sheet(con, service, year, *args):
...
-    """The following connects to the database, establishes connection to the sheets and executes function based on the year of data required."""
+    """The following connects to the database, establishes connection to the sheets 
+    and executes function based on the year of data required.
+    Note: this part is not run by airflow.
+
+    """
+
+    SERVICE_ACCOUNT_FILE = '/home/jchew/bdit_data-sources/vision_zero/key.json' 
+
+    credentials = service_account.Credentials.from_service_account_file(
```
radumas, line 126: this should be renamed to `service` since that's what gets sent to the `pull_from_sheet` function below

KatiRG: There is already a `service` defined so I'm confused. Is `pull_from_sheet()` supposed to read in `credentials` or `service` as defined by `service = build('sheets', 'v4', credentials=credentials, cache_discovery=False)`?

Here's the whole block:
```
if __name__ == '__main__':
    """The following connects to the database, establishes connection to the sheets 
    and executes function based on the year of data required.
    Note: this part is not run by airflow.

    """

    SERVICE_ACCOUNT_FILE = '/home/jchew/bdit_data-sources/vision_zero/key.json'

    credentials = service_account.Credentials.from_service_account_file(
        SERVICE_ACCOUNT_FILE, scopes=SCOPES)

    CONFIG = configparser.ConfigParser()
    CONFIG.read(r'/home/cnangini/googlesheets_db.cfg')
    dbset = CONFIG['DBSETTINGS']
    con = connect(**dbset)

    service = build('sheets', 'v4', credentials=credentials, cache_discovery=False)

    pull_from_sheet(con, service, 2018)
    pull_from_sheet(con, service, 2019)
    pull_from_sheet(con, service, 2020)
```

radumas: 🤦‍♂️ sorry GitHub hides code blocks and I missed that. Nothing to change here.

radumas (approved):
Great!

Issue #353, opened 2020-11-09
Miovision - add new stations


cczhu:
Using [Joven's documented procedure](https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/miovision#adding-intersections), update our Miovision pipeline to accommodate recently installed stations beyond the downtown core.

cczhu:
Quick lookup of new Miovision stations:

```python
import psycopg2
import configparser

config = configparser.ConfigParser()
config.read(YOUR_CONFIG_FILE)
postgres_settings = config['POSTGRES']

# Get intersections currently in miovision_api.intersections.
with psycopg2.connect(**postgres_settings) as conn:
    df_intcodes = pd.read_sql("SELECT * FROM miovision_api.intersections", con=conn)
old_ints = df_intcodes['id']

# Retrieve intersections from Miovision API, then paste the output to this line.
# api_intersections = ...
api_ints = [cai['id'] for cai in api_intersections]

existing_ints = [x for x in old_ints if x in api_ints]

# Decommisioned stations will be in our Postgres but not in the API.
decommissioned = set(old_ints) - set(existing_ints)
print("Decommissioned", decommissioned)

# New stations will be in the API but not our Postgres.
new_ints = set(current_stations) - set(existing_stations)
print("New stations")
print([x for x in api_intersections if x['id'] in new_ints])
```

There are now newly decommissioned stations. The new stations themselves are:

```
[{'id': 'dbf09553-c593-4bb2-90e5-7eb3bc7ebe08',
  'name': 'Bayview Avenue and River Street',
  'customId': None},
 {'id': '0e574ac8-bb57-46f0-b2a1-49bd48e1a503',
  'name': 'Bloor Street West and Bay Street',
  'customId': None},
 {'id': 'fe26f2d1-41db-4079-9764-7405a4b189f2',
  'name': 'Bloor Street West and Dufferin Street',
  'customId': None},
 {'id': '1ebf4ec0-88fd-49ec-8cf4-3e0ae0af0128',
  'name': 'College Street and University Ave',
  'customId': None},
 {'id': '0d062870-58af-464b-969d-33363a01d9a0',
  'name': 'Danforth Avenue and Broadview Avenue',
  'customId': None},
 {'id': '20249466-ff47-4517-b201-960fe2f361f6',
  'name': 'Danforth Avenue and Logan Avenue',
  'customId': None},
 {'id': '5c952bf4-206d-4671-a142-89d86652819d',
[...]

Issue #341, opened 2020-10-08
Manual Miovision Data Refresh


radumas:
Since using the API in #331 was a bit of a bust, we are getting csv downloads for the data.

radumas:
Reusing some of the structure in the `miovision_csv` schema

```sql
CREATE OR REPLACE FUNCTION miovision_csv.insert_volumes_2020()
    RETURNS trigger
    LANGUAGE 'plpgsql'
    COST 100
    VOLATILE NOT LEAKPROOF
AS $BODY$

BEGIN

	INSERT INTO miovision_csv.volumes_2020 (intersection_uid, datetime_bin, classification_uid, leg, movement_uid, volume)
	SELECT B.intersection_uid, (NEW.datetime_bin AT TIME ZONE 'America/Toronto') AS datetime_bin, C.classification_uid, NEW.entry_dir_name as leg, D.movement_uid, NEW.volume
	FROM miovision_api.intersections B 
	INNER JOIN miovision_csv.movements D ON (D.movement ILIKE NEW.movement)
	INNER JOIN miovision_api.classifications C ON C.classification = NEW.classification AND C.location_only = D.location_only
	WHERE NEW.intersection_id = B.id;
	RETURN NULL;
END;

$BODY$;

ALTER FUNCTION miovision_csv.insert_volumes_2020()
    OWNER TO miovision_admins;

CREATE TABLE miovision_csv.volumes_2020
(
    volume_uid bigserial primary key,
    intersection_uid integer,
    datetime_bin timestamp without time zone,
    classification_uid integer,
    leg text COLLATE pg_catalog."default",
    movement_uid integer,
    volume integer
)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

ALTER TABLE miovision_csv.volumes_2020
    OWNER to miovision_admins;

GRANT ALL ON TABLE miovision_csv.volumes_2020 TO aharpal;

GRANT SELECT, REFERENCES, TRIGGER ON TABLE miovision_csv.volumes_2020 TO bdit_humans WITH GRANT OPTION;

CREATE TABLE miovision_csv.raw_data2020
(
    intersection_id TEXT,
    datetime_bin timestamp with time zone,
    classification text COLLATE pg_catalog."default",
    entry_dir_name text COLLATE pg_catalog."default",
    exit_dir_name text COLLATE pg_catalog."default",
    movement text COLLATE pg_catalog."default",
    volume integer
)
WITH (
[...]

Issue #331, opened 2020-09-21
Miovision Data Refresh


radumas:
For reasons #329 #323 we need to refresh all of the data in `miovision_api.volumes`. The steps are as follows

1. Make a copy of the aggregated metrics we are looking at now for performance, we will compare how this change affects numbers before/after in case anyone notices a difference in products we have previously pushed to stakeholders.
2. Ensure the above issues are complete.
3. 🛑 refreshing and publishing/sending aggregated Miovision data products
4. `TRUNCATE miovision_api.volumes CASCADE` 
5. run `pull_intersection_tmc.py` for all of time
6. truncate the following
    ```sql
    TRUNCATE miovision_api.volumes_15min_tmc;
    TRUNCATE miovision_api.volumes_15min;
    TRUNCATE miovision_api.report_dates;
    TRUNCATE miovision_api.unacceptable_gaps;
    ```

7. to re-aggregate the data, `miovision_api.gapsize_lookup` needs to be refreshed for the dates in question since it is currently refreshed daily to look at the last 60 days. I think the ideal would be to do this in 2 month chunks: 
    a) https://github.com/CityofToronto/bdit_data-sources/issues/334
    b) loop over the `process_data()` Python function in a Python script https://github.com/CityofToronto/bdit_data-sources/blob/a22e407b09534dc01da99e15893725c2917f0248/volumes/miovision/api/intersection_tmc.py#L208-L243
8. Refresh downstream materialized views
9. Do a comparison of before/after data change

KatiRG:
So far I have not worked with the miovision data stored on the RDS. Joven has some scripts that read in bicycle data which I will need later, so I will make sure not to do anything until the refresh is complete.

cczhu:
Backed up materialized views required by covid webapp. Ready to dump raw data as needed.

aharpalaniTO:
Thanks @cczhu, that was my only concern. Nothing else to add here - an estimate on duration of the refresh once you have an idea would be helpful.

Thank you!

cczhu:
Performed `TRUNCATE miovision_api.volumes CASCADE;`. Now running

```bash
python intersection_tmc.py run-api --path /etc/airflow/data_scripts/volumes/miovision/api/config.cfg --start_date 2020-01-01 --end_date 2020-09-23 --pull &> log_refresh_2020.txt
```

cczhu:
[...]

PR #356, opened 2020-11-11
Update Miovision API pull and postprocessing scripts

### LINKED ISSUE #334: Update Miovision Gap Lookup to be more dynamic
To re-aggregate the data for #331, the process to generate the `miovision_api.gapsize_lookup` needs to be refreshed for the dates in question since it is currently refreshed daily to look at the last 60 days from the time at which it is created.
https://github.com/CityofToronto/bdit_data-sources/blob/89941363a54d9a272394d380eed9cb04f8edaaf1/volumes/miovision/sql/create-mat-view-gapsize_lookup.sql#L12
 
Since this was getting refreshed every time `intersection_tmc.py` was being run, we know it doesn't actually take that long to run. So it could just get folded into the [`miovision_api.find_gaps(start_date, end_date)` function](https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/miovision/sql/function-find_gaps.sql). 

I think just add a new CTE at the start that would replicate the structure of the materialized view, but change `now` for the end_date parameter.

https://github.com/CityofToronto/bdit_data-sources/blob/89941363a54d9a272394d380eed9cb04f8edaaf1/volumes/miovision/sql/create-mat-view-gapsize_lookup.sql#L5-L29

and then just delete the code block to refresh the gap size volume lookup table preceding the call to `find_gaps()` in the Python script.

https://github.com/CityofToronto/bdit_data-sources/blob/89941363a54d9a272394d380eed9cb04f8edaaf1/volumes/miovision/api/intersection_tmc.py#L215-L218

### LINKED ISSUE #352: Miovision API - include more recently installed Miovision stations in bulk pulling
Based on [this comment](https://github.com/CityofToronto/bdit_data-sources/issues/341#issuecomment-713335146) in #341. The basic problem is that `intersection_tmc.py` currently checks whether a station has come online before `start_date` and hasn't yet gone offline. When we download an entire year's worth of of data the `start_date` is the start of the year and `end_date` the present, meaning we lose any stations that have gone offline before the present or came online in the middle of the year.

The solution is to pull the activation and deactivation days from `miovision_api.intersections` and check these for each intersection and 6-hour subinterval of time. The overhead to this check shouldn't be that large compared to the [1-second delay](https://github.com/CityofToronto/bdit_data-sources/issues/331#issuecomment-704518820) we instituted a while back.

Once instituted we can do a manual pull of intersections 33 - 40.

cczhu:
Because we need to aggregate data from Miovision intersections that came online in October 2020, we'll need to productionize both the new data pulling method and new 15-minute aggregation technique.

- Remove `create-mat-view-gapsize_lookup.sql` and `function-refresh_gapsize_lookup.sql`, moving the query for gaps into a CTE in `function-find_gaps.sql`. Modified `intersection_tmc.py` to no longer refresh the gap size materialized view before running `find_gaps`.
- Adds notebooks and test scripts related to testing the new gap finder (though there's also work on #331 and #341 in there, oh well).
- Replaces [L312-L329 of `intersection_tmc.py`](https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/miovision/api/intersection_tmc.py#L312) with a class that stores intersection activation/deactivation dates. Now the pull look checks whether a station is active for each individual time interval rather than requiring it be active for the entire duration being pulled.
- A few updates to the Miovision Readme.md to clarify ambiguities found while handling #353.

## Issue(s) this solves:

- Closes #334
- Closes #352

- Review of the query in `function-find_gaps.sql` to make sure it performs the same function (but on a broader timescale) to the gapsize materialized view from before.
- Review of `intersection_tmc.py` to make sure the new activation date class or its use in the `pull_data` loop doesn't have any glaring bugs.

Once productionized, I'll replace the current functions in `miovision_api` and delete any old matviews as needed.


volumes/miovision/README.md
```diff
@@ -143,7 +143,7 @@ intersection_uid|integer|Identifier linking to specific intersection stored in `
 datetime_bin|timestamp without time zone|Start of 1-minute time bin in EDT|2017-10-13 09:07:00|
 classification_uid|text|Identifier linking to specific mode class stored in `classifications`|1|
 leg|text|Entry leg of movement|E|
-movement_uid|integer|Identifier linking to specific turning movement stored in `movements`|2|
+movement_uid|integer|Identifier linking to specific turning movement stored in `movement_map`|2|
```
radumas, line 146: I just created `miovision_api.movements` to store the lookup for API movements. Did not really realise these weren't available before

volumes/miovision/README.md
```diff
@@ -165,7 +165,7 @@ The process in [**Processing Data from CSV Dumps**](#4-processing-data-from-csv-
 
 #### `volumes_15min_tmc`
 
-`volumes_15min_tmc` contains data aggregated into 15 minute bins. Interpolation process is no longer included in the new process now that we have enough data. In the new gap filling process, we try to identify hours of data that we can trust (aka not found in the `unacceptable_gaps` table) and set those 15min bin to NULL instead of 0. To be more specific, 
+`volumes_15min_tmc` contains data aggregated into 15 minute bins. Interpolation process is no longer included in the new process now that we have enough data. In the new gap filling process, we try to identify hours of data that we can trust (aka not found in the `unacceptable_gaps` table) and set those 15min bin to 0 instead of NULL. To be more specific,
```
radumas, line 168: This change is incorrect. Gaps should be stored as `NULL` so they don't get included in averages. This paragraph is conflating two things though:
- Gap finding: where there is no data because we suspect a data issue
- zero-filling: inserting 0s where there is no data and we think that is normal.

volumes/miovision/README.md
```diff
@@ -229,7 +229,7 @@ An example of `miovision_api.movement_map`:
 ![TMC movements](img/intersection_tmc.png)
 
 **Blue line represents `leg_old` whereas yellow arrows represent `movement_uid`**
```
radumas, line 231: This image should be fixed I think. Or you could extend the blue line or reverse the yellow arrows. The blue line half represents `leg_new`, it does not represent `leg_old`, that should be the originating leg.

volumes/miovision/README.md
```diff
@@ -362,7 +362,8 @@ Adding intersections on the other hand is not as simple as removing an intersect
...
 	
-	b) First run the SELECT query below and validate those new intersection movements. The line `HAVING COUNT(DISTINCT datetime_bin::time) >= 20` is there to make sure that the movement is actually legit and not just a single observation. Then, INSERT INTO `intersection_movements` table which only has all the valid movements for old / decommissioned intersections. Keeping the decommissioned ones there too just in case that we might need those in the future.
+	b) First run the SELECT query below and validate those new intersection movements. The line `HAVING COUNT(DISTINCT datetime_bin::time) >= 20` is there to make sure that the movement is actually legit and not just a single observation. Then, INSERT INTO `intersection_movements` table which has all valid movements for intersections. These include decommissioned intersections, just in case we might need those in the future.
```
radumas, line 365: This should maybe mention this is a starting point. As you discovered, the user should definitely validate manually

volumes/miovision/README.md
```diff
@@ -374,14 +375,25 @@ Adding intersections on the other hand is not as simple as removing an intersect
...
 	
-	c) The step before only include valid intersection movements for classification_uid IN (1,2,6) which are light vehicles, cyclists and pedestrians. The reason is that the counts for other mode may not pass the mark of having 20 distinct datetime_bin. However, we know that if vehicles can make that turn, so do trucks, vans and buses which are classification_uid IN (3, 4, 5, 8). Therefore, we will run the below query for each classification_uid that was not included in the previous steps. Note that we can only do one classification_uid at a time and so the following query has to be run 4 times for each intersection_uid, changing classification_uid each time.
+	c) The step before only include valid intersection movements for classification_uid IN (1,2,6) which are light vehicles, cyclists and pedestrians. The reason is that the counts for other mode may not pass the mark of having 20 distinct datetime_bin. However, we know that if vehicles can make that turn, so do trucks, vans, buses and unclassified motorized vehicles, which are classification_uid IN (3, 4, 5, 8, 9). Therefore, we will run the below query for each classification_uid that was not included in the previous steps.
```
radumas, line 378: Great improvement to the query!

volumes/miovision/api/intersection_tmc.py
```diff
@@ -305,70 +301,125 @@ def daterange(start_time, end_time, dt):
         yield (c_start_t, c_start_t + dt)
 
 
+class IntersectionInfo:
```
radumas, line 304: This is an excellent idea.

Was thinking that it might be a more easily interpreted if the `Intersection` class represents a single `Intersection` and then there's a class factory that returns a list. Then the instance methods would apply to singular intersection instances.

radumas (changes requested):
WIP review. Haven't checked out the notebooks yet.

volumes/miovision/README.md
cczhu: Ahh, `movements` should also be documented in the `README.md`.

volumes/miovision/README.md
cczhu: Do you know where Joven kept the (PPT? Illustrator?) file that makes this picture?

volumes/miovision/README.md
cczhu: I'm just confused about the statement:

> identify hours of data that we can trust (aka not found in the `unacceptable_gaps` table) 

which sounds like the documentation is referring to the data that we can trust doesn't have any detection issues, and therefore any zeros in those 15min bins should be 0, not NULL.

radumas: ugggggggggggggh. You're right.

volumes/miovision/README.md
radumas: 😭 
that would have been a good idea to ensure future maintainability.
Don't worry if the graphic sounds like a lot of effort, I would focus on clarifying the error in the text and maybe emphasize that the graphic is a... WIP

volumes/miovision/README.md
cczhu: I'm not certain whether eliminating movements performed by fewer than 20 vehicles even really makes sense, since it potentially leads to eg. u-turns allowed only along one leg. I recommend just removing that clause and asking the user to make sure all movements are physically possible, if not necessarily legal.

volumes/miovision/api/intersection_tmc.py
cczhu: Slightly confused - I should make `Intersection` and `IntersectionArray` classes, the former of which includes date range lookup methods and the latter of which instantiates and stores instances of the former, right? I though a class factory returns classes of different types rather than making a bunch of classes of the same type?

radumas: Maybe factory was the wrong word. I meant a class function that returns a list of `Intersection` objects. An `IntersectionArray` class seems overkill to represent a list of `Intersection` objects IMO.

volumes/miovision/README.md
cczhu: Did some Photoshopping, and revised the diagrams and surrounding text.

cczhu:
Basic tests of intersection class refactoring:

```python
import sys
import datetime
import pytz
import dateutil.parser
import configparser
import pathlib
import psycopg2


config = configparser.ConfigParser()
config.read({YOUR CONFIG FILE})
postgres_settings = config['POSTGRES']


class IntersectionInfoOld:
    """Intersection"""

    def __init__(self, conn, intersection=()):

        with conn.cursor() as cur:
            sql_query = """SELECT intersection_uid,
                                  id,
                                  intersection_name,
                                  date_installed,
                                  date_decommissioned
                           FROM miovision_api.intersections"""
            if len(intersection) > 0:
                sql_query += """ WHERE intersection_uid IN %s"""
                cur.execute(sql_query, (intersection, ))
            else:
                cur.execute(sql_query)
            intersection_list = cur.fetchall()
            intersection_list = dict(
                [(x[0], x[1:]) for x in intersection_list])

        self.int_uids = list(intersection_list.keys())
        self._int_list = intersection_list

    def __getitem__(self, key):
        return self._int_list[key]

    def getinfo(self, key):
        """Returns the ID and name of an intersection.
        
        Parameters
        ----------
        key : int
            Intersection UID (from `miovision_api.intersections`).
        """
        return self[key][:2]

    def is_active(self, int_uid, ctime):
        """Checks if an intersection's Miovision system is active.
        
        Parameters
        ----------
        int_uid : int
            Intersection UID (from `miovision_api.intersections`).
        ctime : datetime.datetime
            Time to check.
        """
        cdate = ctime.date()
        # Check if inputted time is at least 1 day after the activation date
        # of the station.
        active = cdate > self[int_uid][2]
        # If deactivation date exists, check that the time is at least 1 day
        # before.
        if self[int_uid][3] is not None:
            active = active and (cdate < self[int_uid][3])
        return active
    
class IntersectionInfo:

    def __init__(self, uid, id1, name, date_installed,
                 date_decommissioned):
        self.uid = uid
        self.id1 = id1
        self.name = name
        self.date_installed = date_installed
        self.date_decommissioned = date_decommissioned
        
    def __repr__(self):
        return ("intersection_uid: {u}\n"
                "intersection_id1: {id1}\n"
                "name: {n}\n"
                "date_installed: {di}\n"
                "date_decommissioned: {ddc}\n"
                .format(u=self.uid, id1=self.id1,
                        n=self.name, di=self.date_installed,
                        ddc=self.date_decommissioned))

    def is_active(self, ctime):
        """Checks if an intersection's Miovision system is active.
        
        Parameters
        ----------
        ctime : datetime.datetime
            Time to check.
        """
        cdate = ctime.date()
        # Check if inputted time is at least 1 day after the activation date
        # of the station. If deactivation date exists, check that the time is
        # at least 1 day before.
        if self.date_decommissioned is not None:
            return (cdate > self.date_installed) & (cdate < self.date_decommissioned)
        return cdate > self.date_installed


def get_intersections(conn, intersection=()):

    with conn.cursor() as cur:
        sql_query = """SELECT intersection_uid,
                              id,
                              intersection_name,
                              date_installed,
                              date_decommissioned
                       FROM miovision_api.intersections"""
        if len(intersection) > 0:
            sql_query += """ WHERE intersection_uid IN %s"""
            cur.execute(sql_query, (intersection, ))
        else:
            cur.execute(sql_query)
        intersection_list = cur.fetchall()
    
    return [IntersectionInfo(*x) for x in intersection_list]


with psycopg2.connect(**postgres_settings) as conn:
    oldints = IntersectionInfoOld(conn)
    newints = get_intersections(conn)


check_times = [
    dateutil.parser.parse("2019-11-03"),
    dateutil.parser.parse("2020-05-01"),
    dateutil.parser.parse("2020-09-30"),
    dateutil.parser.parse("2020-12-01")
]

for intersection in newints:
    intersection_id1, intersection_name = (
        oldints.getinfo(intersection.uid))
    assert intersection_id1 == intersection.id1
    assert intersection_name == intersection.name

    for ctime in check_times:
        assert (intersection.is_active(ctime)
                == oldints.is_active(intersection.uid, ctime))
```

cczhu:
Okay, should be good for re-review.

volumes/miovision/dev_notebooks/intersection_tmc_notebook03test.py
```diff
@@ -0,0 +1,388 @@
+"""Used to test notebook 03. Identical to intersection_tmc.py on 2020-10-01
```
radumas, line 1: Should update the `dev_notebooks/README` to include an explanation of this script and the 03 notebook

volumes/miovision/README.md
```diff
@@ -10,6 +10,7 @@
 		- [`classifications`](#classifications)
 		- [`intersections`](#intersections)
 		- [`movement_map`](#movement_map)
+		- [`movements`](#movements)
```
radumas, line 13: I think movements should come before the movement_map.

volumes/miovision/README.md
radumas: Sure.

volumes/miovision/sql/function-find_gaps.sql
```diff
@@ -14,10 +14,36 @@ DECLARE tot_gaps integer;
...
+    d.period,
+    mio.hourly_bin::time without time zone AS time_bin,
+    avg(mio.vol) AS avg_vol,
+        CASE
+            WHEN avg(mio.vol) < 100::numeric THEN 20
+            WHEN avg(mio.vol) >= 100::numeric AND avg(mio.vol) < 500::numeric THEN 15
+            WHEN avg(mio.vol) >= 500::numeric AND avg(mio.vol) < 1500::numeric THEN 10
+            WHEN avg(mio.vol) > 1500::numeric THEN 5
+            ELSE NULL::integer
+        END AS gap_size
```
radumas, line 38: this should be called `gap_tolerance` so it's less ambiguous with `gap_size` further down.

volumes/miovision/sql/function-find_gaps.sql
```diff
@@ -14,10 +14,36 @@ DECLARE tot_gaps integer;
 BEGIN
 
 -- FIRST, FIND ALL GAPS
```
radumas, line 16: Is this comment still in the right spot?

volumes/miovision/sql/function-find_gaps.sql
```diff
@@ -27,56 +53,55 @@ SELECT generate_series(start_date, end_date, interval '1 minute')::timestamp wit
...
+	ON sel.intersection_uid = gapsize_lookup.intersection_uid
+	AND DATE_TRUNC('hour', sel.gap_start)::time = gapsize_lookup.time_bin
+	AND (CASE WHEN EXTRACT(isodow FROM sel.gap_start) IN (1,2,3,4,5) THEN 'Weekday'
+	WHEN EXTRACT(isodow FROM sel.gap_start) IN (6,7) THEN 'Weekend'
+	ELSE NULL END)  = gapsize_lookup.period
+), fail AS (
+	-- INSERT INTO THE TABLE
+	-- DISTINCT ON cause there might be more than 1 gap happening in the same hour for the same intersection
+	INSERT INTO miovision_api.unacceptable_gaps(intersection_uid, gap_start, gap_end, gap_minute, allowed_gap, accept)
+	SELECT DISTINCT ON (intersection_uid, DATE_TRUNC('hour', acceptable.gap_start)) *
```
radumas, line 99: I don't love that this isn't actually storing all unacceptable gaps if multiple occur in an hour, but...

I think it's something to note, maybe in the README, maybe in an issue, and come back to eventually.

radumas (changes requested):

volumes/miovision/sql/function-find_gaps.sql
cczhu: Ahh, no, will move it.

volumes/miovision/sql/function-find_gaps.sql
cczhu: I've made a note of it. We can make a new issue after merging with `master`.

volumes/miovision/sql/function-find_gaps.sql
cczhu: Done

volumes/miovision/README.md
cczhu: Done

volumes/miovision/dev_notebooks/intersection_tmc_notebook03test.py
cczhu: Done

radumas (approved):
bananadance

PR #366, opened 2020-11-30
Bulk postprocessing script and CSV data infrastructure

cczhu:

- Adds `miov_bulk_postprocessor.py`, a script that performs only the gap-finding and 15-minute aggregation steps of `intersection_tmc.py` on either CSV or API data.
- Adds functions and tables to handle the recent CSV data dumps.
- Adds development notebooks that generated plots and analyses in #331 and #341.

## Issue(s) this solves:

- #331 
- #341 

- Scan of `miov_bulk_postprocessor.py` for blatant errors.
- Check that all the Readme updates are sensible.
- I don't think there's a huge need to check the Jupyter Notebooks, since they weren't really used to develop any in-production code.


volumes/miovision/api/csv_scripts/miov_bulk_postprocessor.py
```diff
@@ -0,0 +1,172 @@
...
+import dateutil.parser
+import traceback
+import sys
+import datetime
+from requests import exceptions
+from time import sleep
+import math
+
+# Hack to import from parent directory, so don't move this script...
+import sys
```
radumas, line 48: CodeFactor doesn't like that you import sys twice

radumas (changes requested):
Looks good besides the one CodeFactor issue raised.

I also feel like postgresql should do implicit conversion of a `'YYYY-MM-DD'` string to a date but... not the end of the world

cczhu:
Everything should be good to merge now. I'll do a check of whether the API and CSV aggregation functions are identical after merging (since it would require updating the API scripts and so should be a separate PR).

radumas:
Are those the bulk processing ones or the main pipelines ones?

cczhu:
I think there may be some change made to the bulk processing SQL queries that we want to port over to the main pipeline, but I'll have to double-check. If I discover anything I'll make a PR.

To be clear, I don't think anything breaks once `master` is pushed to production.

Issue #374, opened 2021-01-04
New Miovision API crashes our pipeline


cczhu:
Today we discovered that the API has been breaking for the last several days because the crosswalkTMC feed is now reporting bicycles as well as pedestrians crossing at crosswalks. For example, using the Miovision API, for `intersectionID = cc869625-3a14-447f-9156-38283845c47b` from `startTime = 2021-01-03T18:30` to `endTime = 2021-01-03T18:40`, we're getting:

```
...
  {
    "timestamp": "2021-01-03T18:35:00.000-05:00",
    "class": "Bicycle",
    "direction": "CCW",
    "crosswalkSide": "E",
    "qty": 1
  },
  {
    "timestamp": "2021-01-03T18:35:00.000-05:00",
    "class": "Pedestrian",
    "direction": "CCW",
    "crosswalkSide": "E",
    "qty": 1
  },
...
```

We discovered this issue because Raph recently changed the way the volume insert triggers to `miovision_api.volumes_2020` and `miovision_api.volumes_2021` handle duplicates, so it's likely this has been an issue whenever bikes in crosswalks started being included in the streaming data. Before, `intersection_tmc.py` would [incorrectly relabel](https://github.com/CityofToronto/bdit_data-sources/blob/18e300aba8aa270a66f198b7fba02dd1a16219b1/volumes/miovision/api/intersection_tmc.py#L207) any volume at a crosswalk as a pedestrian volume, meaning there is incorrect/excessive ped volumes in the API tables. It's also possible (since the CSV processes mirror the API ones) that the same error exists in the CSV tables.

Things that need to be fixed:
- `intersection_tmc.py` needs to be modified to treat bikes in crosswalk as a Class 7.
- The aggregation process for both API and CSV needs to be updated to handle class 7 movements. This includes appending valid movements in `miovision_api.intersection_movements`.
- We'll need to drop and reaggregate all pedestrian data from the day when bikes in crosswalks came online.

As a stopgap measure, we can hack the `intersection_tmc.py` script to drop any volumes that aren't peds.

cczhu:
I should have noticed this earlier (because why would they only include bike in crosswalk volumes and not also the bike approach volumes) - Miovision is now also including bike approach and departure volumes, which the API is not equipped to handle. For example, for `intersectionId = 2cd73fbd-a5c2-4083-80f4-c487c114f48b` from `startTime = 2021-01-01T12:40:00` to `endTime = 2021-01-01T12:45:00`

[...]

PR #375, opened 2021-01-11
Miovision API bike volume bugfixes, endTime bugfix

cczhu:

- Adds Python code needed to handle bicycles at crosswalks by introducing a Class 7 `classification_uid`. Separate out classification handling for crosswalks into its own function, `get_crosswalk_class`.
- Rename `get_pedestrian_tmc` to `get_crosswalk_tmc`.
- Fix a long-standing bug where `end_iteration_time` is treated as though it were exclusive in `intersection_tmc.py` but `endTime` is actually treated inclusively in the Miovision API. This discrepancy will occasionally lead to vehicle volumes with `datetime_bin = end_iteration_time`. I'm not sure how this managed to never happen until Miovision switched their aggregation backend, which they did at the same time as introducing changes to the bike API, but it's possible their more accurate aggregation does a better job of finding all relevant entries in their data lake.
- Introduce a hack that drops any bike approach volumes. We don't support them yet in the Postgres.

## Issue(s) this solves:

- #374

- Manual check that the changes don't fail on edge cases.
- Confirmation that Class 7 has been properly defined, at least in Python.
- Since we're introducing hacks into `master`, @cczhu **NEEDS TO RAISE TWO NEW ISSUES**:
  - An issue to add Class 7 to `miovision_api.intersection_movements` and `miovision_api.classifications`. We should take that opportunity to also audit `intersection_volumes`. Currently, I don't think it contains every valid movement at an intersection, only those that are more than (an arbitrary) 20 movements in the (also arbitrary) two days prior to Joven or myself adding the movements to the table.
  - An issue to remove the hacks dropping bike approach volumes when we finally support them.


radumas (approved):
Looks good. Ignore the CodeFactor issues....

PR #380, opened 2021-02-02
Synchronize functions between Miovision CSV and API pipelines

### LINKED ISSUE #379: Synchronize Miovision CSV and API Postgres functions
There are currently a few differences between Miovision API and CSV functions:

1. There is an `ON CONFLICT DO NOTHING` on [line 108](https://github.com/CityofToronto/bdit_data-sources/blob/1d5bb273fe630180915481f7be2ad20db60e67d2/volumes/miovision/sql/csv_data/function-find_gaps_2020.sql#L104) of `function-find_gaps_2020.sql` and on [line 52](https://github.com/CityofToronto/bdit_data-sources/blob/1d5bb273fe630180915481f7be2ad20db60e67d2/volumes/miovision/sql/csv_data/function-aggregate-volumes_15min_tmc_2020.sql#L52) of `function-aggregate-volumes_15min_tmc_2020.sql`. They don't exist in the equivalent API files. These are meant to ensure that if we reaggregated historical volumes that volumes already in the 15-minute aggregation are skipped. This lets us, for example, process a vehicle classification or movement that is in the historical data but wasn't supported by the aggregation pipeline before. It might also leave the API pipeline more tolerant to errors like failing silently when aggregating duplicate data.
2. `function-aggregate-volumes_15min_tmc_2020.sql` uses [a better method](https://github.com/CityofToronto/bdit_data-sources/blob/1d5bb273fe630180915481f7be2ad20db60e67d2/volumes/miovision/sql/csv_data/function-aggregate-volumes_15min_tmc_2020.sql#L19) to generate zero-padding movements. This better method doesn't matter if data is being ingested in one-day blocks, but will matter when processing multiple days.
3. Some syntactical differences between `function-find_gaps.sql` and `csv_data/function-find_gaps_2020.sql`. These don't make any difference to how the functions operate, but make it slightly more difficult to do a diff between them just in case we ever need to in the future.

I'm not sure if 1 should be implemented in the API pipeline, and need to chat with @radumas. 2 seems a reasonable thing to implement. 3 has already been implemented.

cczhu:

- Synchronizes Postgres functions for Miovision API and CSV pipelines. In particular
  - Makes variable names more clear in API gap-finding function.
  - Ports the zero-padding CTE from `function-aggregate-volumes_15min_tmc_2020.sql` (CSV version) to `function-aggregate-volumes_15min_tmc.sql` (API).

## Issue(s) this solves:

- Closes #379.

- Scan of the changes to make sure I haven't introduced silly errors.

### DIFF
--- a/volumes/miovision/sql/csv_data/function-aggregate-volumes_15min_tmc_2020.sql
+++ b/volumes/miovision/sql/csv_data/function-aggregate-volumes_15min_tmc_2020.sql
@@ -12,12 +12,12 @@ BEGIN
 
 WITH zero_padding_movements AS (
 	-- Cross product of legal movement for cars, bikes, and peds and the bins to aggregate
-	SELECT m.*, datetime_bin15 
+	SELECT m.*, datetime_bin15
 	FROM miovision_api.intersection_movements m
 	CROSS JOIN generate_series(start_date - interval '1 hour', end_date - interval '1 hour 15 minutes', INTERVAL '15 minutes') AS dt(datetime_bin15)
 	-- Make sure that the intersection is still active
-	LEFT JOIN miovision_api.intersections mai USING (intersection_uid)
-	-- Date installed or 
+	JOIN miovision_api.intersections mai USING (intersection_uid)
+	-- Only include dates during which intersection is active.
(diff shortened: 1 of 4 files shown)


radumas (approved):
Looks good thanks Charles

Issue #196, opened 2019-05-28
Update/Modernize Bluetooth Detector Tables


aharpalaniTO:
Ideally, we should have two tables with the following schemas:

`bluetooth.detectors`: this exists already, but should be modified
**detector_name** - 5-character name unique to the dectector
**street1** - Name of primary street
**street2** - Name of secondary street
**bluetooth_id** - current 4-digit hardware ID for bluetooth detection
**wifi_id** - current 4-digit hardware ID for wifi detection
**detector_name_old** - Any legacy name for the detector
**project** - Project name under which detector was installed
**date_start** - Date when data was first collected at location
**date_end** - Date when data was last collected at location (if applicable)
**latitude**
**longitude**
**loc** - point geometry

`bluetooth.detector_history`:
**detector_name** - same as above
**bluetooth_id** - same as above
**wifi_id** - same as above
**date_installed** - date unit was installed
**date_removed** - data unit was removed (if applicable)

KatiRG:
Created the two tables with the above listed columns.

What rows are supposed to go in them?

aharpalaniTO:
I'll Slack you the file that will inform what goes into those tables

KatiRG:
Is the `bluetooth_id` same as `Unit #` in the xlsx file?

aharpalaniTO:
Yes it is

KatiRG:
@aharpalaniTO ready for your review. The two tables are under my schema.

Both tables created in python notebooks and pushed to EC2 using an `sqlalchemy_mate` engine, as documented here:
https://github.com/CityofToronto/bdit_team_wiki/wiki/Python#Creating-a-new-table-with-geometry-to-push-to-database

radumas:
Does this supersede #175 ?

rickl4:
Can we also tag the `PX` or `int_id` to the bluetooth tables?

aharpalaniTO:
Some weird stuff in the source data - 7 detectors seem to be active with the same Blip ID. I'll need to check with Quin.

KatiRG:
`PX` added to `cnangini.detectors` from `gis.traffic_signals` by string matching street names.  
18 detectors do not have a matching intersection in the gis table:

| detector_name  | street1 | street2
|---|---|---|
|QU_RO | Queen/King |Roncesvalles|
|A | Gardiner |Wickham|
|B | Gardiner |Grand|
|B1 | Gardiner |Roncesvalles|
[...]

PR #381, opened 2021-02-09
Btdag

webgisgeek:

- review and merge changes

## Issue(s) this solves:

- # documentation on creating new bt routes

-


dags/check_brokenreaders.py
```diff
@@ -0,0 +1,81 @@
...
+from psycopg2.extras import execute_values
+from psycopg2 import connect, Error
+import logging
+import configparser
+
+LOGGER = logging.getLogger(__name__)
+logging.basicConfig(level=logging.DEBUG)
+
+#To connect to pgadmin
+CONFIG = configparser.ConfigParser()
```
radumas, line 17: I should set up a bt bot account for Airflow so then the credentials would be imported 
https://github.com/CityofToronto/bdit_data-sources/blob/master/dags/pull_wys.py#L59-L60

dags/check_brokenreaders.py
```diff
@@ -0,0 +1,81 @@
...
+from airflow.hooks.base_hook import BaseHook
+from airflow.contrib.operators.slack_webhook_operator import SlackWebhookOperator
+from psycopg2 import sql
+from psycopg2.extras import execute_values
+from psycopg2 import connect, Error
+import logging
+import configparser
+
+LOGGER = logging.getLogger(__name__)
+logging.basicConfig(level=logging.DEBUG)
```
radumas, line 14: Airflow is already taking care of the `logging` configuration, the code should simply call `logging.info`, etc, further down.

https://stackoverflow.com/a/41841689/4047679

dags/check_brokenreaders.py
```diff
@@ -0,0 +1,81 @@
...
+                
+SLACK_CONN_ID = 'slack'
+def task_fail_slack_alert(context):
+    slack_webhook_token = BaseHook.get_connection(SLACK_CONN_ID).password
+    slack_msg = """
+            :red_circle: A bt Reader may be broken!
+            *Task*: {task}  
+            *Dag*: {dag} 
+            *Execution Time*: {exec_date}  
+            *Log Url*: {log_url}
```
radumas, line 43: This should be updated to our more recent Slack message style.

See for example

https://github.com/CityofToronto/bdit_data-sources/blob/master/dags/pull_wys.py#L28-L44

dags/bt_read_history.py
```diff
@@ -0,0 +1,60 @@
...
+                 'retries': 0,
+                 'retry_delay': timedelta(minutes=5),
+                 'on_failure_callback': task_fail_slack_alert
+                }
+
+dag = DAG('update_bt_status_history',default_args=default_args, schedule_interval='0 12 * * *')
+
+## Postgres update bluetooth reader_status_history DAG
+# Task to update the reader_status_history table daily
+bt_status_history = PostgresOperator(sql='''SELECT * from mohan.reader_status_history(current_date)''',
```
radumas, line 55: This isn't going to work because `current_date` isn't a template parameter.

See for example https://github.com/CityofToronto/bdit_data-sources/blob/master/dags/refresh_wys_monthly.py#L72 and our documentation "**How to use {{ ds }}**"

In short it should just be `{{ ds }}` here

radumas (changes requested):
Some suggestions on the DAGs, also wondering where the `sql` is.

radumas:
@KatiRG @chmnata if you could review the route updating documentation here that would be great.

dags/check_brokenreaders.py
```diff
@@ -0,0 +1,81 @@
...
+
+#To connect to pgadmin
+CONFIG = configparser.ConfigParser()
+CONFIG.read(r'/home/mohan/cre.cfg')
+dbset = CONFIG['DBSETTINGS']
+con = connect(**dbset)
+
+def bad_readers(con, run_date):
+    with con.cursor() as cursor: 
+        select_query = """SELECT reader from mohan.broken_reader()"""
```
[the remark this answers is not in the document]
webgisgeek: forgot about SQL altogether. Will add it to the branch too. Yes date should be a parameter. Will change that

KatiRG:
@webgisgeek Hi Mohan, the readme is pretty clear, I edited a few minor things but there were some things that I didn't understand:

- _In this update, there exists an Excel Sheet template that contains details of newly added bluetooth detectors (`update/img/template.PNG`). The details include proposed route name, description, intersection name (BDIT convention), sensor id and lat/lon at start point, and sensor id and lat/lon at the end points along with numerous other fields._ 
=> there is no `sensor id` column in the `Start point` section

- _The routes that were updated by adding this batch of new detectors were named with a prefix "DT3_"._ 
=> screenshot does not show the new detectors DT3_

- _Depending on how many new readers there are, it may be worthwile to develop an automated process in PostgreSQL, but this update was done manually in the Excel template. In addition to the lat/lon, the streets where the readers are is also needed in 2 columns, and the segment name._ 
=> What are the 2 columns called? There is no "segment name" in the template. 

- _The segment name is always the first two letters of each street, with the corridor street first and the intersecting street second. For example, a reader at Bloor/Christie measuring travel times on Bloor would be named `BL_CH`. For each proposed route, the excel sheet is populated with the `start reader`, `end_reader` and assigned a unique `analysis_id`. For this batch of new readers, analysis ids starting from 1600000 were added._ 
=> There are no columns `start reader`, `end_reader` and `analysis_id` in the template.

- _Therefore the reader table will have the following fields populated:  `analysis_id`, `street`, `direction`, `from_street`, `to_street`, `from_id`, `to_id`, `start_point_lat`, `start_point_lon`, `end_point_lat` and `end_point_lon`._
=> these columns are not in the screenshot. Where are they supposed to be added? Also, there are many columns in the template that are not explained. e.g. `!ozone`, `!zonelist`, `Maxtime`, etc

- _After uploading excel data to PostgreSQL_
=> to where? `bluetooth.all_analyses` ?

- Should the query `SELECT DISTINCT mohan.new_added_detectors.from_id::integer AS bluetooth_id`  have `CREATE TABLE mohan.bluetooth_nodes AS` at the beginning?

- _especially for oddly shaped intersections._
=> can you give an example of oddly-shaped intersections? What happens with them? How to fix them?

- _The table is now ready to be appended to the existing routes table._ 
=> can you show the query to append the table?

dags/check_brokenreaders.py
```diff
@@ -0,0 +1,122 @@
...
+#dbset = CONFIG['DBSETTINGS']
+con = bt_postgres.get_conn()
+broken_list = []
+
+def pipeline_check(con):
+    with con.cursor() as cursor:
+        select_query1 = '''SELECT MAX (datetime_bin)::date from bluetooth.aggr_5min'''
+        cursor.execute(select_query1)
+        latest_date = cursor.fetchone()
+        if (latest_date[0]) >= (datetime.now().date() - timedelta(1)):
```
radumas, line 32: This should be using `{{ ds }}` as well.

dags/check_brokenreaders.py
```diff
@@ -0,0 +1,122 @@
...
+                }
+with DAG('blip_check_update', default_args=default_args, schedule_interval='0 17 * * *') as blip_pipeline:
+    task1 = PythonOperator(
+    task_id = 'pipeline_check',
+    python_callable = pipeline_check,
+    dag=blip_pipeline,
+    op_kwargs={
+        'con': con
+        })
+    task2 = PostgresOperator(sql='''SELECT * from mohan.insert_report_date()''',
```
radumas, line 95: This should probably be using `{{ ds }}`

dags/check_brokenreaders.py
```diff
@@ -0,0 +1,122 @@
...
+LOGGER = logging.getLogger(__name__)
+logging.basicConfig(level=logging.DEBUG)
+
+#to connect to pgadmin bot
+bt_postgres = PostgresHook("bt_bot")
+
+#To connect to pgadmin
+#CONFIG = configparser.ConfigParser()
+#CONFIG.read(r'/home/mohan/cre.cfg')
+#dbset = CONFIG['DBSETTINGS']
```
radumas, line 23: these should be deleted

dags/check_brokenreaders.py
```diff
@@ -0,0 +1,122 @@
...
+                'on_failure_callback': task_fail_slack_alert
+                }
+with DAG('blip_check_update', default_args=default_args, schedule_interval='0 17 * * *') as blip_pipeline:
+    task1 = PythonOperator(
+    task_id = 'pipeline_check',
+    python_callable = pipeline_check,
+    dag=blip_pipeline,
+    op_kwargs={
+        'con': con
+        })
```
radumas, line 94: This should be tabbed over properly for readability

dags/check_brokenreaders.py
```diff
@@ -0,0 +1,122 @@
...
+                            retries = 0,
+                            dag=blip_pipeline)                        
+    task5 = PythonOperator(
+    task_id = 'broken_readers',
+    python_callable = broken_readers,
+    dag=blip_pipeline,
+    op_kwargs={
+        'con': con,
+        'check_date': '{{ ds }}'
+        })
```
radumas, line 121: This should be tabbed over properly for readability

dags/check_brokenreaders.py
```diff
@@ -0,0 +1,122 @@
...
+                            dag=blip_pipeline)                        
+    task5 = PythonOperator(
+    task_id = 'broken_readers',
+    python_callable = broken_readers,
+    dag=blip_pipeline,
+    op_kwargs={
+        'con': con,
+        'check_date': '{{ ds }}'
+        }) 
+task1 >> [task2, task3, task4, task5]
```
radumas, line 122: Are tasks 2 to 5 really independent of each other?

radumas (changes requested):
We're getting there! Close to functioning automated processes.

This section needs to be updated to discuss the updated functions you've written. 

https://github.com/CityofToronto/bdit_data-sources/blame/5dad382a6f709d9c397148008000b8768bab90ec/bluetooth/README.md#L301-L308

When the DAG works we're going to have to change everything to run from the `bluetooth` schema.

bluetooth/README.md
```diff
@@ -167,6 +171,79 @@ substring(cod::bit(24) from 17 for 6) as minor_device_class,
...
+- `serial_no_wifi` : Some readers have both wifi and bluetooth sensors. Those readers which has the wifi serial number is available is populated in this field
+
+- `date_installed` : Date the reader is installed at the location
+
+- `date_uninstalled` : Date when reader is uninstalled from a location
+
+
+#### reader_locations 
+
+This is a table of all locations at which Bliptrack readers have been installed and are physically existing. The installed readers could be online or offline but has NOT been removed physically. Each intersection has only **one** reader that is assigned to a route/routes. Therefore, if there are more than one readers in a locations that has not been removed, such detectors are listed in the `reader_history` table. This table consists of the following fields:
```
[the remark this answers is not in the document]
webgisgeek: @radumas. There is a lookup table named reader_status_history which is also updated by the function and is included as one of the tasks in DAG. Is it necessary to include that table as well ?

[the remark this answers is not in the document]
radumas: yep

radumas:
`check_brokenreaders.py` should be renamed to `bluetooth_check_readers.py` so that the data source is the first word. I should update the documentation to specify a standard for how we name DAGs.

bluetooth/README.md
```diff
@@ -167,6 +172,91 @@ substring(cod::bit(24) from 17 for 6) as minor_device_class,
 substring(cod::bit(24) from 12 for 5) as major_device_class
 ```
 
+As a part of updating/modernizing Bluetooth Reader Tables [ISSUE #196](https://github.com/CityofToronto/bdit_data-sources/issues/196), the following four tables are prepared.  
+* `reader_history`
+* `reader_locations`
+* `routes`
+* `reader_status_history`
```
radumas, line 179: I know this is supposed to be introducing the new tables you've created, but this comment doesn't make sense where it is, which is under the `ClassOfDevice` heading.
This should be placed under the `Internal Tables`, and maybe move the sections for the individual tables above the `ClassOfDevice` section

radumas (changes requested):
We're getting there.

There should only be 1 DAG in this repository, however, please delete the other one. The proper way to rename a file is to use `git mv` to reduce the effect on the history.

There needs to be an explanation of what that DAG does under section 5 of the main README: explaining what the purpose of the validation routines are.

Please carefully review the comments on what's been committed, I've noticed some still haven't been acted upon. Let me know if you have questions about them.

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,118 @@
...
+
+    COST 100
+    VOLATILE 
+AS $BODY$	
+	begin
+		--return query
+			with x AS 
+			(select DISTINCT(analysis_id), 
+			 CASE
+			 WHEN MAX (datetime_bin::date)>= (insert_value-1) then (insert_value-1)
```
radumas, line 18: This comparison is wrong. `insert_value` is the `execution_date`, which would be the day over which the datetime_bins came in. So there shouldn't be `-1` in this comparison.

bluetooth/sql/functions/broken_readers.sql
```diff
@@ -0,0 +1,37 @@
...
+select *
+from bluetooth.reader_status_history
+left join x using (reader_id)
+where bluetooth.reader_status_history.active IS false and (bluetooth.reader_status_history.dt = insert_dt and last_active_date = insert_dt-2)
+
+)
+INSERT INTO bluetooth.broken_readers_log(reader_id, reader_name, check_date)
+select reader_id, name, dt
+from y
+left join bluetooth.reader_locations using(reader_id)
```
radumas, line 27: This `broken_readers_log` is redundant. The same information can be gleaned from the `reader_status_history` table.
Also, this function doesn't return anything, it is only `INSERT`ing, so the check in the DAG task won't do anything.

dags/bluetooth_check_readers.py
```diff
@@ -0,0 +1,127 @@
...
+                'start_date': datetime(2021, 4, 29),
+                'email': ['mohanraj.adhikari@toronto.ca'],
+                'email_on_failure': False,
+                'email_on_success': False,
+                'retries': 0,
+                'retry_delay': timedelta(minutes=5),
+                'on_failure_callback': task_fail_slack_alert
+                }
+with DAG('bluetooth_check_readers', default_args=default_args, schedule_interval='0 17 * * *', catchup=False) as blip_pipeline:
+    task1 = PythonOperator(
```
radumas, line 93: Rename the `taskN` Python variables with their `task_id` so it's easier to understand what each one is doing.

dags/bluetooth_check_readers.py
```diff
@@ -0,0 +1,127 @@
...
+                            dag=blip_pipeline)                        
+    task5 = PythonOperator(
+    task_id = 'broken_readers',
+    python_callable = broken_readers,
+    dag=blip_pipeline,
+    op_kwargs={
+        'con': con,
+        'check_date': '{{ ds }}'
+        }) 
+task1, task2, task3, task4, task5
```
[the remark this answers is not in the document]
radumas: The `broken_readers` task depends on the `reader_status_history` one, they can't happen independently.

dags/bluetooth_check_readers.py
```diff
@@ -0,0 +1,127 @@
...
+                raise Exception ('There is no data in bluetooth.aggr_5min for yesterday')
+        except:
+            pass
+
+def broken_readers(con, check_date):
+    with con.cursor() as cursor: 
+        select_query2 = '''SELECT * from bluetooth.broken_readers(%s::date)'''
+        cursor.execute(select_query2, (check_date,))
+        broken_readers = cursor.fetchall()
+        broken_list.append(broken_readers)
```
radumas, line 43: This step seems unnecessary

dags/bluetooth_check_readers.py
```diff
@@ -0,0 +1,127 @@
...
+        num_broken = len(broken_list)
+        #please suggest if this try except works
+        try:
+            if broken_readers[0][0] == '':
```
radumas, line 47: why not use `num_broken`?

The Slack message here should print out the broken readers

dags/bluetooth_check_readers.py
```diff
@@ -0,0 +1,127 @@
...
+                raise Exception ('Some readers are down. Please check broken_readers_log')
+        except:
+            pass    
+           
+        
+SLACK_CONN_ID = 'slack_data_pipeline'
+def task_fail_slack_alert(context):
+    slack_webhook_token = BaseHook.get_connection(SLACK_CONN_ID).password
+    
+    if context.get('task_instance').task_id == 'task1':
```
radumas, line 59: These values for `task_id` are wrong, in this first case it would `pipeline_check`

dags/bluetooth_check_readers.py
```diff
@@ -0,0 +1,127 @@
...
+        task_msg = """:among_us_dead: Task {task} in blip check update failed, 
+            <@U01858E603T> fix it asap.""".format(
+                task=context.get('task_instance').task_id,)
+    
+    elif context.get('task_instance').task_id == 'task5':
+        task_msg = """:among_us_ghost: some readers {task} have failed. Check the log, <@U01858E603T>.""".format(
+                task=context.get('task_instance').task_id,)
```
radumas, line 66: This can be deleted.

dags/bluetooth_check_readers.py
```diff
@@ -0,0 +1,127 @@
...
+        except:
+            pass    
+           
+        
+SLACK_CONN_ID = 'slack_data_pipeline'
+def task_fail_slack_alert(context):
+    slack_webhook_token = BaseHook.get_connection(SLACK_CONN_ID).password
+    
+    if context.get('task_instance').task_id == 'task1':
+        task_msg = """:among_us_dead: Task {task} in blip check update failed,
```
radumas, line 60: This should read "No bluetooth data was found in the database for this date. Remote desktop to the terminal server and check the `blip_api` log".

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,118 @@
...
+    active.id,
+	active.dt
+   FROM active)
+   
+   INSERT INTO bluetooth.reader_status_history (reader_id, last_active_date, active, dt)
+   SELECT DISTINCT reader_id, max(max),route_status::bool, dt
+   from final
+   where reader_id IS NOT NULL
+   group by reader_id, route_status, dt
+   ORDER BY reader_id
```
radumas, line 112: Most of the CTEs in this function and `reader_locations_dt_update` is the same. This `INSERT` can be in a CTE and then the `UPDATE` from `reader_locations_dt_update.sql` could be run below it.

Then there could be one task to do both in the DAG

radumas (changes requested):
Some changes to simplify automated tasks and some bug fixes.

chmnata:
Seems like the below two DAGs have been deleted and are no longer in the branch. I will rm them in the dag folder on the ec2 as well. 
bt_read_history.py -> /etc/airflow/dev_scripts/dags/bt_read_history.py
bt_test_dag.py -> /etc/airflow/dev_scripts/dags/bt_test_dag.py

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,122 @@
...
+           FROM b
+        UNION
+         SELECT z.analysis_id,
+            z.end_detector AS detector_name,
+            z.route_status,
+            z.last_reported,
+			z.dt
+           FROM z
+        )
+		, active AS (
```
chmnata, line 78: can be included in the `final` cte

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,122 @@
...
+         SELECT DISTINCT c.detector_name,
+            max(c.last_reported) AS max,
+            c.route_status,
+            bluetooth.detectors_history_final.reader_id as id,
+			c.dt
+           FROM c
+             LEFT JOIN bluetooth.detectors_history_final ON c.detector_name = bluetooth.detectors_history_final.read_name::text
+          WHERE c.route_status = 'True'::text
+          GROUP BY c.route_status, c.detector_name, bluetooth.detectors_history_final.reader_id, c.dt
```
chmnata, line 87: no need to add schema name

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,122 @@
...
+		--return query
+			with x AS 
+			(select DISTINCT(analysis_id), 
+			 CASE
+			 WHEN MAX (datetime_bin::date)>= (insert_value) then (insert_value)
+			 ELSE MAX (datetime_bin::date)
+			 END 
+			 as last_reported,
+			CASE
+			WHEN  MAX (datetime_bin::date)>= (insert_value) then 'True'::text
```
chmnata, line 23: this should be kept as boolean because at the very end it got casted back to boolean

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,122 @@
...
+           FROM z
+        )
+		, active AS (
+         SELECT DISTINCT c.detector_name,
+            max(c.last_reported) AS max,
+            c.route_status,
+            bluetooth.detectors_history_final.reader_id as id,
+			c.dt
+           FROM c
+             LEFT JOIN bluetooth.detectors_history_final ON c.detector_name = bluetooth.detectors_history_final.read_name::text
```
chmnata, line 85: no need to cast read_name to text

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,122 @@
...
+         SELECT z.analysis_id,
+            z.end_detector AS detector_name,
+            z.route_status,
+            z.last_reported,
+			z.dt
+           FROM z
+        )
+		, active AS (
+         SELECT DISTINCT c.detector_name,
+            max(c.last_reported) AS max,
```
chmnata, line 80: should be called last_reported

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,122 @@
...
+          WHERE c.route_status = 'True'::text
+          GROUP BY c.route_status, c.detector_name, bluetooth.detectors_history_final.reader_id, c.dt
+        ), final as (
+ SELECT DISTINCT c.detector_name,
+    max(c.last_reported) AS max,
+    c.route_status,
+    bluetooth.detectors_history_final.reader_id,
+			c.dt
+   FROM c
+     LEFT JOIN bluetooth.detectors_history_final ON c.detector_name = bluetooth.detectors_history_final.read_name::text
```
chmnata, line 95: not exactly sure what is detectors_history_final as it is not bluetooth schema and not documented

bluetooth/sql/functions/reader_locations_dt_update.sql
```diff
@@ -0,0 +1,113 @@
+-- FUNCTION: bluetooth.reader_locations_dt_update(date)
+
+-- DROP FUNCTION bluetooth.reader_locations_dt_update(date);
+
+CREATE OR REPLACE FUNCTION bluetooth.reader_locations_dt_update(
```
chmnata, line 5: Since this is included in reader_status_history I think this can be deleted

bluetooth/sql/functions/reader_status_history.sql
chmnata: Found 5 `detectors_history_*` tables in schema `mohan`, 
`detectors_history_final`
`detectors_history`
`detectors_history_late`
`detectors_history_latest`
`detectors_history_corrected`

chmnata: seems like the reason to join to this table is to get the corresponding `reader_id` which is needed in the output. This should also be an inner join

bluetooth/README.md
```diff
@@ -134,6 +161,72 @@ The script pulls the route configurations nightly from the Blip server. These ar
...
+|`reader_history_id`|integer| Unique ID for each reader|
+|`reader_id`|varchar|foreign key to `reader_locations`. This is a location id. Multiple readers could have been installed in a same location at different times.|  
+|`serial_no_bluetooth`|integer|This is a four digit number that is assigned to each bluetooth reader. This serial number corresponds to the **Zone** in the bliptrack table.| 
+|`serial_no_wifi`|integer|Some readers have both wifi and bluetooth sensors. For those readers which has the wifi sensor its serial number is populated.|
+|`date_installed`|date|Date the reader is installed at the location|
+|`date_uninstalled`|date|Date the reader is uninstalled at the location|
+
+Except the `reader_history_id` all other fields in this table has to be updated manually.
+
+#### reader_locations
```
chmnata, line 183: detector_history table is missing in this doc

chmnata:
For function `reader_status_history`, there is a join to a `detectors_history_final` table which was not documented in the readme. There are also 5 versions of it in the schema currently. The purpose of the join is to get the reader_id from `detectors_history_final` using `read_name` = `detector_name`. However, there are duplicated `read_name` in `detectors_history_final` as it contains all the past read_name for every reader_id. I don't think the historical read_name matters for this function as we are only running the route configurations that are currently in place. So I checked out another table called `reader_locations` and it seems to have the same read_name and detector_name relationship as the  `detectors_history_final` table, but without the historic values. I think this would be a more suitable table to get the reader_id unless there is a look up table somewhere else...

bluetooth/sql/functions/broken_readers.sql
```diff
@@ -0,0 +1,40 @@
...
+    RETURNS TABLE(read_id integer, reader_name character varying) 
+    LANGUAGE 'plpgsql'
+
+    COST 100
+    VOLATILE 
+    ROWS 1000
+AS $BODY$	
+	begin
+    return query 
+    with x as(
```
chmnata, line 16: cte name should be called something descriptive

bluetooth/sql/functions/broken_readers.sql
```diff
@@ -0,0 +1,40 @@
...
+)
+select distinct (reader_id), read_name from x 
+left join y using(reader_id)
+left join bluetooth.detectors_history_corrected using(reader_id)
+where active_yesterday is true and active_today is false
+;
+end; $BODY$;
+
+ALTER FUNCTION bluetooth.broken_readers(date)
+    OWNER TO bluetooth;
```
chmnata, line 36: bluetooth is not a group or a role, should be `bt_admins`

bluetooth/sql/functions/broken_readers.sql
```diff
@@ -0,0 +1,40 @@
...
+
+GRANT EXECUTE ON FUNCTION bluetooth.broken_readers(date) TO PUBLIC;
+
```
chmnata, line 39: Add a comment to describe what this function is doing

bluetooth/sql/functions/reader_locations_dt_update.sql
```diff
@@ -0,0 +1,113 @@
...
+group by reader_id
+order by reader_id)
+
+UPDATE bluetooth.reader_locations
+set date_last_received = (SELECT date_last_received from date_refresh where date_refresh.reader_id = bluetooth.reader_locations.reader_id)
+	;
+end; $BODY$;
+
+ALTER FUNCTION bluetooth.reader_locations_dt_update(date)
+    OWNER TO bluetooth;
```
chmnata, line 113: change to bt_admins

bluetooth/sql/functions/broken_readers.sql
```diff
@@ -0,0 +1,40 @@
...
+    with x as(
+select reader_id, last_active_date as yesterday, active as active_yesterday
+	from bluetooth.reader_status_history
+where last_active_date = insert_dt::date - 2
+order by reader_id
+) 
+,y as (
+select reader_id, last_active_date as today, active as active_today
+	from bluetooth.reader_status_history
+where last_active_date = insert_dt::date - 1 and active is false
```
chmnata, line 25: the active is false is not needed as it is also in the where clause in the below cte

bluetooth/sql/functions/broken_readers.sql
```diff
@@ -0,0 +1,40 @@
...
+order by reader_id
+)
+select distinct (reader_id), read_name from x
```
chmnata, line 28: I think this can be simplified to
WHERE active is False and last_active_date was yesterday. Aka the reader is currently not active but the last active day was yesterday. Then we will not need any ctes from above.

bluetooth/sql/functions/broken_readers.sql
```diff
@@ -0,0 +1,40 @@
...
+) 
+,y as (
+select reader_id, last_active_date as today, active as active_today
+	from bluetooth.reader_status_history
+where last_active_date = insert_dt::date - 1 and active is false
+order by reader_id
+)
+select distinct (reader_id), read_name from x 
+left join y using(reader_id)
+left join bluetooth.detectors_history_corrected using(reader_id)
```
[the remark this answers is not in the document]
chmnata: left join also results in duplicated rows, with one reader_id and multiple reader_names even when some are offline

bluetooth/sql/functions/broken_readers.sql
```diff
@@ -0,0 +1,40 @@
...
+where active_yesterday is true and active_today is false
+;
+end; $BODY$;
+
+ALTER FUNCTION bluetooth.broken_readers(date)
+    OWNER TO bluetooth;
+
+GRANT EXECUTE ON FUNCTION bluetooth.broken_readers(date) TO PUBLIC;
+
+GRANT EXECUTE ON FUNCTION bluetooth.broken_readers(date) TO bluetooth;
```
chmnata, line 40: this should be either bt_bot or bt_admin_bot

chmnata: going to put it as bt_bot as of now

bluetooth/sql/functions/insert_report_date.sql
```diff
@@ -0,0 +1,28 @@
...
+CREATE OR REPLACE FUNCTION bluetooth.insert_report_date(
+	)
+    RETURNS void
+    LANGUAGE 'plpgsql'
+    COST 100
+    VOLATILE PARALLEL UNSAFE
+AS $BODY$
+	
+	begin
+		with x AS
```
chmnata, line 14: change to descriptive cte name

dags/bluetooth_check_readers.py
```diff
@@ -0,0 +1,119 @@
...
+logging.basicConfig(level=logging.DEBUG)
+
+#to connect to pgadmin bot
+bt_postgres = PostgresHook("bt_bot")
+
+con = bt_postgres.get_conn()
+broken_list = []
+# check if the blip data was aggregated into aggr_5min bins as of yesterday.
+# if the aggr_5min table has at least one record for yesterday, it means pipeline is working 
+def pipeline_check(con, check_date):
```
chmnata, line 25: !!! I can we can use SQLcheckoperator here !!!!

bluetooth/sql/create_tables/add_new_segments.sql
```diff
@@ -0,0 +1,31 @@
...
+)
+SELECT analysis_id, street, direction, from_street, to_street,
+	CASE WHEN geom_dir != direction THEN ST_reverse(geom) 
+	ELSE geom 
+	END AS 
+	geom
+FROM ( 
+SELECT analysis_id, street, direction, from_street, to_street, 
+		gis.twochar_direction(gis.direction_from_line(ST_linemerge(ST_union(geom)))) AS geom_dir,
+		ST_linemerge(ST_union(geom)) AS geom
```
chmnata, line 29: the table should also have a length column

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,117 @@
...
+   GROUP by          reader_id, route_status, dt
+   ORDER BY          reader_id;
+
+   UPDATE bluetooth.reader_locations
+   set date_last_received = (SELECT DISTINCT max(last_reported) from detector_status where bluetooth.detectors_history_final.reader_id = bluetooth.reader_locations.reader_id);
+END;
+$BODY$;
+
+ALTER FUNCTION bluetooth.reader_status_history(date)
+    OWNER TO bluetooth;
```
chmnata, line 117: owner to bt_admins

bluetooth/update/README.md
```diff
@@ -2,145 +2,133 @@
...
+`from_street` 		- Name of the intersecting street where the route begins.
+`to_street`			- Name of the intersecting street where the route ends 
+`from_name`			- The reader name (bdit convention) at the start point of the route
+`from_id` 			- The four digit Unique bluetooth id at the start point
+`from_lat` 			- Latitude at the start point of the route
+`from_lon` 			- Longitude at the start point of the route
+`to_name` 			- The reader name (bdit convention) at the end point of the route
+`to_id`				- The four digit Unique bluetooth id at the end point
+`to_lat`			- Latitude at the end point of the route
+`to_lon`			- Longitude at the end point of the route
```
chmnata, line 47: add a length column

bluetooth/update/README.md
```diff
@@ -2,145 +2,133 @@
...
 
-```SQL
-WITH from_join AS (
-SELECT * FROM rliu.new_bt_corridors B
-CROSS JOIN rliu.new_bt_readers A
-WHERE levenshtein(A.reader_name, B.from_intersection, 1, 1, 2) < 4)
-```
 
-Matches the corridors to the names based on similar from intersections.
+This .csv file was then imported into QGIS and pushed into the mohan schema in postgres. The geometry column is created by adding a column using `ALTER TABLE mohan.new_added_detectors ADD COLUMN from_geom GEOMETRY` we need two geometry columns: `from_geom` and `to_geom`.  Filling in the from_geom column with `UPDATE mohan.new_added_detectors SET to_geom = ST_SetSRID(ST_MakePoint(to_lon, to_lat), 4326)` and modify the query for `from_geom`.
```
chmnata, line 60: I feel like the csv can be imported directly to postgres instead of going through QGIS, since QGIS function wasn't used

bluetooth/update/README.md
```diff
@@ -2,145 +2,133 @@
...
-, joined_readers AS (
-SELECT A.analysis_id, A.corridor, A.reader_name AS from_intersection, A.long AS from_long, A.lat AS from_lat, B.reader_name AS to_intersection, B.long AS to_long, B.lat AS to_lat FROM from_join A
-INNER JOIN to_join B USING (analysis_id)
-)
-```
+## Finding Nearest Intersection IDs
 
-Makes a single table for the corridors and readers at the from/to intersections.
+To get the intersection ids that are close to the newly added detectors location, create a table named `bluetooth_nodes`. This table has four fields:
+`bluetooth_id`, `geom` (this is geometry of bluetooth detectors), `int_id` (nearest intersection id) and `int_geom` (geometry of the nearest intersection to the )
```
chmnata, line 65: unfinished sentence

bluetooth/update/README.md
```diff
@@ -2,145 +2,133 @@
...
+CREATE TABLE mohan.bluetooth_nodes AS(
+SELECT DISTINCT mohan.new_added_detectors.from_id::integer AS bluetooth_id,
+    mohan.new_added_detectors.from_geom,
+    nodes.int_id,
+    st_transform(nodes.node_geom, 4326) AS int_geom
+   FROM mohan.new_added_detectors
+     CROSS JOIN LATERAL ( SELECT z.int_id,
+            st_transform(z.geom, 98012) AS node_geom
+           FROM gis.centreline_intersection z
+          ORDER BY (z.geom <-> mohan.new_added_detectors.from_geom)
```
chmnata, line 78: would we miss detectors because only from detector's geometry was used?

chmnata: feel like we should be using a distinct detector id of both from and to detector

bluetooth/update/README.md
```diff
@@ -2,145 +2,133 @@
...
+         LIMIT 1) nodes);
 ```
-Draws lines using the reader locations. This is the result.
+Check that correct intersections are returned from this query especially for oblique intersections with an offset. If required, correct the intersection_id and geom for such intersections and finalize the table `mohan.bluetooth_nodes`. An example is shown below. The intersection between Dundas St W, Roncesvalles Ave and Boustead Ave is oblique with an offset. The BT reader 5263 (red dot) picked up intersection_id `13466305` (greed dot) as nearest intersection to the reader. But the planned route does not go through this intersection_id. The correct intersection_id for this reader is `113466258`. The intersection_id and geom for this redear was therefore corrected manually.
```
chmnata, line 81: extra 1 in `113466258`

bluetooth/update/README.md
```diff
@@ -2,145 +2,133 @@
...
+INNER JOIN gis.centreline ON edge=geo_id
+INNER JOIN mohan.new_added_detectors USING (analysis_id)
+ORDER BY analysis_id
 )
-
-SELECT * 
-INTO rliu.new_bt_coflate1000 
-FROM conflation
+SELECT analysis_id, street, direction, from_street, to_street,
+	CASE WHEN geom_dir != direction THEN ST_reverse(geom)
```
chmnata, line 111: not sure if this is needed because we are using centreline with direction as a base network for which the direction of the line is already flipped. This might result in double flipped geometries if the dirs didnt match because of a NB vs WB comparison

bluetooth/update/README.md
```diff
@@ -2,145 +2,133 @@
...
 ```
 
-Matches the centreline road segment to the drawn lines. Note the buffer is really large (1000 units) to account for segments that run accross multiple roads.
-
-After proper QC (see next section) is applied, it should something like this.
+![bt_new_segments](img/new_segments.JPG)
 
-![bt_segments](img/bt_segments.PNG)
+## Things to note 
+Geostatistical lines and planning boundaries need to be avoided while pgrouting.
```
chmnata, line 126: also railways and rivers and stuff

bluetooth/update/README.md
```diff
@@ -2,145 +2,133 @@
...
-After proper QC (see next section) is applied, it should something like this.
+![bt_new_segments](img/new_segments.JPG)
 
-![bt_segments](img/bt_segments.PNG)
+## Things to note 
+Geostatistical lines and planning boundaries need to be avoided while pgrouting. 
 
-## Cleanup, Warnings, and Alternative Methods
+## Validating Output
+Validate the length of the segments with length ST_length(geom) and direction using gis.direction_from_line(geom) functions.If the detectors are located very close to the centerline intersections, it is not necessary to do the  centreline cutting. If any bluetooth detectors are not located at the start or end point of a centreline, we will need to cut the centreline using ST_linelocatepoint() as explained in [here.](https://github.com/CityofToronto/bdit_data-sources/issues/234) .
```
chmnata, line 129: I think it would be helpful to include the steps to cut the line here in the readme instead of linking the issue.

bluetooth/update/README.md
```diff
@@ -2,145 +2,173 @@
...
+```sql
+ST_linelocatepoint(geom, closest_point_geom)
+```
+3) Cut the line using [`ST_linesubstring()`](https://postgis.net/docs/ST_LineSubstring.html) 
+```sql
+ST_linesubstring(geom, 0, st_linelocatepoint)
+```
 
-At this point, the layer has multiple entries containing centreline segments for each analysis_id. Instead, we need a single line/entry for each analysis_id. This query concatenates the individual centreline segments into one line. The main function doing this work is [`ST_LineMerge()`](https://postgis.net/docs/ST_LineMerge.html)
+The new routes table is now ready to append to the existing routes table.
```
chmnata, line 172: The `bluetooth.segments` table has the columns:     

- segment_name ,
- analysis_id,
- street ,
- direction,
- start_cross ,
- end_cross ,
- length ,
- bluetooth ,
- wifi ,
- geom ,
- duplicate ,
- reversed ,
- end_street

So its missing: 
- segment_name , bluetooth, widi, duplicate, revered, end_street

bluetooth/update/reader_status_details.md
```diff
@@ -0,0 +1,40 @@
+# Bluetooth reader's status
```
chmnata, line 1: Maybe this can go under the readme?

chmnata:
List of modified things:
SQL
- Functions:
	- broken readers: simplified, need to know which version of `detectors_history_corrected` should be used
	- insert report date: simplified 
	- reader locations dt update: should be deleted
	- reader status history: simplified, need to know which version of `detectors_history_final` should be used
- Table:
	- added length column to mohan.bt_segments_new
	- OWNER to bt_admins 
	- Change create new segment to not ST_reverse geom

DAG: 
	- updated to use sqlsensor
	- updated to send alert message for broken readers and not fail the task

README
	- added sql examples in readme
	- added steps to cut lines with linesubstring

Things that needs to be fixed:
- readme is missing the explanation of detectors_history, currently in another md under update
- mohan.bt_segments_new is not ready to insert to bluetooth.segments, its missing a couple columns
- need to figure out which version of `detectors_history_corrected` to use then update function: broken reader and reader status history 
- sqls are only updated in github not in postgres
-

chmnata:
Looking to move the `detectors_history_*` to bluetooth schema, `mohan.detectors_history_corrected` is the corrected version, however there are some null values on several columns.  I will dig into how to populate those. 
e.g. some geoms are null and bt_id are null 
image

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,117 @@
...
+                     detector_list.dt
+
+         FROM        detector_list
+         LEFT JOIN   bluetooth.detectors_history_corrected ON detector_list.detector_name = detectors_history_corrected.read_name -- Should be changed
+         WHERE       detector_list.route_status = TRUE
+         GROUP BY    detector_list.route_status, detector_list.detector_name, detectors_history_corrected.reader_id, detector_list.dt)
+      
+      -- Combine active and deactive detector 
+      , detector_status as (
+         SELECT      DISTINCT detector_list.detector_name,
```
chmnata, line 83: should select distinct on reader_id instead of detector_name, because we are inserting reader_id's status at the end of the query and multiple detector_names can be refer to one reader_id
```
-- return 216 rows
select distinct reader_id, read_name from mohan.detectors_history_corrected

-- return 178 rows
select distinct reader_id from mohan.detectors_history_corrected

File: bluetooth/sql/functions/reader_status_history.sql [hunk 55]
```diff
@@ -0,0 +1,117 @@
...
+         SELECT      active.detector_name,
+                     active.last_reported,
+                     active.route_status,
+                     active.id,
+                     active.dt
+         FROM        active)
+   
+   INSERT INTO       bluetooth.reader_status_history (reader_id, last_active_date, active, dt)
+
+   SELECT            DISTINCT reader_id, max(last_reported), route_status, dt
```
chmnata, line 105: I don't think this select statement is necessary because the previous cte already remove possible dups

bluetooth/sql/functions/reader_status_history.sql
```diff
@@ -0,0 +1,117 @@
...
+   FROM              detector_status
+   WHERE             reader_id IS NOT NULL
```
chmnata, line 107: there is no null reader_id in the detector_history table

chmnata:
Working on the new segments tables and adding the missing columns: segment_name , bluetooth, wifi, duplicate, reverse, end_street
For bluetooth and wifi, according to https://github.com/CityofToronto/bdit_data-sources/blob/e98182d30de5511a2a4c246e4771f14c937c6b0f/bluetooth/sql/create_tables/segments.sql seems like bluetooth was assigned TRUE and wifi was assigned FALSE, but when I check the bluetooth.segments table there are values for wifi that are true 😕 
Also not sure what reversed means, reversed geometry (?)

radumas:
Just in case future spelunkers come here.
`wifi = TRUE` on highways where the problems associated with using WIFI observations are reduced.
`reversed` was a flag to identify segments had been drawn against the direction of travel and which had been subsequently reversed. I'm really not sure why it stuck around.
`duplicate` appears to have been to identify overlapping segments, which primarily exist on College. Since A LOT of those routes are offline, and some appear to have never been created, I am rather confused by the state of those detectors.

chmnata:
In bluetooth.routes_temp, there are 61 routes out of 402 with no data in `bluetooth.aggr_5min` and `bluetooth.observations`. 

image


But it does seems like some routes with data do have coverage on those routes with no data. (Green dotted lines are the ones without data and grey lines are the ones with data) 
image
image

Ones that still have coverage (or partial coverage):
- Adelaide from parliament to broadview 1495398, covered by 1454095
- Parliament from gerrard to calton 1496472, covered by 1455692 (Parliament from dundas to calton)
- College from ossington todufferin (1600090), sorta half covered by college from dovercourt to bathurst 1412888
- Yonge between Chatsworth and Castlefield (1460798, 1461616), covered by various short and long routes (1428295, 1448887, 1448891, 1448413, 1428324, 1448861)
- Sheppard from Morningside Ave to Nelson (end street was coded as sheppard incorrectly) 1460834, covered by VARIOUSSSSS short and long routes 5 in total

chmnata:
Checking the distinct status in reader_history table, `routepoint name` does not sounds like a status type (?), `added` is also ???. I guess technically according to this issue https://github.com/CityofToronto/bdit_data-sources/issues/196, this table doesn't need a status column 🤔 
- "duplicate"
- "routepoint name"
- "online"
- "offline"
- "decommissioned"
- "added"
- "moved"

The following readme explained the meaning behind each status, kindddd of feels like the below 3 is confusing or just keeping a log of duplicated data/data of human error. 
https://github.com/CityofToronto/bdit_data-sources/blob/btdag/bluetooth/update/reader_status_details.md

- added (these are the locations where names are inconsistent for example a reader with bt_id 4035 has a name E in one table and E4035 in another. Despite being at the same intersection at Gardiner & Parliament. Therefore, these two names could refer to the same reader. Thus to distinguish one got the status added and another is offline )

- duplicate (duplicate does not have a bt_id but based on its location and/or the reader name, there are more than one at a given intersection. The one that has got less details is assigned as duplicate)

- moved (moved has a bt_id but the same bt_id existed in a different location in the past. For example bt_id 5331 was at Gerrard & University. But the same bt_id also is at Lawrence & Mt Pleasant and is offline as per latest table. Thus 5331 has two entries with status offline and moved.

- routepoint name (does not have a bt_id. It is a name given to a point along a route. Could there be a detector at those points ? most likely not)

dags/bluetooth_check_readers.py
```diff
@@ -0,0 +1,128 @@
+from attr import NOTHING
```
chmnata, line 1: outdated, remove

bluetooth/api/blip_api.py
```diff
@@ -12,16 +9,20 @@
 import logging
 import os
 import sys
+#from syslog import LOG_DEBUG
```
chmnata, line 12: updated in more recent PRs. outdated, remove
https://github.com/CityofToronto/bdit_data-sources/pull/642
https://github.com/CityofToronto/bdit_data-sources/pull/480

bluetooth/sql/create_tables/add_new_segments.sql
```diff
@@ -0,0 +1,80 @@
+CREATE table mohan.bt_segments_new AS
```
chmnata, line 1: The segments in there that does not exist in `bluetooth.routes_temp` (when joining with `analysis_id`) are the following: 

```
'Bloor'	'WB'	'HighPark'	'Runnymede' 
'Bathurst'	'SB'	'Bloor'	'Harbord'
'Bloor'	'EB'	'Runnymede'	'HighPark'
'Queens'	'NB'	'Wellesley'	'Bloor
'Queens'	'SB'	'Bloor'	'Hoskin
'Spadina'	'NB'	'College'	'Harbord'
'Bathurst'	'NB'	'Harbord'	'Bloor'
'Spadina'	'SB'	'Harbord'	'College'
```
They are actually in `bluetooth.routes_temp`, just under a different analysis_id. I assume the one in `bluetooth.routes_temp` is correct because there is data in `bluetooth.aggr_temp` for those segments, and when searching for data with the analysis_id in mohan's table, no data returned.

```
-- returned none
SELECT * FROM bluetooth.aggr_5min where analysis_id in (1600060,
1600030,
1600070,
1600250,
1600220,
1600290,
1600010,
1600300)
```

i believe it is safe to delete this file

bluetooth/readersdown/offline_readers.csv
```diff
@@ -0,0 +1,22 @@
+reader_id,bt_id,past_status,current_status,name,int_id,project_name,geom
```
chmnata, line 1: @radumas do we even need this in the repo? should it be in a K drive or sth instead?

radumas: Aye could be in K `Data Sources\Bluetooth\Bliptrack\Inventory` and include the date it was committed in the filename

Issue #377, opened 2021-01-13
WYS validation - upload ATR count data


KatiRG:
Read in ATR data from xlsx files, save to csv and import to database.

PR #382, opened 2021-02-11
Wys ATR -  read data from files and output single table as CSV

KatiRG:

Creates a notebook `wys/atr/notebooks/tmc_data_preprocessing.ipynb` which performs the following:

- reads a set of xlsx files containing ATR counts recorded at 14 intersections
- formats data into single table with 4 columns: `datetime_bin`, `location`, `class_type`, `count` for each location and 6 class types (i.e. modes)
- saves to csv for upload to database
- performs QC checks on the table against the original files to verify that all rows have been read for each file
- plots the counts for each mode

## Issue(s) this solves:

- #377 

- confirm that final format of data is fine


wys/atr/_sql/create-table-wys_atr_kertcher.sql
```diff
@@ -0,0 +1,19 @@
+/*
+Creates table to store data from xlsx files in
+L:\TDCSB\PROJECT\DATA COLLECTION GROUP\3 - MANUAL COUNTS\2020 Manual Counts\WYS SIGN OBSERVANCE\Kertcher
+
+Files were read in and coverted to csv in notebook:
+wys/atr/notebooks/tmc_data_preprocessing.ipynb
+*/
+
+CREATE TABLE data_requests.wys_atr_kertcher
```
radumas, line 9: Why is this in `data_requests`? It should be in `data_analysis`.

In general I be wary of labelling objects with other employees' names. It implies an ownership that doesn't necessarily exist and which I don't find to be particularly helpful. In this case all she did was validate the manual counts, there's nothing particularly useful in labelling the table with her name. I would suggest `wys_validation_counts` instead.

radumas (changes requested):
Notebook is solid, well described. 
The folder might be better named `validation` and there should be a `readme` file explaining what the purpose of the folder is.

wys/atr/_sql/create-table-wys_atr_kertcher.sql
KatiRG: oh! i didn't know that was her name oops! I thought it was related to the study.

KatiRG:
Permission denied creating table in `data_analysis`:

```
ERROR:  permission denied for schema data_analysis
LINE 9: CREATE TABLE data_analysis.wys_validation_counts
```

```
CREATE TABLE data_analysis.wys_validation_counts
(
        datetime_bin timestamp without time zone,
        location text,
        class_type text,
        count numeric
);

ALTER TABLE data_analysis.wys_validation_counts
    OWNER to bdit_humans;
```

KatiRG:
> wys_validation_counts

data now imported to table !

aharpalaniTO:
Hi @KatiRG - looks good, well documented process. My only comment is the use of a new `atr_id` as opposed to just populating this table with `api_id` in the first place. Any reason for the former approach? I'm OK to merge as is without changing, but just double checking with you before I do so

KatiRG:
That's a good idea, the only issue is that there is no wys sign at `CHESTON RD S/B N OF BELSIZE DR`, which is `atr_id=14` .

PR #384, opened 2021-02-15
Remove former team membership from ownership

cczhu:

- Removes Rick, Joven and myself from `miovision_api`/`miovision_csv` table ownership. 😩

## Issue(s) this solves:

- 😩

- Need admin help to switch ownership of all elements of `miovision_api` and `miovision_csv` schemas to `miovision_admins`.
- Charles to put #380 into production once that's finished (currently Joven owns a table that need to be changed).
- We should consider making it a rule that any permanent table generated outside of one's personal schema must be owned by an admin group. (I suppose that means I'll need to create `collisions_admins` and `vz_admins` groups. For the next purge we should change all schema element ownership in `vz_analysis`, `vz_challenge`, `vz_safety_programs`, etc. to group ownership.

### DIFF
--- a/volumes/miovision/sql/csv_data/create-view-report_volumes_15min.sql
+++ b/volumes/miovision/sql/csv_data/create-view-report_volumes_15min.sql
@@ -33,4 +33,4 @@
 WITH DATA;
 
 ALTER TABLE miovision.report_volumes_15min
-  OWNER TO bdit_humans;
+  OWNER TO miovision_admins;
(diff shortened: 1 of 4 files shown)


volumes/miovision/sql/csv_data/create-view-report_volumes_15min.sql
```diff
@@ -33,4 +33,4 @@
 WITH DATA;
 
 ALTER TABLE miovision.report_volumes_15min
```
aharpalaniTO, line 35: this should be `miovision_csv.report_volumes_15min`

aharpalaniTO (changes requested):

aharpalaniTO:
I changed all the permissions via the admin account. One change requested and then I'll merge.

aharpalaniTO (approved):

cczhu:
Okay done.

Issue #388, opened 2021-04-07
WYS API includes NULL rows


radumas:
Since March 31st the API has been returning some rows with no speed or count. All API values are encoded as string, so they are implicitly converted to timestamps and numeric types, but empty strings fail that implicit conversion.

```python
[2021-04-06 18:59:17,606] {logging_mixin.py:112} INFO - [2021-04-06 18:59:17,606] {wys_api.py:297} CRITICAL - invalid input syntax for integer: ""
LINE 3: ...,'59','1'),(9632,'2021-03-31T03:19:00'::timestamp,'',''),(96...
```

We should ask the vendor what's up with these values, but we can also prevent the script from failing by ensuring that empty strings are treated as null. In the below assignment, this could be explicit with `int(i) if i else None` ([source](https://www.geeksforgeeks.org/python-empty-string-to-none-conversion/) [and](https://www.pythonpool.com/python-inline-if/)). 

https://github.com/CityofToronto/bdit_data-sources/blob/f64364c59307353a2f280fa573fdd6df62717e89/wys/api/python/wys_api.py#L153

KatiRG:
It this what you had in mind?

```
for item in counter:
    this_speed=int(item['speed']) if item['speed'] else None
    this_count=int(item['count']) if item['count'] else None
    speed_row=[api_id, datetime_bin, this_speed, this_count]
```

radumas:
yup ☺️

PR #389, opened 2021-04-07
catches null speed and count values #388

KatiRG:

- checks for null speed and count and only converts to `int` for non-NULL values in `parse_counts_for_location()` that takes in raw data records from the WYS api

## Issue(s) this solves:

- #388 

- pls review the fix


radumas (approved):
Works successfully. Thanks

PR #401, opened 2021-05-06
Fix Miovision Readme

cczhu:

- Fixes link bug in Miovision Readme
- Adds a note about deleting and re-pulling data

## Issue(s) this solves:

- None

- Review changes to Readme


radumas (approved):
Very important caveat.

We should probably create a function to simplify data deletion for repulling

Issue #367, opened 2020-12-07
WYS Sheets: Some wards not being inserted


radumas:
Wards: 5,8,14,20 are missing data.

I suspect this is because of the primary key constraint on `(installation_date, new_sign_number)`. Unfortunately nobody ever checked to see the data was actually successfully inserted. 

This is the log output for Ward 14

```
[2020-12-06 15:00:25,395] {logging_mixin.py:112} INFO - [2020-12-06 15:00:25,395] {wys_google_sheet.py:84} INFO - ward_14 does not have any row with different date format
[2020-12-06 15:00:25,396] {logging_mixin.py:112} INFO - [2020-12-06 15:00:25,396] {wys_google_sheet.py:91} INFO - Uploaded 232 rows to PostgreSQL for ward_14
[2020-12-06 15:00:25,404] {logging_mixin.py:112} INFO - [2020-12-06 15:00:25,404] {wys_google_sheet.py:100} ERROR - Duplicates of primary keys in ward_14 !!!
[2020-12-06 15:00:25,405] {logging_mixin.py:112} INFO - [2020-12-06 15:00:25,404] {wys_google_sheet.py:101} ERROR - ON CONFLICT DO UPDATE command cannot affect row a second time
HINT:  Ensure that no rows proposed for insertion within the same command have duplicate constrained values.
```

- [x] the successful message should happen after data is successfully inserted
- [x] `Duplicates of primary keys in ward_14 !!!` is hard coded for any error in insertion, which is not very helpful.

radumas:
Testing on Ward 14. 

```sql

ALTER TABLE wys.ward_14
DROP CONSTRAINT ward_14_pkey;
--INSERT script needs to have some constraint on the table.
ALTER TABLE wys.ward_14
ADD CONSTRAINT ward_14_pkey PRIMARY KEY (location, from_street, to_street, direction, installation_date, removal_date, new_sign_number)

```

Had to modify `dev` script to change the insert condition (I should've just removed both and then insert would have failed on the other tables but oh well... insert still fails on the other tables).

```python
    insert = sql.SQL('''INSERT INTO {}.{} (ward_no, location, from_street, to_street, direction, installation_date, removal_date,
                       new_sign_number, comments, confirmed) VALUES %s 
                       ON CONFLICT (location, from_street, to_street, direction, installation_date, removal_date, new_sign_number)
 DO UPDATE SET 
[...]

PR #404, opened 2021-05-31
Wys api mobile fixes: check for dupes in mobile sign installation table

KatiRG:

- checks for duplicate rows in table `wys.mobile_sign_installations_dupes` after data has been read in by task `read_google_sheets`
- if the table is not empty, the dag sends a Slack alert message

## Issue(s) this solves:

- #367 

- I have currently put the modifications in a temporary dag called `pull_wys_new.py`. Once the modifications are approved, the old `pull_wys.py` will be deleted and `pull_wys_new.py` will be renamed `pull_wys.py`, and the symbolic link in `/etc/airflow/dags/` to `pull_wys_new.py` will be deleted
- I am wondering about the `task_id` in the slack message. It is currently set at `context.get('task_instance').task_id == 'check_wys_mobile_dupes_id'`, but should it be `t2b`, i.e. the name of the task?
- You can verify that the tasks ran in the correct order (`read_google_sheets` -> `check_wys_mobile_dupes_id`) by  looking at the tree view in airflow for the `pull_wys_new` dag


dags/pull_wys_new.py
```diff
@@ -0,0 +1,115 @@
+"""
```
radumas, line 1: This shouldn't be a `_new` file, this should be the original pull WYS dag.

wys/api/python/wys_google_sheet.py
```diff
@@ -83,12 +83,56 @@ def pull_from_sheet(con, service, dict_table, ward, *args):
...
-    LOGGER.info('Uploaded %s rows to PostgreSQL for %s', len(rows), table_name)
+    insert = sql.SQL('''
+        WITH new_data (ward_no, location, from_street, to_street, direction, 
+                    installation_date, removal_date, new_sign_number, comments, 
+                    confirmed) 
+                    AS (
+                VALUES %s) 
+        , dupes AS(
+            /*Identify rows that will violate the unique constraint in the 
+            upsert*/
```
radumas, line 94: This could be more explicit: Identify as duplicates rows where the same sign is installed on the same date and insert into a `duplicates` table

radumas:
CodeFactor is spitting an error for "No exception type(s) specified". If `strptime` spits out a specific exception that should be caught and a separate generic one could also be logged.

https://github.com/CityofToronto/bdit_data-sources/blob/b082620bb7c298039101ef2c6cc8009b5afb03f1/wys/api/python/wys_google_sheet.py#L68-L75

wys/api/python/wys_google_sheet.py
```diff
@@ -83,12 +83,56 @@ def pull_from_sheet(con, service, dict_table, ward, *args):
...
+            INSERT INTO wys.mobile_sign_installations_dupes
+            SELECT ward_no::INT, location, from_street, to_street, direction, 
+            installation_date, removal_date, new_sign_number, comments
+            FROM new_data
+            NATURAL JOIN (SELECT new_sign_number, installation_date
+                    FROM new_data
+                    GROUP BY new_sign_number, installation_date
+                    HAVING COUNT(*)> 1) dupes
```
radumas, line 102: Beyond duplicates of `new_sign_number, installation_date` I think we could identify where `installation_date` and `removal_date` overlap for a same `new_sign_number`

dags/pull_wys_new.py
```diff
@@ -0,0 +1,115 @@
...
+                 'retries': 0,
+                 'retry_delay': timedelta(minutes=5),
+                 'on_failure_callback': task_fail_slack_alert
+                }
+
+dag = DAG('pull_wys_new',default_args=default_args, schedule_interval='0 15 * * *')
+# Run at 3 PM local time every day
+
+with wys_postgres.get_conn() as con:
+    t1 = PythonOperator(
```
radumas, line 85: In general I think task variables should be the same as their `task_id` to make it easier to differentiate logic further down.

dags/pull_wys_new.py
```diff
@@ -0,0 +1,115 @@
...
+    from wys_google_sheet import read_masterlist
+except:
+    raise ImportError("Cannot import functions to pull watch your speed data")
+
+
+SLACK_CONN_ID = 'slack'
+def task_fail_slack_alert(context):
+    slack_webhook_token = BaseHook.get_connection(SLACK_CONN_ID).password
+
+    if context.get('task_instance').task_id == 't1':
```
radumas, line 32: task_id is incorrect

dags/pull_wys_new.py
```diff
@@ -0,0 +1,115 @@
...
+            )
+
+# task2b will fail when there are a non-zero number of rows in the table
+# and this will trigger a slack alert message
+t2b = CheckOperator(
+    task_id='check_wys_mobile_dupes_id',
+    sql='''SELECT COUNT(1) = 0 FROM wys.mobile_sign_installations_dupes''',
+    conn_id='wys_bot',
+    dag=dag
+)
```
radumas, line 109: Super simple check! this is dope.

radumas (changes requested):
Appears to work. Some improvements noted.

The process to identify duplicates should be documented in the README.

tahaislam:
A new PR (#443) to resolve the issue of failing to pull data older than 2 days (#424). As pulling older data requires longer time (full days take more than 29 seconds), wys_api-py was modified to pull 1-hour data chunks instead of full days

Issue #414, opened 2021-06-22
Miovision - ensure proper duplicate handling for gaps and reports tables


cczhu:
@radumas identified that elements of the Miovision aggregation pipeline are currently not robust to duplicate handling. While the 15-minute ATR and TMC aggregation tables are assisted by foreign key mappings and a `processed` column, respectively, the `find_gaps` function will fail loudly because it doesn't have an `ON CONFLICT DO NOTHING` line below [this one](https://github.com/CityofToronto/bdit_data-sources/blob/b210006db5f34558785d019fcf26ded4057ede53/volumes/miovision/sql/function-find_gaps.sql#L104), and `report_dates` will simply insert duplicates because there's no unique constraint in the [function definition](https://github.com/CityofToronto/bdit_data-sources/blob/b210006db5f34558785d019fcf26ded4057ede53/volumes/miovision/sql/create-table-report_dates.sql#L7) and no corresponding `ON CONFLICT DO NOTHING` [at the end of the function](https://github.com/CityofToronto/bdit_data-sources/blob/b210006db5f34558785d019fcf26ded4057ede53/volumes/miovision/sql/function-report_dates.sql#L23).

Tasks:
- Delete and regenerate `report_dates` table to have a unique constraint
- Include `ON CONFLICT` clauses in `find_gaps` and `report_dates`.

Issue #402, opened 2021-05-10
Miovision Add New Intersections


cczhu:
New Miovision cameras have been installed at

```
1) Steeles / Jane
2) Sheppard / Jane
3) Lawrence / Jane
4) Eglinton / Jane
5) Sheppard / Keele
6) Sheppard / Weston
7) Bayview / River
8) Yonge / Davenport / Church
9) Jane and Wilson
10) St George / Harbord / Hoskin
11) Bloor / Dufferin - installed last year
```

and over the course of the week of 2021-05-10 their data feeds will come online. We need to add them to `miovision_api`.

cczhu:
Going forward with adding:
- Bloor Street West and Dufferin Street
- Eglinton Avenue West and Jane   Street
- Harbord Street/St. George   Street/Hoskin Street
- Jane Street and Lawrence   Avenue West
- Jane Street and Wilson Avenue
- Sheppard Avenue West and   Weston Road
- Yonge Street / Davenport Road   / Church St

as these stations are now online, and we're not certain about the timeline for the remaining few.


Query to get PX info:

```sql
WITH wanted_px AS (
	SELECT * FROM (VALUES (1, 325), (2, 905), (3, 838),
				          (4, 430), (5, 532), (6, 588),
				          (7, 41)) AS t (uid, px)
)
SELECT *
FROM wanted_px a
LEFT JOIN gis.traffic_signal b ON a.px = b.px::integer
ORDER BY uid
```

Uploaded using the Python script in the [Readme](https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/miovision#Adding-Intersections).

cczhu:
Fetched 4 days of raw data using:

```
python3 intersection_tmc.py run-api --path /etc/airflow/data_scripts/volumes/miovision/api/config.cfg --start_date=2021-06-18 --end_date=2021-06-20 --intersection=58 --intersection=59 --intersection=60 --intersection=61 --intersection=62 --intersection=63 --intersection=64 --pull
 2012  python3 intersection_tmc.py run-api --path /etc/airflow/data_scripts/volumes/miovision/api/config.cfg --start_date=2021-06-20 --end_date=2021-06-22 --intersection=58 --intersection=59 --intersection=60 --intersection=61 --intersection=62 --intersection=63 --intersection=64 --pull
```

cczhu:
Airflow auto-ran stuff yesterday. Some QCing:

## Check 15-minute Aggregation

[...]

PR #415, opened 2021-06-22
Enable duplicate control for unacceptable_gaps and report_dates (#414).

cczhu:

- Enables duplicate handing for `miovision_api.unacceptable_gaps` and `miovision_api.report_dates`.

## Issue(s) this solves:

- #414

- Quick check of the code to confirm #414 is solved.

Once merged:
- Delete `miovision_api.report_dates_donotuse`
- Change name of `report_dates` function to `get_report_dates` on Postgres
- Switch whatever version of `intersection_tmc.py` Airflow is using to the new version - *must be done on the same day as renaming the function on Postgres*

These functions and have already been deployed in service of #402.


radumas (approved):
Looks good. Thanks for standing on guard 💂🏻 against duplicate data :meow_salute:

PR #421, opened 2021-07-06
fix to allow text_to_centreline() to handle capitalized text

KatiRG:

Currently, `gis.text_to_centreline()` fails if the input text is capitalized, e.g.:
```
SELECT *
FROM gis.text_to_centreline(111,'BLOOR ST W', 'BETWEEN SHAW ST AND OSSINGTON AVE', NULL);
```
The problem has to do with using  `SPLIT_PART` on fixed lower-case text `and` in the function `function-clean_bylaws_text.sql` which  `gis.text_to_centreline()` calls. The 'Between' is also only checked for first letter capitalization.

This PR adds a replace for any possible AND or BETWEEN with the lower-case before `SPLIT_PART` in `function-clean_bylaws_text.sql`.

- pls confirm that these fixes work


radumas (changes requested):
Thanks for attempting this, I think the code could be simplified by using the `i` flag in `regexp_replace()`

>The regexp_replace function provides substitution of new text for substrings that match POSIX regular expression patterns. It has the syntax regexp_replace(source, pattern, replacement [, flags ]). The source string is returned unchanged if there is no match to the pattern. If there is a match, the source string is returned with the replacement string substituted for the matching substring. The replacement string can contain \n, where n is 1 through 9, to indicate that the source substring matching the n'th parenthesized subexpression of the pattern should be inserted, and it can contain \& to indicate that the substring matching the entire pattern should be inserted. Write \\ if you need to put a literal backslash in the replacement text. The flags parameter is an optional text string containing zero or more single-letter flags that change the function's behavior. Flag i specifies case-insensitive matching, while flag g specifies replacement of each matching substring rather than only the first one. Supported flags (though not g) are described in Table 9-22.

See the paragraph following the `regexp_replace` examples 
https://devdocs.io/postgresql~9.6/functions-matching#FUNCTIONS-POSIX-REGEXP

radumas:
Closing and sending to open issue #628

PR #440, opened 2021-12-31
Automating EOY maintenance 🎐 

### LINKED ISSUE #439: Automate end of year maintenance
We have partitioned tables that are monthly/annually. We need to create new ones for the new year and update triggers etc. Procedure in #251, hopefully as much as possible can be made into a DAG.

- [x] Here traffic analytics
- [x] Bluetooth
- [x] Miovision
- [x] Holidays table
- [x] WYS table

radumas:

- automates the creation of new child tables in the new year
   - HERE
   - Miovision
   - WYS
   - Bluetooth
- automates the replacement of triggers for those new children
- automates updating the holiday table

## Issue(s) this solves:

- Closes #439 

- 😎 
- Where do you think stuff should be documented?


radumas:
Ittttt would be nice to have a Slack message notifying of success of all the tasks so someone can review that tables were properly created.

chmnata:
> Ittttt would be nice to have a Slack message notifying of success of all the tasks so someone can review that tables were properly created.

Can be done by adding a send alert [trigger task](https://airflow.apache.org/docs/apache-airflow/1.10.5/concepts.html?highlight=trigger%20rule#trigger-rules) that only runs if all the above tasks ran successfully. Gonna quickly create that :meow_salute:

chmnata:
Logic looks like dis meow
image

radumas:
This is looking good. Wondering if the HERE trigger task should use the same logic from Bluetooth or Miovision to just replace the trigger instead of sending the slack message. 
Also where should this be documented?

chmnata:
Was gonna merge this but for some reason this dag ran today again. It is only suppose to run once every year but it ran three times this year already. 
This is the schedule which is correct so im confused as to why it was scheduled today and yesterday as well
image

radumas:
lol uh oh

radumas:
seems like it's "Mondays and every day between 14-21 in December"

chmnata (approved):
All data sources' tasks have been checked or renewed. Added congestion task and changed the scheduled time to run once every year on December 1st. Future improvement can include changing the schedule to run on the 1st weekday of December using timetables. Can merge meow.

radumas:
can you deal with the CodeFactor `Except` issues?
I think catch `except Exception:` and then do a string concatenation with it and the custom error messages

chmnata (approved):
Ready for merging

chmnata (approved):

Issue #424, opened 2021-07-13
WYS: Can't pull historical data more than 2 days old using Airflow dag


radumas:
Occasionally the DAG fails, when we try to rerun it then we get a bunch of 504 Errors.

This is what the vendor has told us about what is happening.

>Could you tell what route are you trying to run was it with parameters from/to https://api.streetsoncloud.com/sv2/signs/statistics/location/lcoation_id/date/<date>/from/<from>/to/<to> 
>
> Or was it  just a date https://api.streetsoncloud.com/sv2/signs/statistics/location/location_id/date/<date> 
> 
> If you are trying to fetch data from June 28th  it means the data  is stored in the archived storage already, all data older than 2 days old moved  into archive. Retrieving data from archived storage for a whole day usually takes more than 29 sec the  which means most likely you will receive error 504.
> 
> So our suggestion for all statistics requests  for more than 2 days old data to use the api route with parameters FROM/TO with smaller intervals, ideally by hours, for example https://api.streetsoncloud.com/sv2/signs/statistics/location/<location_id>/date/2021-06-28/from/08:00/to/09:00 
 
Someone needs to test this for the two days we are missing: May 31st and June 28th, 2020

radumas:
Deleted the following days of data with the below query. Not sure why there's a foreign key column between raw data and aggregate data but there's no explicit foreign key relationship.

```sql
WITH agg_delete as (
	DELETE FROM wys.speed_counts_agg_5kph
	WHERE 
	(datetime_bin >= '2021-07-26' AND datetime_bin < '2021-07-26'::DATE + INterval '1 day' ) OR
	(datetime_bin >= '2021-07-28' AND datetime_bin < '2021-07-28'::DATE + INterval '1 day' ) OR
	(datetime_bin >= '2021-07-29' AND datetime_bin < '2021-07-29'::DATE + INterval '1 day' ) OR
	(datetime_bin >= '2021-08-04' AND datetime_bin < '2021-08-04'::DATE + INterval '1 day' ) 
	RETURNING speed_counts_agg_5kph_id)
DELETE FROM wys.raw_data_2021 raw_data
USING agg_delete
WHERE raw_data.speed_count_uid = agg_delete.speed_counts_agg_5kph_id
```

This should get written into the README for wys, or alternatively there could be a "replace data" command written into the script that would use the call written above and also delete data for the days to be re-pulled.

scann0n:
[...]

PR #443, opened 2022-02-01
#424 fixing historical data pulling fails

tahaislam:

- The WYS API script can now pull data older than 2 days, which used to fail.
- As pulling older data requires longer time (full days take more than 29 seconds), the script was modified to pull 1-hour data chunks instead of full days

## Issue(s) this solves:

- #424 

- needs to review the function get_statistics in wys/api/python/wys_api.py to make sure the script can pull historical data efficiently


tahaislam:
Added two functions to update the tables: wys.stationary_signs and wys.mobile_api_id. These tables are still materialized views and need to be converted into tables. The new functions will update these tables in future (within the monthly DAG) taking into consideration the current month and start_date/installation_date.
Also, added a script to update wys.locations so that the earliest start_date for each api_id is the first date on which we have data #444

wys/api/python/wys_api.py
```diff
@@ -357,45 +401,59 @@ def update_locations(conn, loc_table):
...
-                SELECT a.api_id, a.address, a.sign_name, a.dir, start_date, 
-                       a.loc
+                SELECT a.api_id, a.address, a.sign_name, a.dir, a.start_date, 
+                       a.loc, a.geom
                 FROM daily_intersections A
                 LEFT JOIN locations B ON A.api_id = B.api_id
                                       AND B.loc = A.loc
                 WHERE B.loc IS NULL
+                      AND (st_distance_sphere(A.geom, B.geom) > 100 
+                           OR A.dir <> B.dir)
```
radumas, line 452: I don't think this distance comparison will happen because of the anti-join. 
Would need to be 

```sql
LEFT JOIN locations B ON A.api_id = B.api_id
                                     
                WHERE B.loc <> A.loc
                      AND (st_distance_sphere(A.geom, B.geom) > 100 
                           OR A.dir <> B.dir)

**Review by radumas — CHANGES_REQUESTED** [review]
The hourly pull looks good.

One fix in the API that I think is a bug in the SQL.

Can you migrate the monthly functions to the monthly branch with the other open data fixes with a `git cherry-pick`? That way this PR will only pertain to the daily ops of the API pipeline.

So steps for implementation:

1. backup the locations table
2. run the `start_date` query to update the locations table
3. add the `geom` column
4. merge changes into master and then pull down to the production Airflow.

Am I missing anything?

File: wys/api/python/wys_api.py (hunk 1 above)
Comment (tahaislam) [reply wys/api/python/wys_api.py]: By monthly functions, do you mean `update_mobile_api_id` and `update_stationary_signs`?

Comment (tahaislam) [reply wys/api/python/wys_api.py]: More missing dates to pull after merge:
* 2022-03-29
* 2022-03-31

File: wys/api/python/wys_api.py [hunk 2]
```diff
@@ -424,37 +424,35 @@ def update_locations(conn, loc_table):
                        a.loc, a.geom 
                 FROM daily_intersections A
                 LEFT JOIN locations B ON A.api_id = B.api_id
```
radumas, line 426: I'm pretty sure this won't account for new signs, only moved ones. 
There needs to be an `OR api_id NOT IN (SELECT api_id FROM locations) added here

radumas (changes requested):
Good update. I think there's one minor problem with the inserting new signs logic.

Also reflecting that we should probably transform the geometry to 2952 and store it as that so that the spatial index is actually used in future queries

radumas (approved):

Issue #423, opened 2021-07-12
Update READMEs to reflect bike approach volume processing 


radumas:
[Most references to `volumes_15min_tmc`](https://github.com/search?q=%22volumes_15min_tmc%22+repo%3ACityofToronto%2Fbdit_data-sources+filename%3AREADME.md&type=Code&ref=advsearch&l=&l=) need to be changed to `volumes_15min_mvt` and then some extra documentation added.

Also the giant warning should be deleted.

radumas:
Also new classifications from bike approaches.

Also documentation on bike approaches should be explicit there are entrance and exit volumes.

https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/miovision/README.md?plain=1#L73-L75

PR #446, opened 2022-02-09
Mio doc updates

scann0n:

- Edits to volumes/miovision/readme.md to incorporate bicycle approaches
- Other bonus features like an ERD, updated mio-map, more table descriptions, less outdated info and more!!!

## Issue(s) this solves:

- #423

- Are things explained correctly? Should anything else be added?


radumas (commented):
Mildly annoying that I can't comment on past work.
A borked link snuck into the TOC

https://github.com/CityofToronto/bdit_data-sources/blob/f72ab2a30a60ea90897dae41a2d9dbd9a00ac440/volumes/miovision/README.md?plain=1#L20

This should actually be `volumes_15min_mvt`

tahaislam:
I got it from the documentation that all tables and views are in miovision_api schema. But it wasn't clear in the first paragraph. And as there are two other miovision schemas in BigData DB (miovision_csv & miovision_new). Are they relevant to miovision?
How did you draw this ERD? Is there an automatic way to generate it? I'm interested in learning that
Why NULLing the whole hour if only a single 15-min interval is NULL? Is there a reason or for easier/feasible implementation?
Why does unacceptable_gaps have a field "accept"? does it contain all gaps or just unaccepted ones?

radumas:
Solid questions Islam!

> I got it from the documentation that all tables and views are in miovision_api schema. But this wasn't clear in the first paragraph. And as there are two other miovision schemas in BigData DB (miovision_csv & miovision_new). Are they relevant to miovision? 

`miovision_new`: was created for testing something and should be dealt with in a subsequent purge.
`miovision_csv`: I need to look into this more (upcoming purge) but this definitely contains the same data as the `api` schema for a particular period of time because the API data was unreliable so we got a huge dump of data from Miovision via csv. That is definitely duplicated across the schema and needs to be purged. For the King Street pilot we were also getting csv data before the API was up and running. I forget what span of time those cover and I'm not 100% sure where it is.


> How did you draw this ERD? Is there an automatic way to generate it? 

I think this is what Sarah was griping about in pgAdmin.

>I'm interested in learning that Why NULLing the whole hour if only a single 15-min interval is NULL? Is there a reason or for easier/feasible implementation?

The idea was that a typical workflow with the data is summing the data by hour before averaging or aggregating further, so wanted to ensure that low hourly volumes didn't accidentally get generated without someone checking the unacceptable gaps table. 

> Why does unacceptable_gaps have a field "accept"? does it contain all gaps or just unaccepted ones?

I'm going to hypothesize that, in order to validate the process, any gap is identified in the table, and then based on the dynamic volume algorithm it is labelled as acceptable (due to natural low volume) or unacceptable (the camera probably didn't record data).

scann0n:
My two cents in blue below…

tahaislam:
Thanks, Raphael and Sarah. Re the ERD, I guess Sarah's last week lightning talk answered that

volumes/miovision/Archive.md
```diff
@@ -0,0 +1,374 @@
...
+DELETE FROM miovision_api.unacceptable_gaps
+WHERE gap_start BETWEEN '2020-08-20 00:00:00' AND '2020-08-20 23:59:00';
+```
+
+Once you have deleted all the relevant data, you can now re-run the process with the following command line.
+```
+python3 intersection_tmc.py run-api --path /etc/airflow/data_scripts/volumes/miovision/api/config.cfg --start_date 2020-08-20 --end_date 2020-08-21
+```
+
+The data pulling script currently *does not support* deleting and re-processing data that is not in one-day blocks (for example we cannot delete and re-pull data from `'2021-05-01 16:00:00'` to `'2021-05-02 23:59:00'`, instead we must do so from `'2021-05-01 00:00:00'` to `'2021-05-02 23:59:00'`).
```
radumas, line 249: I think this might still be relevant

radumas (changes requested):
Thanks for this extensive rewrite, a few smol change requests

radumas:
oh I tried adding a comment on the geojson that I think it could replace the old geojson so there aren't two geojsons.

radumas (approved):
Noice!

PR #448, opened 2022-02-11
Wys open data fixes

### LINKED ISSUE #390: WYS: mobile_api_id MATERIALIZED VIEW doesn't account for signs changing over time
Not a priority but the `wys_monthly_summary` dag will fail the first task `wys_views` which refreshes the `mobile_api_id` view which matches mobile sign installations from the spreadsheets with the locations (and `api_id`) pulled from the API because it doesn't account for `start_date`: when that `api_id` first appears in the API.

### LINKED ISSUE #435: WYS Schedules API Fails Silently
New stationary signs don't get updated to Open Data because of missing sign schedule (maybe this should be left joined)

I tested this in the browser and got a 200 status but the json response was `{"message":"Forbidden"}`.

This should be validated in the code

https://github.com/CityofToronto/bdit_data-sources/blob/e98182d30de5511a2a4c246e4771f14c937c6b0f/wys/api/python/wys_api.py#L407

This affects 94 signs 🤦🏻‍♂️  

```sql
SELECT api_id, sign_id, address, sign_name, dir, start_date, serial_num, next_start, prev_start, geom
	FROM wys.stationary_signs
	LEFT JOIN wys.sign_schedules_list USING (api_id)
	 	WHERE schedule_name IS NULL 
```

Will email vendor

radumas:

- 
- 

## Issue(s) this solves:

- Works on #390 
- Closes #435 

-


tahaislam:
This branch was merged into #536 and then into `master`

PR #452, opened 2022-02-17
Add log cleanup dag

chmnata:

- add log cleanup dag to master

## Issue(s) this solves:

- There was no issue 

- there is actually no slack alerts for this dag, should we add it? @radumas


radumas:
yeah maybe there should be a slack alert on failure....

chmnata:
hmmmmmmmmmmmmmm this is not in dev_script :meow_dio:

chmnata:
added slack alert and change the file to symlink to dev_script, if this doesn't fail tmr I think we can merge @radumas

tahaislam (approved):

Issue #441, opened 2022-01-27
Automate updating of the collision data within the RDS


radumas:
The MOVE replicator is now kindly sending us collision data to `collisions_replicator.ACC`. However this is `DROP`ped every refresh so we can't have `MATERIALIZED VIEW`s depending on it. Peter has suggested we validate the data before transferring it to a safe copy table currently `collisions_replicator.acc_safe_copy`

https://github.com/CityofToronto/bdit_move_etl/blob/master/scripts/airflow/dags/taskflow_dag.py
the task group is `replicator_update_schema`
the files: https://github.com/CityofToronto/bdit_move_etl/tree/master/scripts/airflow/tasks/replicator_update_schema

Also the old details on transferring data need to be deleted and the documentation updated.

scann0n:
The last feasible (aka not in the future) date for which a collision was recorded in the collisions_replicator database is Jan 21 2022. Suggested next steps: set up an airflow dag (or piggy back on an existing one) to update the collisions_replicator.acc table with a drop cascade, then rebuild the collision_no, events and involved mat views when the acc table is updated.

Does that make sense?
 
Still need to discuss: 
1. The "home" schema - will it still be collisions_replicator? If so, there's no conflict with a VZ mat view (vz_analysis.hin_ksis_network)...
2. What existing dag to build this into (the dag referenced above does not refer to collisions_replicator and the existing .py file used to pull acc data into collisions looks old (and it doesn't use Airflow, so...)

PR #464, opened 2022-05-24
Collisions pipe

scann0n:

- adds a .py file that will 1) update acc_safe_copy at 3am every day, 2) refresh the three mat views 3) have Slack squeak if it doesn't work

## Issue(s) this solves:

- #441 (I forgot this in some commits (sorry))

- Did I DAG right??? It's my first time...


dags/collisions_asc_mv.py
```diff
@@ -0,0 +1,69 @@
...
+                'email': ['sarah.cannon@toronto.ca'],
+                'email_on_failure': False,
+                 'email_on_success': False,
+                 'retries': 0,
+                 'retry_delay': timedelta(minutes=5),
+                 'on_failure_callback': task_fail_sassy_slack_alert
+                }
+
+
+with DAG('BDITTO_ACC_REPLICATOR', # going waaaaaayyyyy out on a limb of the magical assumption tree here....
```
radumas, line 46: I appreciate I haven't documented this, and alllll the other dags are bad examples of this, but I would like dags to be named in the format: `dataset_action` (also all lower case plz), so a good candidate here would be `collisions_replicator_transfer` or some such.

scann0n:
The mat views are refreshed with a `refresh materialized view concurrently` option. The `concurrently` option requires a unique field or index. I created a unique index using `collision_no` for the `collision_no` and `events` mat views. I'm puzzled as to what I should do for `involved`.

I'm trying to find a unique field (or combination of fields) in the `involved` mat view so that I can create a unique index to meet the requirements of the `concurrently` option. I tried using the `collision_no` and `person_no` fields together, but the distinct count is less than the non-distinct count (which means that there are non-unique values ( :meow_sad: )).

Some options include:

1. Dropping the `concurrently` option, which I think means that people will not be able to query the mat view while it is being updated and it might take longer (which may not be a big deal at all at 3am, which is when it's scheduled to run)
2. Adding a unique field using `row_number()`

Other options??? Comments???

dags/collisions_asc_mv.py
scann0n: Wait did you mean that the dag name 'BDITTO_ACC_REPLICATOR' should be changed? Because I changed the name of the .py file to collisions_replicatory_transfer.py and maybe that was a misunderstanding?

tahaislam: I think Raph meant changing the DAG name. This name is what we see in Airflow UI. 
Also, the file name now is more meaningful

tahaislam (approved):

tahaislam (approved):

Issue #465, opened 2022-06-03
Add retry to the HERE pipeline


radumas:
The HERE pipeline occasionally fails due to a timeout (https://github.com/CityofToronto/bdit_data-sources/issues/204) or the data not being available yet. The typical fix is to retry after waiting an hour or so. We should add that to the DAG.

chmnata:
Was trying to see if there is a call_back or parameter to send a message if the retry was successful, but nah. The [callbacks ](https://airflow.apache.org/docs/apache-airflow/2.2.1/logging-monitoring/callbacks.html#callback-types) they have:
Name | Description
-- | --
on_success_callback | Invoked when the task succeeds
on_failure_callback | Invoked when the task fails
sla_miss_callback | Invoked when a task misses its defined SLA
on_retry_callback | Invoked when the task is up for retry

I don't think the on_retry_callback will be necessary as it will just send a msg everytime the dag is retying.

I think it might be possible by using [task instances ](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/taskinstance/index.html#airflow.models.taskinstance.TaskInstanceKey) cause there is a `try_number` and `with_try_number` param which looks useful. I'm gonna leave this till next time cause seems like there's no available callback to quickly implement it.

chmnata:
![image](https://user-images.githubusercontent.com/46324452/172488933-28301171-85d2-4fc1-9c39-99fcdbd5f899.png)
uhhhhhhhhhhhhh it didn't work. When a tasks fail it should have a status `up_for_retry` instead of just `fail`.

PR #467, opened 2022-06-06
#465 Add retry to pull here DAG

chmnata:

- Add `retry` to pull_here DAG

## Issue(s) this solves:

- #465

- the two lines I changed in the DAG
- If a 45 min interval + 3 times retry is good

### DIFF
--- a/dags/pull_here.py
+++ b/dags/pull_here.py
@@ -41,8 +41,8 @@ default_args = {'owner':'rdumas',
                 'email': ['raphael.dumas@toronto.ca'],
                 'email_on_failure': False,
                  'email_on_success': False,
-                 'retries': 0,
-                 'retry_delay': timedelta(minutes=5),
+                 'retries': 3, #Retry 3 times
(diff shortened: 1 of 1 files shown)


radumas (changes requested):
This looks good.

My one suggestion is to include  `retry_exponential_backoff = True`. 

> `retry_exponential_backoff` ([bool](https://docs.python.org/3/library/functions.html#bool)) -- allow progressive longer waits between retries by using exponential backoff algorithm on retry delay (delay will be converted into seconds)

My understanding is then the retries will be after 45 minutes, 90 minutes, 3 hours.

Maybe the retry delay should be something like 2 hours instead? How often do you feel it work again after retrying immediately?

chmnata:
Usually when it fails at 16:30ish, I am able to rerun it at around 17 or so, if not 17:30. How about 1 hour?

radumas:
Deal!

radumas (approved):

Issue #472, opened 2022-06-08
VZ School Sheets: add 2022


radumas:
If you see improvements that could be made to the process so it is more resilient to user input errors... feel free to try to make them, e.g. https://github.com/CityofToronto/bdit_data-sources/issues/438

Issue #438, opened 2021-12-14
Make school zone pipeline resilient to text errors in the google sheet


radumas:
I think the ideal way is to have a PostgreSQL function for inserting data that reports errors and does upserts on the table and then loop over the sheet and send a Slack alert if an error occurred with a link to the log printout of all the erroneous rows.

qu-xiang:
We have found that there are instances where:

- Same school_name can appear multiple times in the same year's table
- Same school_name can appear across different years' tables, but with different information in other fields

Sample SQL code, pull data from 2018/2019 tables
`WITH all_data AS (
    SELECT *, 2018 AS year FROM vz_safety_programs_staging.school_safety_zone_2018_raw
    UNION ALL
    SELECT *, 2019 AS year FROM vz_safety_programs_staging.school_safety_zone_2019_raw)
SELECT * FROM all_data
WHERE school_name IN (
    SELECT school_name FROM all_data
    GROUP BY school_name
    HAVING COUNT(*) > 1)
ORDER BY school_name;`

tahaislam:
After testing the new validation rules (applied on columns: `locations_zone`, `inal_sign_installation`, `locations_fb`, `locations_wyss`), we found some outliers in previous years:

- 2018:
  - extra space preceding a decimal point in some coordinates _(can be fixed in our code if needed)_
    "F H MILLER JUNIOR PUBLIC SCHOOL; ST. NICOLAS OF BARI ELEMENTARY CATHOLIC SCHOOL"	"363 Rogers Rd;300 Caledonia Rd"	"9237848"		"43.683954, -79.458875; 43.682889, -79.460556"	"March 8, 2019"	"43.683016, -79.462317; 43.683212, -79.458610; 43.683297, -79.458100; 43.683431, -79.459481; 43.684938, -79.460248; **43 .685823**, -79.459249"	"43.684190, -79.458426"
- 2020:
  - missing decimal point in some coordinates _(can be fixed in our code if needed)_
    "ST BEDE CATHOLIC ELEMTARY SCHOOL"	"521 Sewell's Rd"	"10026267"	"10026269"	"43.8179875235, -79.2065508937"	"February 24, 2021"	"43.81745184469687, -79.20872942654107; 43.81873, **-79206048**"	"43.81768834878893, -79.20772756633532"
  - school outside Toronto (Markham)
[...]

PR #473, opened 2022-06-10
Added code for pulling 2022 Google sheet SSZ

qu-xiang:

- Added code for pulling 2022 Google sheet SSZ
- Added validation function to check data type for each field

## Issue(s) this solves:

- #472 
- #438 

- New added codes regarding year 2022 
- Validation functions and how they work in various inputs
- Whether the whole DAG can run correctly


vision_zero/schools.py
```diff
@@ -17,12 +17,19 @@
 from psycopg2 import sql
 import logging 
 
+import re
```
radumas, line 20: This doesn't look to be used anymore

vision_zero/schools.py
```diff
@@ -17,12 +17,19 @@
...
+import geopandas as gpd
+from shapely.geometry import Polygon, LineString, Point
+
 from airflow.models import Variable
 dag_config = Variable.get('ssz_spreadsheet_ids', deserialize_json=True)
 ssz2018 = dag_config['ssz2018']
 ssz2019 = dag_config['ssz2019']
 ssz2020 = dag_config['ssz2020']
 ssz2021 = dag_config['ssz2021']
+ssz2022 = dag_config['ssz2022']
```
radumas, line 32: this should get called in the DAG.

vision_zero/schools.py
```diff
@@ -143,7 +109,7 @@ def within_toronto(con, coords):
             pairs = [tuple(map(float, i.split(','))) for i in pairs_str]
             
             to_boundary_sql = "SELECT * FROM gis.toronto_boundary"
-            to_boundary = gpd.GeoDataFrame.from_postgis(to_boundary_sql, con)
+            to_boundary = gpd.GeoDataFrame.from_postgis(to_boundary_sql, con).to_crs(epsg=3348).buffer(50).to_crs(epsg=4326)
```
radumas, line 112: I think this should be 2952 for Toronto

radumas (changes requested):
This looks great. One minor comment about the SRID to create a buffer for the boundary

radumas (commented):
I was on the veeeeeeeeeeeerge of approving but then I wondered if this should follow the folder structure we discussed wrt traffic lights pipeline

radumas (approved):
:meow_praise:

Issue #462, opened 2022-05-10
Automate transfer of count data from `"TRAFFIC_NEW"` to the `traffic` schema


radumas:
Now that the `replicator_bot` is refreshing data in `"TRAFFIC_NEW"` on a nightly basis, we need to automate putting the data into our `traffic` schema.

This process can be automated by performing some comparison of the number of rows in the `"TRAFFIC_NEW"` tables compared to `traffic` and then copying over the data, like with the collisions pipeline

scann0n:
Getting started questions...

1. Does `"TRAFFIC_NEW"` = `traffic_new`?
2. I can haz access to `traffic_new`?
3. Is the general idea to update `traffic.arc_link` with `traffic_new."ARC_LINK"`?

radumas:
`traffic_new` no longer exists, I think that resolves most of your questions :)

scann0n:
Why yes indeed - 
Silly me...
"TRAFFIC_NEW"
Isn't next to "u"...
It's all the way
Ahead of "a"
And question 2
I must redo...
I can haz access to "TRAFFIC_NEW"?

scann0n:
Are all collisions_admins also bdit_humans?

PR #476, opened 2022-06-28
Volume pipeline

scann0n:

- upserts data in 9 tables on traffic using data updated on TRAFFIC_NEW
- logs changes to 6 of the tables (3 tables were excluded from auditing because they are large and frequently updated).

## Issue(s) this solves:

- #462

- tested on airflow; it worked
- compared table record counts in both schemas and they matched
- reasonable output in logged_actions table
- maybe make sure the dag is efficient??? Would it be better to update all of the tables at the same time?

### DIFF
--- a/volumes/sql/traffic_new/func_upd_arc_link.sql
+++ b/volumes/sql/traffic_new/func_upd_arc_link.sql
@@ -1,0 +1,38 @@
+-- FUNCTION: traffic.update_arc_link()
+
+-- DROP FUNCTION IF EXISTS traffic.update_arc_link();
+
+CREATE OR REPLACE FUNCTION traffic.update_arc_link(
+	)
+    RETURNS void
+    LANGUAGE 'sql'
+    COST 100
+    VOLATILE SECURITY DEFINER PARALLEL UNSAFE
+AS $BODY$
+
+    insert into traffic.arc_link (
+	select * from "TRAFFIC_NEW"."ARC_LINK"
+	except
+	select * from traffic.arc_link)
+	on conflict (link_id) 
+	DO UPDATE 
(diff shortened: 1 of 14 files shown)


volumes/sql/traffic_new/func_upd_arc_link.sql
```diff
@@ -0,0 +1,38 @@
...
+	DO UPDATE 
+		SET arc_id = EXCLUDED.arc_id,
+			frequency = EXCLUDED.frequency;
+
+with delrec as (
+	select * from traffic.arc_link 
+	except 
+	select * from "TRAFFIC_NEW"."ARC_LINK")
+
+Delete from traffic.arc_link where link_id in (SELECT link_id from delrec);
```
tahaislam, line 27: I'm wondering if we're not keeping any of the old data (e.g. `traffic.arc_link`), wouldn't `TRUNCATE` and `INSERT` do the same thing faster than `UPSERT`?

tahaislam (changes requested):
Great work! I just left two comments down there

volumes/sql/traffic_new/func_upd_arc_link.sql
radumas: depends on whether we're auditing that table.

tahaislam: I got it. So, we're auditing all these tables

radumas: Not the really big ones, which does point to me that there should be a smol README edit to highlight which tables are being audited.

tahaislam (approved):
Thanks Sarah

Issue #131, opened 2018-07-13
Migrate Blip API Script to use psycopg2 module


radumas:
Used `pygresql` instead of `psycogp2` to develop `blip_api.py` because of installation issues with `psycopg2` on the Windows terminal server (lack of a postgresql binary IIRC).

But I accidentally used the more... unique version of `pygresql`, which won't be as interchangeable as the [DB-API compliant version](http://www.pygresql.org/contents/pgdb/index.html), which should have a very similar API to `psycopg2`

radumas:
Workaround...

The PygreSQL [installation instructions](http://www.pygresql.org/contents/install.html) read

>On Windows, this library is called libpq.dll and is for instance contained in the PostgreSQL ODBC driver (search for “psqlodbc”).

Navigated to the [downloads page for psqlodbc](https://www.postgresql.org/ftp/odbc/versions/) and selected the MSI option. I downloaded the latest one for PostgreSQL 9.

chmnata:
Switched everything except for an upsert command and a json encode step, which require a little more work since I will have to construct the sql for upsert, and check the result of encoding the json. 

Documentation for pg db's upsert. 
http://www.pygresql.org/contents/pg/db_wrapper.html#pg.DB.upsert

pygresql's json encoder
http://www.pygresql.org/contents/pg/db_wrapper.html#encode-decode-json-encode-and-decode-json-data
>This is the same as the json.dumps() function from the standard library.
I will probably just use json.dumps from json

psycopg2 have some sort of json adaptation
https://www.psycopg.org/docs/extras.html#json-adaptation

chmnata:
tested the script out with with blip backfill and backfilled date 2022-03-22, now backfilling 22 to 28.

PR #479, opened 2022-07-04
smol pr for changing pygresql to use psycopg2 in blip api 

chmnata:

- Remove the usage of pygresql package and use psycopg2 instead

## Issue(s) this solves:

- #131

- If it works
- I feel like there was something about the upsert that I had problem with, so the upsert statement
- Now that I think about it more, I think there was a problem with upsert that I haven't fixed yet


chmnata:
closed, superseded by #480

PR #480, opened 2022-07-04
smol pr for changing pygresql to use psycopg2 in blip api fr this time

chmnata:

- Remove the usage of pygresql package and use psycopg2 instead

## Issue(s) this solves:

- #131

- If it works
- I feel like there was something about the upsert that I had problem with, so the upsert statement
- Now that I think about it more, I think there was a problem with upsert that I haven't fixed yet

### DIFF
--- a/bluetooth/api/blip_api.py
+++ b/bluetooth/api/blip_api.py
@@ -1,9 +1,6 @@
 """blip_api.py
-
 Script to pull Bluetooth data from the Blip api. 
-
 main defaults to getting the previous day's data 
-
 """
 
 import argparse
@@ -14,6 +11,7 @@ import os
 import sys
 import time
 import traceback
+from psycopg2 import connect, sql, InternalError, IntegrityError, DatabaseError
 from datetime import date
 from itertools import product
 from typing import List
@@ -21,7 +19,7 @@ from typing import List
 import urllib3
 import zeep
 from dateutil.relativedelta import relativedelta
-from pg import DB, InternalError, DatabaseError, IntegrityError
+from psycopg2.extras import Json, execute_values
 from requests import RequestException, Session
 from tenacity import retry, before_sleep_log, wait_exponential, retry_if_exception_type, RetryError
 from zeep import Client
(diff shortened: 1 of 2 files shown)


bluetooth/api/blip_api.py
```diff
@@ -165,6 +171,10 @@ def get_wsdl_client(wsdlfile, direct=None, live=False):
             config[key] = zeep.xsd.SkipValue
     return blip, config
 
+class Object:
```
radumas, line 174: I have a suspicion about what this does but it should be documented

bluetooth/api/blip_api.py
```diff
@@ -267,22 +307,23 @@ def main(dbsetting: 'path/to/config.cfg' = None,
...
     blip, config = get_wsdl_client(api_settings['WSDLfile'], direct, live)
 
     try:
-        db = DB(**dbset)
+        conn = _get_db(dbset)
     except InternalError as _:
         LOGGER.error('Connection error to RDS, sleeping for ten minutes')
         sleep(60 * 10)
-        db = DB(**dbset)
+        conn = _get_db(dbset)
```
radumas, line 319: I think this part can be changed to similar to the other _get_db() block

radumas (changes requested):
This should also update the requirements file.

I can't see anything obviously wrong with the upsert query

bluetooth/api/blip_api.py
chmnata: like which ones ? 😵

radumas: dis
https://github.com/CityofToronto/bdit_data-sources/blob/2e9a12801bfa6382f3c8113824a469961361afa2/bluetooth/api/blip_api.py#L122-L127

bluetooth/api/blip_api.py
```diff
@@ -177,70 +187,100 @@ def update_configs(all_analyses, dbset):
...
+                cur.execute("SELECT king_pilot.load_bt_data();")
+                query = cur.fetchone()    
+        if query[0] != 1:
+            conn.rollback()
+            raise DatabaseError('king_pilot.load_bt_data did not complete successfully') 
+        ## Truncate and delete if successful 
+        with conn:
+            with conn.cursor() as cur:
+                cur.execute("TRUNCATE bluetooth.raw_data;")
+                cur.execute("DELETE FROM king_pilot.daily_raw_bt WHERE measured_timestamp < now()::DATE;")
```
chmnata, line 275: is this `now()::DATE` bad? 👁️ 👁️

radumas: zomg yes.
Also rather tempted to say we should delete all the `king_pilot` stuff, but mayyyyybe pulling the live data hourly can be useful again.

radumas (changes requested):
I will concede it works. I will reiterate it would be nice to document the subclassing Object to hack the json representation function

bluetooth/api/blip_api.py
radumas: you have to link to the permalink

image

radumas (changes requested):
requirements.txt needs to be updated.
Does the config file need to be changed?

chmnata:
> requirements.txt needs to be updated. Does the config file need to be changed?

oh yea yes

radumas (approved):
Party time. :meow_party: Excellent :deployparrot:

PR #485, opened 2022-07-13
View of bluetooth routes not reporting yesterday

radumas:

- 
- 

## Issue(s) this solves:

- #

-


radumas:
This is already in the database, will probably get reformed by the ongoing work but for now....

PR #486, opened 2022-07-13
Bt detector map

radumas:

- Makes a map of live Bluetooth routes and detectors
- 

## Issue(s) this solves:

- #

- @chmnata I think you have better views this could be pulling from


radumas:
Closing because @chmnata made a better one

PR #497, opened 2022-08-16
Change TA tables to use declarative partitioning and other enhancements

### LINKED ISSUE #494: HERE: Change partitioned table to use declarative partitioning
- [x] Create parent table
- [x] Create child tables from 2012 to now
- [ ] Import data from old table to new partitioned table
- [ ] Investigate [enable_partition_pruning](https://www.postgresql.org/docs/12/runtime-config-query.html#GUC-ENABLE-PARTITION-PRUNING) in config

### LINKED ISSUE #495: HERE: Add sample size column and change tx to date, and tod
- [ ] Create trigger to apply changes to data before inserting
- [x] Change table structure in new tables

### LINKED ISSUE #496: HERE: Move old ta tables to newly structured ta tables
move old ta table `here.ta` to `here.ta_new` with a loop that 
1) Insert data and formats tx -> 1) dt (date) and 2) tod (time)
2) Truncated old ta table

### LINKED ISSUE #499: HERE: Modify here_api to request sample size

chmnata:

Restructure `here.ta` tables:
- Change from partitioning using inheritance to declarative partitioning 
- Separate column `tx` (timestamp) to `dt` (date) and `tod` (time)
- Change insert triggers to go through simple view

## Issue(s) this solves:

- Closes #494 
- Closes #495 
- Closes #496 
- Closes #499 

- `create_table.sql` :  new parent table structure + partitioned table creation + simple view + trigger on view
- `move_table.sql`: the loop to move old data to new data tables
- `function_create_yearly_table.sql`: function to create yearly partition tables
- `trigger_here_insert.sql`: trigger to insert data to parent table through simple view

Question: 
- Having some problem understanding whether there is a need on putting index on `dt` as it is the partition column. Under 5.10.4.  Partitioning and Constraint Exclusion. I'm wondering if we need to add it when a lot of the times we are filtering for months and not specific dates 🤔 🤔 🤔 Maybe Dr. Doc can comment on this :cat_doctor:
>Note that constraint exclusion is driven only by CHECK constraints, not by the presence of indexes. Therefore it isn't necessary to define indexes on the key columns. Whether an index needs to be created for a given partition depends on whether you expect that queries that scan the partition will generally scan a large part of the partition or just a small part. An index will be helpful in the latter case but not the former.
https://www.postgresql.org/docs/10/ddl-partitioning.html

Notes:
- There is no `ON CONFLICT DO NOTHING` for the insert trigger because we can't put that for parent table in PG version 10 or lower (can do in 11). There are unique constraints on tod, dt, and link_dir for each partitioned table to prevent inserting duplicated data.
- Cannot put indexes on parent table because our PG version 10 or lower, can be changed in 11. So there are still indexes created for each partitioned table.


here/traffic/sql/create_tables.sql
```diff
@@ -1,73 +1,72 @@
...
 ALTER TABLE here.ta
   OWNER TO here_admins;
-*/
 
-/*Loops through the years and then months for which we currently have data in order to create a partitioned table for each month*/
+-- simple view structure
+CREATE VIEW here.ta_view AS
+
+SELECT *
+FROM here.ta;
```
radumas, line 24: This needs to exclude your transformed columns (`dt` and `tod`)

here/traffic/sql/create_tables.sql
```diff
@@ -1,73 +1,72 @@
...
+
+CREATE TRIGGER transform_trigger
+        INSTEAD OF INSERT
+        ON here.ta_view
+        FOR EACH ROW
+        EXECUTE PROCEDURE here.here_insert_trigger();
 
+
+--Loops through the years and then months for which we currently have data in order to create a partitioned table for each month
 DO $do$
```
radumas, line 40: Instead of doing this looping you could loop over that yearly create tables function.

Or use the eoy DAG and run it for 2012 to 2022

here/traffic/sql/create_tables.sql
```diff
@@ -1,73 +1,72 @@
...
-  pct_5 int NOT NULL,
-pct_10 int NOT NULL,
-pct_15 int NOT NULL,
-pct_20 int NOT NULL,
-pct_25 int NOT NULL,
-pct_30 int NOT NULL,
-pct_35 int NOT NULL,
-pct_40 int NOT NULL,
-pct_45 int NOT NULL,
-pct_50 int NOT NULL,
```
radumas, line 23: I think the median speed might be worth keeping.

here/traffic/sql/create_tables.sql
```diff
@@ -1,73 +1,72 @@
...
-pct_55 int NOT NULL,
-pct_60 int NOT NULL,
-pct_65 int NOT NULL,
-pct_70 int NOT NULL,
-pct_75 int NOT NULL,
-pct_80 int NOT NULL,
-pct_85 int NOT NULL,
-pct_90 int NOT NULL,
-pct_95 int NOT NULL
```
radumas, line 32: from your speeding analysis of HERE, are any of the upper percentiles worth keeping?

here/traffic/sql/move_tables.sql
```diff
@@ -0,0 +1,35 @@
+-- Move data from here.ta_* to the newly structured here.ta_*
+-- while performing separation of tx column to dt (Date) and tod (time)
+-- loop through each table, move data to new table, then truncate the old one
+--Loops through the years and then months for which we currently have data in order to create a partitioned table for each month
+
```
radumas, line 5: This needs to be done in Python, because otherwise committing won't happen until the loop is finished and it's going to be a big deal

here/traffic/sql/trigger_here_insert.sql
```diff
@@ -0,0 +1,22 @@
...
+							NEW.tx::date,
+							NEW.tx::time without time zone,
+							NEW.tx,
+							NEW.length,
+							NEW.mean,
+							NEW.stddev,
+							NEW.min_spd,
+							NEW.max_spd,
+							NEW.confidence,
+                            NEW.sample_size); -- postgresql <=10 does not support ON CONFLICT DO NOTHING
```
radumas, line 19: Can you link to this?

here/traffic/sql/create_tables.sql
radumas, line 24: This needs to exclude the transformed columns `dt` and `tod`

radumas (changes requested):
Great work documenting this PR.

Can you write out the transition plan you have for moving over the data?

>Having some problem understanding whether there is a need on putting index on dt as it is the partition column. Under 5.10.4. Partitioning and Constraint Exclusion. I'm wondering if we need to add it when a lot of the times we are filtering for months and not specific dates

But you did create an index on this column for the child tables. I think that's what the documentation is also discussing, whether it's worthwhile to put an index on the children tables. `constraint_exclusion` will quickly determine which tables to query based on the requested dates, but we still need an index to quickly retrieve rows within that table.

Some other comments on the individual elements

here/traffic/sql/trigger_here_insert.sql
chmnata: Under 5.10.2.3. Limitations
> Using the ON CONFLICT clause with partitioned tables will cause an error, because unique or exclusion constraints can only be created on individual partitions. There is no support for enforcing uniqueness (or an exclusion constraint) across an entire partitioning hierarchy.

I'm guessing we will have to create additional triggers for each partitioned table if `ON CONFLICT` is needed.

chmnata:
## Transition plan

1) Create `here.ta_new` with declarative partitioning. Because it is not possible to alter `here.ta` (technically a regular table) to a partitioned table.
> It is not possible to turn a regular table into a partitioned table or vice versa. However, it is possible to add an existing regular or partitioned table as a partition of a partitioned table, or remove a partition from a partitioned table turning it into a standalone table; this can simplify and speed up many maintenance processes. See [ALTER TABLE](https://www.postgresql.org/docs/current/sql-altertable.html) to learn more about the ATTACH PARTITION and DETACH PARTITION sub-commands.
2) Create a new table with all transformation using old child table for every month.
```
CREATE TABLE here.ta_201201_new AS
SELECT columns
FROM here.ta_201201
```
3) Attach newly created table as a partition child table to the newly created declarative parent table. Adding a check because it allow the system to skip the scan when validating the partition constraint. see below text. (Step 2 and 3 will be in a loop in a .py`
```
ALTER TABLE here.ta_201201_new ADD CONSTRAINT y2012m01
   CHECK ( dt>= DATE '2012-01-01' AND dt< DATE '2012-02-01' );
   
ALTER TABLE here.ta_201201_new ATTACH PARTITION here.ta_new
    FOR VALUES FROM ('2012-01-01') TO ('2012-02-01' );
  
DROP CONSTRAINT y2012m01;
```
>Before running the ATTACH PARTITION command, it is recommended to create a CHECK constraint on the table to be attached matching the desired partition constraint. That way, the system will be able to skip the scan to validate the implicit partition constraint. Without the CHECK constraint, the table will be scanned to validate the partition constraint while holding an ACCESS EXCLUSIVE lock on the parent table. It may be desired to drop the redundant CHECK constraint after ATTACH PARTITION is finished.

4) Change all dependencies of `here.ta` to `here.ta_new`. @radumas Most of them are in `data_requests` schema, I think we can just drop those without recreating them since they should all be committed anyways. Some are the reporting mat views for rapidto. Some are in our personal schemas. Will make a list and we can decide on what to do with each of them.

5) Drop `here.ta`, and make sure there are no other hidden dependencies

6) Rename `here.ta_new` to `here.ta`

7) Pray everything works 🙏🏼

here/traffic/sql/create_tables.sql
chmnata: ahhhhhhhhhhhhhhhhhhh yes, seems like we used pct_50, pct_85, and the max. I will include those.

here/traffic/move_monthly_table.py
```diff
@@ -0,0 +1,28 @@
...
+CONFIG = configparser.ConfigParser()
+CONFIG.read(str(Path.home().joinpath('db.cfg')))
+dbset = CONFIG['DBSETTINGS']
+con = connect(**dbset)
+#Configure logging
+FORMAT = '%(asctime)-15s %(message)s'
+logging.basicConfig(level=logging.INFO, format=FORMAT)
+LOGGER = logging.getLogger(__name__)
+
+yyyy_list = ['2012', '2013', '2014', '2015', '2016', '2017', '2018']
```
radumas, line 15: shouldn't this go up to 2017-09?

here/traffic/move_monthly_table.py
```diff
@@ -0,0 +1,28 @@
...
+mm_list = ['01', '02', '03', '04', '05', '06', '07', '08', '09', '10', '11', '12']
+
+for y in yyyy_list:
+    LOGGER.info('Begin processing year %s', y)
+    for m in mm_list:
+        LOGGER.info('Processing month %s of year %s', m, y)
+        with con:
+            with con.cursor() as cur:
+                cur.execute("SELECT here.move_data(%s, %s);", (y,m))
```
radumas, line 24: `here_staging.move_data`

here/traffic/sql/transition.sql
```diff
@@ -0,0 +1,105 @@
...
+    pct_85 integer NOT NULL,
+    confidence integer NOT NULL,
+	sample_size integer
+) PARTITION BY RANGE (dt);
+
+
+-- Create view in staging schema
+CREATE VIEW here_staging.ta_view AS
+ 
+select link_dir, tx, length, mean, stddev, min_spd, max_spd, pct_50, pct_85, confidence, sample_size
```
radumas, line 31: does this need an epoch place-holder?

here/traffic/sql/trigger_here_insert.sql
radumas: ah shoot well I guess at least insert will fail if trying to insert duplicate data, which is fine. It's not like other pipelines where we might get partial data

chmnata:
https://github.com/CityofToronto/bdit_data-sources/blob/0deda9c35cca500bf0b85b72cce7850a6a22a8a9/here/traffic/here_api.py#L65
I feel like we don't specify the mapversion 🤔 and the default is 18_3

chmnata:
_Update on PR as of 2022-08-24_

Restructure `here.ta` tables:
- Change from partitioning using inheritance to declarative partitioning 
- Separate column `tx` (timestamp) to `dt` (date) and `tod` (time)
- Added `sample_size` column
- Change insert triggers to go through simple view

Updates on here_api:
- Direct COPY to `here.ta_view` instead of `here.ta`
- Modified request query parameters to include sample_size and exclude pct_x other than pct_50 and pct_85
- Attempted to catch Key Error (#500)

- `create_table.sql` :  new parent table structure + partitioned table creation + simple view + trigger on view
- `trigger_here_insert.sql`: trigger to insert data to parent table through simple view
- `here_api.py` : Python script to pull here ta data from here api (mainly the query part and error catching portion)

here/traffic/here_api.py
```diff
@@ -47,6 +47,20 @@ def get_access_token(key_id, key_secret, token_url):
...
+        LOGGER.error('Error in requesting access token')
+        LOGGER.error(err)
+        try:
+            err_msg = query_response.json()['message']
+        except JSONDecodeError:
+            err_msg = query_response.text
+        finally:
+            raise HereAPIException(err_msg)
+
     access_token = r.json()['accessToken']
```
radumas, line 64: this line should be in the try block

here/traffic/here_api.py
```diff
@@ -47,6 +47,20 @@ def get_access_token(key_id, key_secret, token_url):
     LOGGER.info('Getting Access Token')
     r = requests.post(token_url, auth=oauth1, json=payload, headers=headers)
 
+    try:
+        r.raise_for_status()
+    except KeyError as k_err:
+        LOGGER.error(k_err)
```
radumas, line 53: This should be 
```python
LOGGER.error('Key Error in getting access token, response was')
raise HereAPIException(r.text)
```

here/traffic/here_api.py
```diff
@@ -108,6 +122,7 @@ def get_download_url(request_id, status_base_url, access_token, user_id):
         LOGGER.info('Polling status of query request: %s', request_id)
         query_status = requests.get(status_url, headers = status_header)
         try:
+            query_status.raise_for_status()
```
radumas, line 125: Should probably include the same `except requests.exceptions.HTTPError as err:` from above

here/traffic/here_api.py
```diff
@@ -165,7 +182,7 @@ def send_data_to_database(ctx=None, datafile = None, dbsetting=None):
         unzip = subprocess.Popen(['gunzip','-c',datafile], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         #Second uses check_call and 'ON_ERROR_STOP=1' to make sure errors are captured and that the third 
         #process doesn't run befor psql is finished.
-        copy = r'''"\COPY here.ta FROM STDIN WITH (FORMAT csv, HEADER 
+        copy = r'''"\COPY here_staging.ta_view FROM STDIN WITH (FORMAT csv, HEADER
```
radumas, line 185: should this be `here.ta_view` after transition is complete?

here/traffic/sql/create_tables.sql
```diff
@@ -1,73 +1,53 @@
...
+GRANT INSERT, UPDATE, SELECT, TRIGGER ON TABLE here_staging.ta TO here_bot;
+
+-- Simple view structure 
+CREATE VIEW here.ta_view AS
+ 
+SELECT  link_dir, 
+        tx, 
+        NULL:integer as epoch_min, 
+        ength, 
+        sample_size,
```
radumas, line 30: flagging that sample size is not in the same location as table. Confirm this is correct location

radumas (changes requested):
Great work. a few edits in the error handling in `here_api.py`.

SQL looks fine just confirm that sample_size is in the correct location in the view vs. the table.

here/traffic/sql/create_tables.sql
chmnata: yup! confirming that this is the correct column order for the input csv.gz to copy to. Then the trigger reorganize the columns to fit the here.ta table column order.

here/traffic/here_api.py
chmnata: should it be catching different errors than line 50 to 62? Just Jsondecodeerror? and maybe another keyerror?

radumas (changes requested):
sql is good. Some problems in the python file.

here/traffic/here_api.py
```diff
@@ -51,7 +51,13 @@ def get_access_token(key_id, key_secret, token_url):
...
     except KeyError as k_err:
         LOGGER.error('Key Error in getting access token, response was') 
-        raise HereAPIException(r.text)
+        LOGGER.error(k_err)
+        try: 
+            err_msg = r.json()['message'] 
+        except JSONDecodeError:
+            err_msg = r.text
+        finally:
+            LOGGER.error(err_msg)
```
radumas, line 60: stil need `raise HereAPIException(r.text)` otherwise the script won't actually fail properly here.

here/traffic/here_api.py
```diff
@@ -130,7 +136,7 @@ def get_download_url(request_id, status_base_url, access_token, user_id):
         LOGGER.info('Polling status of query request: %s', request_id)
         query_status = requests.get(status_url, headers = status_header)
         try:
-            query_status.raise_for_status()
```
radumas: still wanted this line in

here/traffic/here_api.py
```diff
@@ -137,6 +138,7 @@ def get_download_url(request_id, status_base_url, access_token, user_id):
         query_status = requests.get(status_url, headers = status_header)
         try:
             status = str(query_status.json()['status'])
+            query_status.raise_for_status()
```
radumas, line 141: these lines are in the wrong order.

radumas (changes requested):
Simplifying the error catching code for the two functions parsing responses from the API. Hopefully this... works.

for the getting access token:

```python
    try:
        r.raise_for_status()
        access_token = r.json()['accessToken']
    except (requests.exceptions.HTTPError, KeyError, JSONDecodeError, ValueError)  as err:
        error = 'Error in requesting access token \n'
        error += 'Response was:\n'
        try:
            resp = r.json()['message']
        except JSONDecodeError:
            resp = r.text
        finally:
            error += resp
            error += '\n'
            raise HereAPIException(error)
```

For the get download URL

```python
        try:
            query_status.raise_for_status()
            status = str(query_status.json()['status'])
        except (requests.exceptions.HTTPError, KeyError) as err:
            error = 'Error in polling status of query request \n'
            error += 'err\n'
            error += 'Response was:\n'
            try:
                resp = str(query_status.json()['message'])
            except JSONDecodeError:
                resp = query_status.text
            finally:
                error += resp
                error += '\n'
                raise HereAPIException(error)
        except JSONDecodeError as json_err:
            LOGGER.warning("JSON error in query status response.")
            LOGGER.warning(query_status.text)
            continue
```

radumas:
@chmnata CodeFactor raised an smol ❌

radumas (approved):
My name is @radumas and I approve this Pull Request :here: :octocat: :meow_praise:

Issue #501, opened 2022-08-26
Getting json.decoder.JSONDecodeError from `wys_api.py`


tahaislam:
The DAG run of Aug 22, 2022 stopped due to an unhandled exception: `json.decoder.JSONDecodeError`. Here is the full stack:

```
[2022-08-25 12:12:05,095] {taskinstance.py:1150} ERROR - Expecting value: line 1 column 1 (char 0)
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.6/site-packages/airflow/models/taskinstance.py", line 984, in _run_raw_task
    result = task_copy.execute(context=context)
  File "/home/airflow/.local/lib/python3.6/site-packages/airflow/operators/python_operator.py", line 113, in execute
    return_value = self.execute_callable()
  File "/home/airflow/.local/lib/python3.6/site-packages/airflow/operators/python_operator.py", line 118, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/etc/airflow/data_scripts/wys/api/python/wys_api.py", line 336, in api_main
    table, loc_table = get_data_for_date(start_date, signs_list, api_key)
  File "/etc/airflow/data_scripts/wys/api/python/wys_api.py", line 259, in get_data_for_date
    statistics=get_statistics_hourly(api_id, start_date, hr, api_key)
  File "/etc/airflow/data_scripts/wys/api/python/wys_api.py", line 120, in get_statistics_hourly
    error=response.json()
  File "/home/airflow/.local/lib/python3.6/site-packages/requests/models.py", line 890, in json
    self.content.decode(encoding), **kwargs
  File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

PR #502, opened 2022-08-26
Handling invalid json responses in wys_api.py #501

tahaislam:

- It handles the exception `json.decoder.JSONDecodeError` in `wys_api.py` by logging the returned error message and passing the exception back to the calling function in order to skip the failing hour 

## Issue(s) this solves:

- #501 

- function `get_statistics_hourly` in `wys/api/python/wys_api.py`

### DIFF
--- a/wys/api/python/wys_api.py
+++ b/wys/api/python/wys_api.py
@@ -18,7 +18,7 @@ import psycopg2
 from psycopg2 import connect
 from psycopg2.extras import execute_values
 from requests import Session, exceptions
-
+import json
 
 class WYS_APIException(Exception):
     """Base class for exceptions."""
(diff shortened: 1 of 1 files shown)


wys/api/python/wys_api.py
```diff
@@ -110,23 +110,23 @@ def get_statistics_hourly(location, start_date, hr, api_key):
                         ':00/to/'+str(hr).zfill(2)+':59/speed_units/0', 
                         headers=headers)
     # Error handling
+    try:
+        res = response.json()
```
radumas, line 114: I think this should start with a `response.raise_for_status()` and follow some of the structure from the `here_api.py`

https://github.com/CityofToronto/bdit_data-sources/pull/497/files

radumas (changes requested):
I think handling json responses could be improved to more closely match the here_api structure

tahaislam:
This branch was merged into #536 and then into `master`

Issue #487, opened 2022-07-22
Make the GCC puller work in Airflow


radumas:
It will run on the On-Prem Dev server and send data to both the on-prem database and our RDS.

The state of the code is currently in https://github.com/CityofToronto/bdit_data-sources/tree/gcc_view/gis/gccview

Layers that were already being pulled should continue to be pulled and probably automated to be pulled with an audit table. 

Only the layers to be used for geoprocessing VFH data should be sent to the On-Prem database.

As part of this we will want to version the centreline and intersections datasets with a parent table of allllll the centreline, child table of versions of centreline (where we add a `pulled_date` for each version), and then create a view that will also show the most current version.

The query to the centreline API should filter out things we know are not streets in the `fcode_desc` column:
`'Creek/Tributary', 'Ferry Route','Geostatistical line', 'Hydro Line', 'Major Railway', 'Major Shoreline', 'Minor Shoreline (Land locked)', 'River'`
It would probably be good to alert when there are new fcode_desc that we do not currently have in the database. 

I'm not sure what to propose for a similar filter for intersections.

qu-xiang:
Current layers being pulled:

'0_0': ward,
'0_2': centreline,
'2_2': bike,
'2_3': traffic_camera,
'2_9': traffic_signal,
'2_11': permit_parking,
'2_35': prai_transit_shelter, 
'2_37': tmms_service_request,
'2_38': bylaw_pt,
'2_39': bylaw_line,
'20_1': ev_charging_station,
'22_1': day_care,
'22_2': middle_child,
'23_1': bia,
'23_13': proposed_bia,
'23_9': film_permit,
'23_10': film_parking,
'23_12': hotel,
'26_1': convenience_store,
'26_4': supermarket,
'26_3': worship,
'26_6': ymca,
'26_7': census_tract,
'26_11': neighbourhood_impro,
'26_13': priority_neigh,
'26_16': neigh_demo,
'26_45': aborginal,
'26_46': attraction,
'26_47': dropin, 
'26_48': early_year,
'26_49': family_resource,
'26_50': food_bank,
'26_53': long_term_care,
'26_54': parenting_family_lit,
'26_58': retirement,
'26_59': senior_housing, 
'26_61': shelter,
'26_62': social_housing,
'27_13': private_road,
'28_17': school,
'28_28': library

radumas:
[...]

PR #506, opened 2022-10-13
Updated GCC puller and DAG

qu-xiang:

- Updated GCC puller scripts
- Updated the DAG that periodically pulls layers using Airflow
- Have layers that will be used in VFH issue 32 [done]
- Updated README

## Issue(s) this solves:

- #487 

- gcc_puller_functions.py
- gcc_layers_pull.py
- Other sql definition scripts under `gcc/gccview/sql`
- Proper permissions of tables and schemas in both bigdata / ptc servers
- README under `gis/gccview`
- Whether it works properly in Airflow


gis/gccview/gcc_layers_pull.py
```diff
@@ -0,0 +1,130 @@
...
+            exec_date=context.get('execution_date'),
+            log_url=context.get('task_instance').log_url,
+        )
+    failed_alert = SlackWebhookOperator(
+        task_id='slack_test',
+        http_conn_id='slack',
+        webhook_token=slack_webhook_token,
+        message=slack_msg,
+        username='airflow',
+        proxy='http://137.15.73.132:8080')
```
tahaislam, line 43: Can you update the proxy to be `'http://'+BaseHook.get_connection('slack').password+'@137.15.73.132:8080'` to contain username and password?

gis/gccview/create-centreline.sql
```diff
@@ -0,0 +1,59 @@
...
+    high_num_even integer,
+    geom geometry
+) PARTITION BY LIST (version_date)
+WITH (
+    OIDS = FALSE
+)
+TABLESPACE pg_default;
+
+ALTER TABLE IF EXISTS gis_core.centreline
+    OWNER to bqu;
```
tahaislam, line 53: Please, change all ownership to `gis_admins`

gis/gccview/gcc_layer_refresh.py
```diff
@@ -0,0 +1,662 @@
...
+            keep_adding = find_limit(return_json)
+            if keep_adding == True:
+                max_number = max_number + record_max
+            else:
+                LOGGER.info('All records from [mapserver: %s, layerID: %d] have been inserted into %s', mapserver, layer_id, output_table)
+    
+    if is_audited:
+        successful_task_run = update_table(output_table, insert_column, excluded_column, primary_key, schema_name, con)
+
+    return successful_task_run
```
tahaislam, line 655: Sorry, but it turns out that just returning True/False isn't enough to mark a task as `FAILED`. The correct way to do it is to raise `AirflowFailException`. So instead of `return successful_task_run`, it could be something like
```
from airflow.exceptions import AirflowFailException
if not successful_task_run:
    raise AirflowFailException
```

gis/gccview/gcc_layers_pull.py
```diff
@@ -0,0 +1,138 @@
...
+# Credentials - to be passed through PythonOperator
+from airflow.hooks.postgres_hook import PostgresHook
+# bigdata connection credentials
+bigdata_cred = PostgresHook("gcc_bot_bigdata")
+# On-prem server connection credentials
+morbius_cred = PostgresHook("gcc_bot")
+
+SLACK_CONN_ID = 'slack_data_pipeline'
+# Slack IDs of data pipeline admins
+dag_config = Variable.get('slack_member_id', deserialize_json=True)
```
tahaislam, line 23: need to add `from airflow.models import Variable`

gis/gccview/gcc_layers_pull.py
```diff
@@ -0,0 +1,138 @@
...
+from airflow.operators.python_operator import PythonOperator
+from airflow.hooks.base_hook import BaseHook
+from airflow.contrib.operators.slack_webhook_operator import SlackWebhookOperator
+
+# Credentials - to be passed through PythonOperator
+from airflow.hooks.postgres_hook import PostgresHook
+# bigdata connection credentials
+bigdata_cred = PostgresHook("gcc_bot_bigdata")
+# On-prem server connection credentials
+morbius_cred = PostgresHook("gcc_bot")
```
tahaislam, line 19: better rename it as `ptc_cred`

gis/gccview/gcc_layer_refresh.py
```diff
@@ -0,0 +1,664 @@
...
+    return successful_execution
+#-------------------------------------------------------------------------------------------------------
+'''
+@click.command()
+@click.option('--mapserver_n', '-ms', help = 'Mapserver number, e.g. cotgeospatial_2 will be 2', type = int)
+@click.option('--layer_id', '-ly', help = 'Layer id', type = int)
+@click.option('--schema_name', '-s', help = 'Name of destination schema', type = str)
+@click.option('--is_audited', '-a', help = 'Whether the table is supposed to be audited (T) or partitioned (F)', type = bool)
+'''
+#-------------------------------------------------------------------------------------------------------
```
radumas, line 583: Not that I'm really sure we still need command-line interfaces (but they can useful!).

The here pipeline has a decent example of having the click decorated function being a wrapper around a "main" Python function. (note there's some complexity in that program due to trying to allow for running the sub-functions of the pipeline that's not relevant here. AKA anything involving the `ctx` object)

qu-xiang:
Currently the airflow scheduler cron expression is `schedule_interval='0 0 0 3,6,9,12 *'` , I think it is correct but it doesn't run as intended. 😞

tahaislam:
> Currently the airflow scheduler cron expression is `schedule_interval='0 0 0 3,6,9,12 *'` , I think it is correct but it doesn't run as intended. 😞

I think the day of the month shouldn't be 0. However, this isn't the reason why it seems to be not working

tahaislam (approved):

gis/gccview/README.md
```diff
@@ -0,0 +1,92 @@
+# GCCVIEW pipeline
+
+## Overview
+
+The GCC (Geospatial Competency Centre) offers a number of spatial products on their arcgis rest api. They host 26 map servers and has over [300 layers](https://github.com/CityofToronto/bdit_data-sources/blob/gcc_view/gis/gccview/available_layers.md) from different divisions within the City of Toronto. To ensure the geospatial layers we use are the most up to date at a regular basis, we are creating a data pipeline to automate the process of retrieving layers from gccview to our postgresql database and conflate street networks through sharedstreets.
+
+Since we can't access the GCC rest API outside the City's firewall, the pipeline has to run in the on-premise server (Morbius / Bancroft) environment, using Airflow 2.1.4.
```
radumas, line 7: since there are layers that can't be access from the GCC's external REST API.

gis/gccview/gcc_layer_refresh.py
```diff
@@ -0,0 +1,664 @@
...
+                insert_audited_data(output_table, insert_column, return_json, schema_name, con)
+            else:
+                insert_partitioned_data(output_table_with_date, insert_column, return_json, schema_name, con)
+            
+            counter += 1
+            keep_adding = find_limit(return_json)
+            if keep_adding == True:
+                max_number = max_number + record_max
+            else:
+                LOGGER.info('All records from [mapserver: %s, layerID: %d] have been inserted into %s', mapserver, layer_id, output_table)
```
radumas: I don't feel like this code needs to get repeated after the outer `else`.
The only thing in the outer `else` could be  
```python
else:
    return_json = get_data(mapserver, layer_id, max_number = max_number, record_max = record_max)
#and then 
if is_audited:
    #etc
```

gis/gccview/gcc_layer_refresh.py
```diff
@@ -0,0 +1,664 @@
...
+    
+    if is_audited:
+        successful_task_run = update_table(output_table, insert_column, excluded_column, primary_key, schema_name, con)
```
radumas: Took me a moment but realized the update function should always delete the temp table if there's an error, so this should be idempotent on error. Nice work.

radumas (changes requested):
Really impressive work, especially overcoming the challenges of parametrising the SQL calls.

I have a few comments in the files.

Also want to highlight that I think `gis/gccview/new_gcc_puller_notebook.ipynb` should replace the old notebook instead of being called `new`

gis/gccview/sql/ptc_audit_gis.sql
```diff
@@ -0,0 +1,210 @@
+CREATE TABLE gis.logged_actions (
```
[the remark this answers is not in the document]
tahaislam: I think re-using that script from the other repo makes more sense to make it easier for any future updates. The only difference between the two scripts is the `ptc_humans` vs. `bdit_humans` groups on PTC servers vs. `bigdata`

gis/gccview/gcc_layers_pull.py
```diff
@@ -0,0 +1,139 @@
+# The official new GCC puller DAG file
```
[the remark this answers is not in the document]
tahaislam: It's actually under `dags`. I think this file was moved to the correct folder in later commits
image

gis/gccview/README.md
tahaislam: modified

gis/gccview/README.md
```diff
@@ -0,0 +1,92 @@
...
+- schema_name (string): name of destination schema
+- is_audited (Boolean): True if the layer will be in a table that is audited, False if the layer will be inserted as a child table part of a parent table
+- cred (Airflow PostgresHook): the credentials to enable a connection to a particular database
+
+In the DAG file, the arguments for each layer is stored in a dictionary called "bigdata_layers" or "morbius_layers", in the order of the parameters above. The DAG will be executed once every XXX days. 
+
+
+## Manually fetch layers
+
+Currently you can use [this notebook](https://github.com/CityofToronto/bdit_data-sources/blob/gcc_view/gis/gccview/get_layer_gccview.ipynb) on Morbius server environment to fetch layer from gccview rest api and send it to postgresql in the schema you want.
```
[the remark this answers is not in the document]
tahaislam: updated

gis/gccview/gcc_layer_refresh.py
```diff
@@ -0,0 +1,664 @@
...
+    except requests.exceptions.RequestException as err:
+        LOGGER.error("Error: ", err)
+    else:
+        ajson = r.json()
+        layers = ajson['layers']
+        for layer in layers:
+            if layer['id'] == layer_id:
+                output_name = (layer['name'].lower()).replace(' ', '_')
+            else:
+                continue
```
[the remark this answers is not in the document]
tahaislam: updated

gis/gccview/gcc_layer_refresh.py
```diff
@@ -0,0 +1,664 @@
...
+
+    return output_name
```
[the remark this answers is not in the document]
tahaislam: moved `return output_name` inside the `else` block.
Do you think we need to try more than once before exiting?

gis/gccview/gcc_layer_refresh.py
```diff
@@ -0,0 +1,664 @@
...
+             "returnZ": "false",
+             "returnM": "false",
+             "orderByFields": "OBJECTID", 
+             "returnDistinctValues": "false",
+             "returnExtentsOnly": "false",
+             "resultOffset": "{}".format(max_number),
+             "resultRecordCount": "{}".format(record_max),
+             "f":"json"}
+    
+    while True:
```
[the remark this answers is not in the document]
tahaislam: oh no. We have to limit the number of retries

gis/gccview/gcc_layer_refresh.py
tahaislam: I think these decorators are a bit outdated. In the final implementation, only `manual_get_layer`, which is called in `__main__`, should be decorated with `click`

radumas (approved):
:meow_praise: :meow_purrfecto:

Issue #434, opened 2021-11-10
Update DAGs to use Airflow connections instead of user config files


radumas:
e.g. https://github.com/CityofToronto/bdit_data-sources/blob/master/vision_zero/schools.py#L143

PR #513, opened 2022-11-28
School safety zones

### LINKED ISSUE #511: Updating file paths in the school safety zones DAG
Updating the path to `schools.py` script which is imported within vz_google_sheets DAG.
Old path: `/etc/airflow/data_scripts/vision_zero/`
New path: `/etc/airflow/data_scripts/gis/school_safety_zones`

### LINKED ISSUE #512: Flagging invalid SSZ records
Flagging invalid school safety zone records by marking the corresponding task as `failed`

tahaislam:

- Updated the path to `schools.py` script that is imported from `vz_google_sheets` DAG
  - now the path is independent of the repo path 
- Marking any year with invalid records as `failed`
  - while continuing inserting all other valid records

## Issue(s) this solves:

- resolves #511 
- resolves #512

### DIFF
--- a/dags/vz_google_sheets.py
+++ b/dags/vz_google_sheets.py
@@ -58,8 +58,16 @@ sheets = {
 def task_fail_slack_alert(context):
     slack_webhook_token = BaseHook.get_connection(SLACK_CONN_ID).password
     # print this task_msg and tag these users
-    task_msg = """The Task vz_google_sheets (ssz):{task} failed. {slack_name} please fix it """.format(
-        task=context.get('task_instance').task_id, slack_name = list_names,) 
+    invalid_rows = context.get('task_instance').xcom_pull(task_ids=context.get('task_instance').task_id, 
+                                                            key='invalid_rows')
+    if invalid_rows:
+        task_msg = """The Task vz_google_sheets (ssz):{task} failed.
+                    Found one or more invalid records. {slack_name} please email David Tang """.format(
+            task=context.get('task_instance').task_id, slack_name = list_names,) 
+    else:
+        task_msg = """The Task vz_google_sheets (ssz):{task} failed. {slack_name} please fix it """.format(
+            task=context.get('task_instance').task_id, slack_name = list_names,) 
+    
     # this adds the error log url at the end of the msg
     slack_msg = task_msg + """ (<{log_url}|log>)""".format(
             log_url=context.get('task_instance').log_url,)
@@ -100,6 +108,7 @@ DEFAULT_ARGS = {
     'start_date': datetime(2019, 9, 30),
     'retries': 0,
     'retry_delay': timedelta(minutes=5),
+    'provide_context':True,
     'on_failure_callback': task_fail_slack_alert
 }
 
--- a/dags/vz_google_sheets.py
+++ b/dags/vz_google_sheets.py
@@ -102,7 +102,7 @@ con = vz_api_bot.get_conn()
 DEFAULT_ARGS = {
     'owner': 'cnangini',
     'depends_on_past' : False,
-    'email': ['cathy.nangini@toronto.ca'],
+    'email': ['islam.taha@toronto.ca'],
     'email_on_failure': True,
     'email_on_retry': True,
     'start_date': datetime(2019, 9, 30),
--- a/dags/vz_google_sheets.py
+++ b/dags/vz_google_sheets.py
@@ -4,7 +4,7 @@ Pipeline for pulling two vz google sheets data and putting them into postgres ta
 from datetime import datetime, timedelta
 from airflow import DAG
 from airflow.operators.python_operator import PythonOperator
-from airflow.hooks.postgres_hook import PostgresHook
+from airflow.providers.postgres.hooks.postgres import PostgresHook
 from airflow.models import Variable 
 from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
 from googleapiclient.discovery import build
--- a/dags/vz_google_sheets.py
+++ b/dags/vz_google_sheets.py
@@ -100,7 +100,7 @@ vz_api_bot = PostgresHook("vz_api_bot")
 con = vz_api_bot.get_conn()
 
 DEFAULT_ARGS = {
-    'owner': 'cnangini',
+    'owner': 'itaha',
     'depends_on_past' : False,
     'email': ['islam.taha@toronto.ca'],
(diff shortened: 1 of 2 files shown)


radumas:
My two comments based on today's experience with the following log lines as example:

```
ERROR - WYSS : outside Toronto
ERROR - An error occurs at [row]
```

1. Can the first line include the coordinates or whatever that failed? "cannot be transformed into a date" is an example that includes it (although whatever failed should also be quoted).
2. the second line should read "on row: [row]" or something similar to link the error message with the more complete context of where the erroneous data point is. I initially thought the two lines weren't related.

gis/school_safety_zones/schools.py
```diff
@@ -231,6 +236,8 @@ def pull_from_sheet(con, service, year, spreadsheet, *args):
             cur.execute(truncate)  
             execute_values(cur, query, rows)
     LOGGER.info('Table %s is done', table)
+    if any_error:
+        raise AirflowFailException
```
radumas, line 240: Is it possible to pass text to this? Such that the log line would be something like

```
[2022-11-28 13:56:12,548] {taskinstance.py:1150} ERROR - invalid rows pulled. See errors documented above.
```

radumas (changes requested):

radumas:
Reporting it back seems simpler at the moment, what do you think?
Wish there was an easy validation that could work in Sheets formulas

chmnata (changes requested):
Looks good and well documented :meow_party:!! One general question that is not related to the tagged issues: We are still using a `key.json` from joven's home directory (which I thought we killed?), and a db config from cathy's home, should we change this? They are in line 258 and 264 respectively in `schools.py`. Cathy's email is also listed for alert emeows.

tahaislam:
> Looks good and well documented :meow_party:!! One general question that is not related to the tagged issues: We are still using a `key.json` from joven's home directory (which I thought we killed?), and a db config from cathy's home, should we change this? They are in line 258 and 264 respectively in `schools.py`. Cathy's email is also listed for alert emeows.

good catch. we can make it more generic and read these variables as arguments to the script, if we still need to keep the file's `__main__` environment. These hard-coded local paths are only used when `schools.py` run on its own outside Airflow. Hence, it doesn't have access to Airflow's variables and connections (right?)

gis/school_safety_zones/schools.py
```diff
@@ -204,10 +208,12 @@ def pull_from_sheet(con, service, year, spreadsheet, *args):
                     LOGGER.info('Reading %s columns of data from Google Sheet', len(row))
```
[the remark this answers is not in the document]
tahaislam: agree, printing the same statement doesn't add any value, especially that all rows almost have the same number of columns. I guess the message was intended initially to print out the number of columns in each processed row to identify any outliers. But this's not needed now, I guess

chmnata:
> > Looks good and well documented :meow_party:!! One general question that is not related to the tagged issues: We are still using a `key.json` from joven's home directory (which I thought we killed?), and a db config from cathy's home, should we change this? They are in line 258 and 264 respectively in `schools.py`. Cathy's email is also listed for alert emeows.
> 
> good catch. we can make it more generic and read these variables as arguments to the script, if we still need to keep the file's `__main__` environment. These hard-coded local paths are only used when `schools.py` run on its own outside Airflow. Hence, it doesn't have access to Airflow's variables and connections (right?)

hmmm yea I'm not sure if we need it then, since we can run the script from /airflow and those paths doesn't exist

chmnata (changes requested):

radumas:
CodeFactor makes a good point this is a dangerous "bare" `except:` 

https://github.com/CityofToronto/bdit_data-sources/blob/c5cf17ea11b8a6c513ce3a1ad487ee2d556d2eb7/gis/school_safety_zones/schools.py#L108-L122

gis/school_safety_zones/schools.py
```diff
@@ -153,7 +154,7 @@ def is_int(n):
 
 #--------------------------------------------------------------------------------------------------
 
-def pull_from_sheet(con, service, year, spreadsheet, *args):
+def pull_from_sheet(con, service, year, spreadsheet, **kwargs):
```
radumas, line 157: Is this function actually receiving keyword arguments?

tahaislam: yes, it's used to send some variables over XCom to the slack alert function

radumas (commented):
This is awesome! We should have a lightning talk about this use of xcom.

Do you think we should set up the code with task_instance and AirflowFailException such that it could run without being in the Airflow environment or is that irrelevant going forward for this pipeline?

tahaislam:
> This is awesome! We should have a lightning talk about this use of xcom.
> 
> Do you think we should set up the code with task_instance and AirflowFailException such that it could run without being in the Airflow environment or is that irrelevant going forward for this pipeline?

Maybe. This would require having a small intermediate function in the DAG calling `pull_from_sheet` and sending whatever needed over xcom and returning the appropriate exception when needed. Or is there another way to do it?

gis/school_safety_zones/schools.py
radumas: @tahaislam do you think there should be something logged before the `AirflowFailException` would appear?

tahaislam: This's already resolved in this [commit](https://github.com/CityofToronto/bdit_data-sources/commit/0fb49c656f1cc1cba17930e138b897fd7d6319d7), but unfortunately, it's not showing here in this PR
https://github.com/CityofToronto/bdit_data-sources/blob/4eca6882859696f2794a5e5e0ab39ef029ee2618/gis/school_safety_zones/schools.py#L246

There're more commits in this branch that are not showing in this PR. And I don't know why this's happenning

radumas (approved):
I see the changes!

radumas:
@chmnata can you also approve :)

dags/vz_google_sheets.py
```diff
@@ -90,12 +102,13 @@ def task_fail_slack_alert(context):
 DEFAULT_ARGS = {
     'owner': 'cnangini',
```
chmnata, line 103: let's change this to itaha

chmnata (approved):

chmnata:
Possible closing of this #434

dags/vz_google_sheets.py
tahaislam: Done

PR #488, opened 2022-07-22
New HERE documentation!

scann0n:
## What this pull request accomplishes:

- Organizes, updates and cross-references documentation for our HERE data
- 

## Issue(s) this solves:

- #481

## What, in particular, needs to reviewed:

- Is it accurate and 
- do you like the format and 
- is the level of detail right

PR #527, opened 2023-02-15
Here docs update

Nate-Wessel:
- updates Here docs to reflect current schema

- `here/README.md`

### DIFF
--- a/here/README.md
+++ b/here/README.md
@@ -44,9 +44,9 @@ Historical data is acquired through the Traffic Analytics download portal. Data 
 | min_spd | integer | | Observed minimum speed |
 | max_spd | integer | | Observed maximum speed |
 | pct_50 | integer | | Observed median speed |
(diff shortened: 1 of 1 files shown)


chmnata:
looks like this is branched off from sarah's doc_here branch? We also have this #488 HERE doc PR outstanding, wondering if you also modified the doc according to Raph's comment on that PR.

Nate-Wessel:
I hadn't looked at those - just updated the table to reflect the current schema, and my best understanding of the various fields. 

This is branched off of master. 

I'd be happy to take a look at those other branches too and see if we can incorporate their changes as well, after this one. We'll probably have some conflicting edits to deal with at some point...

chmnata (approved):
much clearer, thanks for adding the description for samplecounts and confidence!

Nate-Wessel:
@chmnata thanks for the quick review!

Issue #529, opened 2023-02-17
`log cleanup` DAG: Deprecated core/base_log_folder in Airflow 2.2.5


tahaislam:
After upgrading Airflow to 2.2.5, `base_log_folder` moved to `logging` section in the configuration file. Hence, we're getting the following error:

```
Broken DAG: [/etc/airflow/dags/log_cleanup.py] Traceback (most recent call last):
  File "/home/airflow/airflow_venv/lib/python3.6/site-packages/airflow/configuration.py", line 373, in get
    return self._get_option_from_default_config(section, key, **kwargs)
  File "/home/airflow/airflow_venv/lib/python3.6/site-packages/airflow/configuration.py", line 383, in _get_option_from_default_config
    raise AirflowConfigException(f"section/key [{section}/{key}] not found in config")
airflow.exceptions.AirflowConfigException: section/key [core/base_log_folder] not found in config
```

tahaislam:
> Oh! We got this as a warning before and I just moved the base_log_folder under a new section `logging`, following the config file in morbius

I think they keep changing this from one version to another. It won't break anything also if we added the same field and value to the core

PR #532, opened 2023-02-22
#529 update core to logging

chmnata:

- Change path from `core` to `logging` since the new config file moved base_log_folder from core to logging in the new airflow version
- unbreak broken dag

## Issue(s) this solves:

- #529

- tested on dev_script, it works 👍🏼

### DIFF
--- a/dags/log_cleanup.py
+++ b/dags/log_cleanup.py
@@ -84,7 +84,7 @@ START_DATE = datetime(2020, 2, 25)
 SCHEDULE_INTERVAL = '@daily'
 DAG = create_dag(__file__, __doc__, START_DATE, SCHEDULE_INTERVAL)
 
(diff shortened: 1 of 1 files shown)


dags/log_cleanup.py
```diff
@@ -84,7 +84,7 @@ def create_dag(filepath, doc, start_date, schedule_interval):
 SCHEDULE_INTERVAL = '@daily'
 DAG = create_dag(__file__, __doc__, START_DATE, SCHEDULE_INTERVAL)
 
-BASE_LOG_FOLDER = conf.get("core", "BASE_LOG_FOLDER")
+BASE_LOG_FOLDER = conf.get("logging", "BASE_LOG_FOLDER")
```
chmnata, line 87: testing

tahaislam (approved):

PR #541, opened 2023-03-01
Collision docs - update `collisions` schema to `collisions_replicator`

### LINKED ISSUE #538: `/collisions/Readme.md` needs updating
At the least, it refers extensively to a `collisions` schema that no longer exists. We now have:
* `collision_factors`
* `collisions_DO_NOT_USE`
* `collisions_replicator`

though given this difference, there are likely other changes that should be documented as well.

Nate-Wessel:

- updates schema name from `collisions` to `collisions_replicator`

## Issue(s) this solves:

- closes #538 

- Ensure all claims are still valid. Was this a simple renaming? If other things were changed, some information may have changed.


Nate-Wessel:
@scann0n, Raphael said you would be a good person to review this?

scann0n (approved):
Re `ACC` vs `acc_safe_copy`: the latter table is the one that is truncated every night. @radumas mentioned that we should maybe prohibit people from querying `acc_safe_copy` directly - the mat views are there for a reason.
There is a `uid` in `acc_safe_copy` - it's called `RECID` I think.
This is an improvement over what we had. I think we should merge this and then I can update it as per Sprint last week.

PR #545, opened 2023-03-10
Remove `parking/`

### LINKED ISSUE #35: Add readme to /parking/
@aharpalaniTO to add readme for [parking](https://github.com/CityofToronto/bdit_data-sources/) subfolder

Nate-Wessel:
- removes the `parking/` directory, which is not actually a BDIT data source but rather just local a copy of data from Open Data

## Issue(s) this solves:
- closes #35 

- that this is the right thing to do


radumas:
Many of our data sources are direct copies from other City sources, whether they be OpenData or the GCC. Given that we have use parking ticket data for previous analyses and that we are about to support the Parking Strategy, I think these datasets may yet be useful and it would be premature to delete this folder.

Nate-Wessel:
I'll go ahead and just add a basic README then.

radumas:
:meow_thx:

PR #515, opened 2022-12-12
Set primary key on classifications table

Nate-Wessel:
## What this pull request accomplishes:

- creates a primary key on the table

## Issue(s) this solves:

- lets other tables reference this key

## What, in particular, needs to reviewed:

- check syntax

PR #564, opened 2023-03-31
Add primary keys to miovision `classifications` and `intersections` tables

Nate-Wessel:

- adds a primary key to the `miovision_api.intersections` table
- points out that `miovision_api.classifications` is defined here as using `classification_uid` as a primary key but does in fact not do that. 

## Issue(s) this solves:

- NA

- definitions here match reality in the database

### DIFF
--- a/volumes/miovision/sql/create-table-intersections.sql
+++ b/volumes/miovision/sql/create-table-intersections.sql
@@ -1,6 +1,6 @@
 ﻿CREATE TABLE miovision_api.intersections
 (
-    intersection_uid integer,
+    intersection_uid integer PRIMARY KEY,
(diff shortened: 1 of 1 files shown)


chmnata (approved):
Looks good thanks! Made the change in the database as well 👍🏼

chmnata:
Wait sorry, is there an sql for `miovision_api.classifications` as well?

Nate-Wessel:
> Wait sorry, is there an sql for `miovision_api.classifications` as well?

It looks like that was changed in https://github.com/CityofToronto/bdit_data-sources/pull/515 but I apparently didn't actually follow up to confirm that the change had been made to the table.

Issue #575, opened 2023-04-11
WYS | Update the start_date in locations table


tahaislam:
Update `wys.locations` so that the earliest `start_date` for each `api_id` is the first date on which we have data.

This was originally part of issue #444

PR #577, opened 2023-04-12
Update the start_date in wys.locations

tahaislam:

- Update the table `wys.locations` so that the earliest `start_date` for each `api_id` is the first date on which we have data
- refresh the two materialized views feeding into Open Data

## Issue(s) this solves:

- #575 

- A sql script to run one time to apply the above changes


gabrielwol (approved):

radumas:
@tahaislam to merge and apply

Issue #595, opened 2023-05-03
Remove references to individual users in Slack DAG notification setup


gabrielwol:
>Not for this PR, but in general do we need to keep all of our names in the Slack notification? If so, should we use something like `' '.join(dag_config.values())` to avoid changing all the DAGs whenever we want to add a new name, like adding Gabe at some point

PR #597, opened 2023-05-05
#595 Remove hardcoded references to owners, emails, slack_ids in DAGs…

gabrielwol:
… and refer to Airflow var instead.

-Removes hardcoded references to users (Raph, Islam, Natalie) in DAGs. 
-Refer to new `dag_owner` var in Airflow to identify DAG Owner and also identify their slack ids for Slack failure notifications. 
-Removes emails from dag default parameters. 

## Issue(s) this solves:
#595 

-


radumas (approved):
Natalie says this worked for two other dags so

Issue #573, opened 2023-04-06
Create a table of dubious dates for RESCU VDS Stations 


radumas:
Similar to #570

scann0n:
Here's what I've done so far:
1. Looked at continuous bin gap length >> `scannon.rescu_gaps_21`
2. Calculated gap frequency based on `scannon.rescu_gaps_21` and:
    - ~80% of the gaps were 1h 15 min or less (note - this looks like 1h 30 min in the data because bin start times are used (so the gap would start 15 minutes later than we think it does)
3. ID detectors with 91 bins per day instead of 96 >> `scannon.rescu_data_days_91`
4. Within those detectors, how many meet Raph’s transportation rules of thumb?
5. Made `scannon.rescu_dayvol_stats_21`
6. Used 20000 and 10000 as the “enough daily volume” thresholds; detectors and days with enough volume are in `scannon.rescu_enuf_vol`
7. Made a table that has all the detectors with 100 or more days of good data called `scannon.rescu_good_vol_days`

The branch is called i573_bad_rescu_dates; the sqls are in bdit_data-sources/volumes/rescu/date_evaluation!

PR #605, opened 2023-05-11
I573 bad rescu dates

scann0n:

- This is the 2021 RESCU data quality work we did for the TEPS request.
- It can be used as a prototype for other similar pieces of work, but it is not quite a "package" that can be automated due to the "look at the graphs and find a logical minimum threshold for each corridor" aspect.
- This does not include ramps.

## Issue(s) this solves:

- #573 

- rescu_data_days_91.sql
- rescu_dayvol_stats_21.sql
- rescu_enuf_vol.sql
- rescu_gaps_21.sql
- rescu_good_vol_days.sql
- I just took a look at these and they aren't the final sqls. Some of the thresholds are different but the methodology is there. Happy to update the code with the thresholds we landed on.
- The RESCU readme.md was updated to reflect this work as well (see the Data Quality section).

- Nothing yet; this is still very TEPS specific.


volumes/rescu/date_evaluation/x1_rescu_good_vol_days.sql
```diff
@@ -0,0 +1,21 @@
...
+        COUNT(ev.dt) AS ev_day_ct 
+    FROM scannon.rescu_enuf_vol AS ev
+    LEFT JOIN rescu.detector_inventory AS di USING (detector_id)
+    GROUP BY 
+        ev.detector_id,
+        di.det_group,
+        di.number_of_lanes,
+        di.primary_road || ' and ' || di.cross_road,
+        ST_SETSRID(ST_MakePoint(di.longitude, di.latitude), 4326)
+    HAVING COUNT(ev.dt) >= 100
```
gabrielwol, line 19: I think `HAVING COUNT(ev.dt) >= 100` should be removed since the description is "see how many days". Or change desc to "see how many detectors meet/exceed 100 days of data availability".

gabrielwol (approved):
Given we are in the middle of overhauling the RESCU -> VDS schema (which will include additional disaggregate data to use for quality checks), @radumas thinks we should get this work over the finish line as a current state for RESCU data validation, which can be built on in the next phase (#658). The next phase could consider a more automated approach using a rolling horizon and individual sensor thresholds.

PR #589, opened 2023-04-24
Miovision anomalous ranges table

scann0n:
NOTE: comment edited substantially from the original, as this PR has evolved quite a bit

## What this pull request accomplishes:

- Creates a new table `miovision_api.anomalous_ranges` that serves to identify and document (potential) issues with the Miovision data
- Documents the above along with the process of using / updating / populating / managing it

## Issue this solves:

- #570 

## What, in particular, needs to reviewed:

- that the documentation makes clear enough what this table is and how to use it to filter data

## After this is merged, please:
- [x] Migrate the anomaly table and referenced tables to `miovision_api`
- [x] Create audit table structure for anomalies?
- [x] Create a `miovision_data_detectives` group with `INSERT`, `DELETE`, `UPDATE` permissions on the anomaly table. Add `nwessel` to group.

PR #611, opened 2023-05-19
Correct docs regarding miovision streetcar classification

Nate-Wessel:

- reclassifies streetcars as "buses"
- adds link to Miovision's documentation of this
- applies more standard formatting to that markdown table for readability

- That streetcars are in fact classified as buses

### DIFF
--- a/volumes/miovision/README.md
+++ b/volumes/miovision/README.md
@@ -63,21 +63,22 @@ classification|text|Textual description of mode|Bicycles|
 location_only|boolean|If TRUE, represents movement on crosswalk (as opposed to road)|FALSE|
 class_type|text|General class category (Vehicles, Pedestrians, or Cyclists)|Cyclists|
 
-Here is a description of the `classification_uid`s and corresponding types. For more information, see [Miovision's documentation on this topic](https://help.miovision.com/s/article/North-American-Classification-Guide). Note that bicycles are available at both a turning movement level and at an approach level. Approach level bicycle counts should be used for the large majority of applications as the data is considered more accurate.
+Here is a description of the `classification_uid`s and corresponding types. For more information, see [Miovision's documentation on this topic](https://help.miovision.com/s/article/North-American-Classification-Guide). **Note:** Some classifications have changed over time, see notes related to this below the table and see progress on determining when these changes happened in [Issue 634](https://github.com/CityofToronto/bdit_data-sources/issues/634).
 
 | classification_uid | classification | definition |
 | ------------------ | -------------- | ---------- |
 1 | Light | Cars and other passenger vehicles (like vans, SUVs or pick-up trucks)|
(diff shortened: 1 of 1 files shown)


radumas:
That document doesn't say anything about streetcars being classified as buses

radumas:
From a 2019 email

>The reason you might see two 'Bus' Classes is this: 'Motorized Vehicles' happens to be what we label Streetcars. We have found in Toronto that it seems like the only thing our Motorized Vehicle class is catching are Streetcars. Per an earlier conversation we had, I think it was your preference that we label streetcars as buses, so we have introduced this mapping,
Motorized Vehicle -> Bus
>Bus is also a separate class type.
>So we generate two items in the API if in the same bin, at least one bus and one street car make the same movement.

Nate-Wessel:
It's near the bottom under "additional information"

> Trams and streetcars are grouped with the Bus class. If your study requires a separate Trams volume, select Trams as a [custom classification](https://help.miovision.com/s/article/Custom-Classifications-in-Traffic-Data-Online?language=en_US).

Nate-Wessel:
> From a 2019 email
> 
> > The reason you might see two 'Bus' Classes is this: 'Motorized Vehicles' happens to be what we label Streetcars. We have found in Toronto that it seems like the only thing our Motorized Vehicle class is catching are Streetcars. Per an earlier conversation we had, I think it was your preference that we label streetcars as buses, so we have introduced this mapping,
> > Motorized Vehicle -> Bus
> > Bus is also a separate class type.
> > So we generate two items in the API if in the same bin, at least one bus and one street car make the same movement.

Is this saying that streetcars would have been be double-counted? Once as a bus, and again as a MotorizedVehicle?

My _suspicion_ at this point is that this email may have been correct in 2019, but something changed sometime later on and now streetcars are simply classified as buses. It's the only way that current volumes at e.g. King and Spadina make sense.

Nate-Wessel:
Copying a message from Jesse posted in Slack:
> Just a quick followup, miovision did confirm that streetcars did in fact change classification with an algorithm change at some point, they were going to investigate further to figure out when exactly etc  (they used to be classified as motorized vehicles, and now seem to be classified as buses) ...

Nate-Wessel:
This seems like a decent case for being handled in part through the Miovision QC tables, proposed in https://github.com/CityofToronto/bdit_data-sources/pull/589. This would be a note applying to this `classification_uid` for all intersections, separately both for the pre- and post- periods. I think there may also be some variation in implementation time between intersections, though my best guess at the moment is that this rolled out around 2019-08-01 for most. 

Should of course also be noted in the documentation here.

Nate-Wessel:
I added a note to the docs to really spell this out. IMO, we can open a separate issue to properly pin down the exact date of the change.

radumas (approved):

PR #613, opened 2023-05-24
Retry running pull_wys DAG in case of failure

### LINKED ISSUE #612: Retry running the pull_wys DAG in case of failure
We noticed the `pull_wys` DAG failing on some days due to an unknow error while pulling the schedules `{'result': {'error': {'error-exception': 'An error occurred during execution; please try again later.'}}}`. The `pull_wys` task usually runs without errors when we clear it later.

Suggested solution: set the `retries` task/DAG parameter to a non-zero value (say 3) and may also activate the `retry_exponential_backoff` option. 

https://airflow.apache.org/docs/apache-airflow/2.5.2/_api/airflow/models/baseoperator/index.html#airflow.models.baseoperator.BaseOperator

tahaislam:

- Set the number of retries in `pull_wys` DAG to 3 

## Issue(s) this solves:

- Resolves #612

- The DAG file at `dags/pull_wys.py`

- Deploy into production

### DIFF
--- a/dags/pull_wys.py
+++ b/dags/pull_wys.py
@@ -73,7 +73,7 @@ default_args = {'owner':'rdumas',
                  'email_on_success': False,
                  'retries': 3,
                  'retry_delay': timedelta(minutes=5),
-                 'retry_exponential_backoff' = True,
+                 'retry_exponential_backoff': True,
                  'on_failure_callback': task_fail_slack_alert
                 }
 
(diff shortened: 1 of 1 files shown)


dags/pull_wys.py
```diff
@@ -71,8 +71,9 @@ def task_fail_slack_alert(context):
                 'email': ['raphael.dumas@toronto.ca'],
                 'email_on_failure': False,
                  'email_on_success': False,
-                 'retries': 0,
+                 'retries': 3,
                  'retry_delay': timedelta(minutes=5),
+                 'retry_exponential_backoff': True,
```
gabrielwol, line 76: Add comment: #progressive longer waits between retries

gabrielwol (approved):
Add comment and good (y)

PR #618, opened 2023-05-26
Make `text_to_centreline` more case-insensitive

Nate-Wessel:

- removes case-sensitivity from words like "of" and "point"

## Issue(s) this solves:

- This briefly tripped me up on a recent data request where the input came in all-caps, and was transformed with `InitCap`. I wouldn't expect "A Point 100 Metres South..." to be handled differently from "a point 100 metres south..."

- minor change to one function

replace `gis.text_to_centreline` with the updated function


Nate-Wessel:
I'm happy to fluff this, though this small change will be easier to review if I keep the changes simple.

radumas (approved):
Cimpl Change Chanks.

Nate-Wessel:
@radumas or @chmnata can this get implemented and merged?

radumas:
Implement and merged

PR #647, opened 2023-06-30
removing volumes open data notebooks

### LINKED ISSUE #551: Clean Up the `volumes/open_data`  folder
We're not the ones currently publishing these to Open data.
- [ ] Figure out if the jupyter notebooks are at all useful

radumas:

- removes Open Data Jupyter Notebooks

From @ShahrzadBorjian2 

>While the notebooks are served as a valuable reference and followed a commendable procedure, I think they should be deleted from open data because of the following reasons:
Outdated Data: The dataset includes information that is no longer up to date.
Updated Gap Analysis and Quality Check:  @Nate-Wessel &  @scann0n  have conducted a thorough gap analysis and quality check on Miovison, resulting in updated information. Their efforts have provided more reliable and accurate data
Notebooks as Drafts for Public Sharing: The notebooks containing the dataset appear to be in the draft stage, intended for sharing with the public.
Data from Different Years: The outlier check is done on data from various years. This temporal disparity could introduce discrepancies when comparing values or identifying outliers over a ten-year span.

## Issue(s) this solves:

- Closes #551 

- whether we agree with Shahrzad's final act before becoming a mother (I for one am inclined to say yes)
    - Nate I think maybe poke at one notebook and compare with Shahrzad's notes and then we can merge.

Nothing


Nate-Wessel (approved):
I did a spot check on several of these files and I agree; I don't see a lot of value in keeping these. They reference tables that no longer exist and it's not super clear what they're documenting or trying to accomplish.

radumas:
Thanks Nate!

PR #653, opened 2023-07-12
Lengthen the `traffic.det` TMC table

### LINKED ISSUE #652: We should have a view that converts turning movement counts to a format like `miovision_api.volumes_15min_tmc`
_This is actually @Nate-Wessel's idea 💡_

The wide format can be quite clunky.

The sql for this should go in [`volumes/short_term_counting_program/sql`](https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/short_term_counting_program/sql) and a _brief_ explanation can be added to the [Useful Views](https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/short_term_counting_program#5-useful-views).

For future reference the process produces 43,087,968 rows and takes 8 minutes to run

Nate-Wessel:

- Adds a view that lengthens the wide-format `traffic.det` for increased compatibility with `miovision_api.volumes_15min_tmc`

## Issue(s) this solves:

closes #652 

- code, permissions, ownership

- run the script creating the matview
- set up a process for keeping it refreshed


radumas (commented):
Do you think `count_info_id` would be useful to include? Is there any other useful count metadata that someone would need to join to the info table for?

Someone from Data Ops can add this refreshing this to the traffic refresh DAG next week. 

I think it probably warrants an index on the timestamp field too. 

We also need an addition to the README on how people can use this

Nate-Wessel:
> Do you think `count_info_id` would be useful to include? Is there any other useful count metadata that someone would need to join to the info table for?

Yeah, `count_info_id` is probably the key one. Anything else you'd be able to get from that. I'll add that and also an index for it. 

We should be passing through all the information from `det` now with that change.

scann0n (dismissed):
I have created the mat view in `traffic` with the small change noted after line 122.

Also we have a decision to make regarding "DAG-ification"...

I wrote a function to update the mat view `concurrently with data` which the DAG will run. The function needs a unique index. We can:
- add a serial unique id and make an index off of that
- make a unique id based on multiple fields, which means we'd need to resolve the nulls in the     
    - `datetime_bin` (by eliminating entries with null `count_date` values (this occurs for 6 unique `count_info_id` values and there's no data loss associated with this)
    - `classification_uid` (by assigning a value for `other` modes; either 9 (for Miovision's "motorized vehicle" or something like 99 to show there's no alignment with the Miovision classifications)
    - if we resolve these nulls we shoooouuuulllld be able to make a unique index off of datetime_bin, classification_uid, arterycode, count_info_id and leg but it's still a bit uncertain.
    - I left `movement_uid` out because the short term counts don't indicate movement for `pedestrians`, `bikes` and `other`.
- I'm happy to chat more!

volumes/short_term_counting_program/sql/tmc_miovision_long_format.sql
```diff
@@ -0,0 +1,129 @@
...
+
+CREATE INDEX ON traffic.tmc_miovision_long_format (arterycode);
+CREATE INDEX ON traffic.tmc_miovision_long_format (classification_uid);
+CREATE INDEX ON traffic.tmc_miovision_long_format (datetime_bin);
+CREATE INDEX ON traffic.tmc_miovision_long_format (count_info_id);
+
+COMMENT ON MATERIALIZED VIEW traffic.tmc_miovision_long_format
+IS 'converts traffic.det TMC table into a long format more compatible with the way we store Miovision TMC data';
+
+COMMENT ON COLUMN traffic.tmc_miovision_long_format.count_date
```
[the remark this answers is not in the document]
Nate-Wessel: Thanks! Hard to fully check these scripts without actually being able to run them.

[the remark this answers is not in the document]
scann0n: 100%

Nate-Wessel:
There are duplicate entries in `traffic.det`:
```sql
SELECT *
FROM traffic.det
WHERE count_info_id = 82
ORDER BY count_time::time
```

Do we not have any checks/constraints on this table?

scann0n:
As far as I'm concerned, this is good to go.
I have updated the `traffic_transfer.py` dag to include updates to this mat view, and I tested it using a symbolic link.
 @radumas we still need a sys admin to get the dag running from `data_scripts`; then this will be completely complete!

scann0n:
The updated traffic_transfer.py can be found here: https://github.com/CityofToronto/bdit_data-sources/blob/i652_dagified/dags/traffic_transfer.py

radumas:
> The updated traffic_transfer.py can be found here: https://github.com/CityofToronto/bdit_data-sources/blob/i652_dagified/dags/traffic_transfer.py

you should create a PR to incorporate that update into this branch.

dags/traffic_transfer.py
```diff
@@ -24,20 +36,30 @@
...
+            *Dag*: {dag}
+            *Execution Time*: {exec_date}
+            *Log Url*: {log_url}
+            """.format(
+            slack_name=' '.join(list_names),
             task=context.get('task_instance').task_id,
-            slack_name = ' '.join(list_names))
-        
-    slack_msg = task_msg + """(<{log_url}|log>)""".format(
-            log_url=context.get('task_instance').log_url,)
```
radumas, line 34: Can you revert the Slack Alert message format?

scann0n: @radumas done!

dags/traffic_transfer.py
```diff
@@ -3,19 +3,31 @@
...
+import os
+import sys
+from threading import local
+import psycopg2
+from psycopg2 import sql
+import requests
+from psycopg2.extras import execute_values
+from airflow.operators.python_operator import PythonOperator
+from airflow.models import Variable
+from dateutil.parser import parse
```
radumas, line 15: @scann0n All of these seem redundant or not-used.

radumas (changes requested):
Sorry just one more change to the DAG....

dags/traffic_transfer.py
scann0n: They have been deleted!

radumas (approved):

radumas:
@scann0n you can create the mat view?

PR #661, opened 2023-08-02
Ecocounter data

Nate-Wessel:

- defines a table structure for the ecocounter data
- provides a python script to assist in importing data downloaded from the web
- will document the data source and how to use it (high level at least)

## Issue(s) this solves:

We have a great source of bike volume data that's only going to keep expanding and getting better but hadn't yet really acknowledged it's existence as a data source or documented it's use. 

everything

nothing


Nate-Wessel:
@chmnata This is still a bit of a work in progress. It's obviously not yet up to the standard set by some of our other datasets, but I'd appreciate some feedback all the same. I'm happy to add more documentation where needed! 

Thanks!

volumes/ecocounter/tables/sites.sql
```diff
@@ -0,0 +1,20 @@
...
+    geom geometry(POINT, 4326),
+    facility_description text,
+    replaced_by numeric REFERENCES ecocounter.sites (site_id)
+);
+
+ALTER TABLE ecocounter.sites OWNER TO ecocounter_admins;
+
+GRANT SELECT ON ecocounter.sites TO bdit_humans;
+
+COMMENT ON COLUMN ecocounter.sites.site_id
```
chmnata, line 13: typo site in the comments

volumes/ecocounter/tables/sites.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE ecocounter.sites (
```
chmnata, line 1: Can we add comments for each table description what it stores? Is site_id the unique location id? Still have a slight confusion of what replaced_by contains

volumes/ecocounter/tables/flows.sql
```diff
@@ -0,0 +1,15 @@
...
+    site_id numeric NOT NULL REFERENCES ecocounter.sites (site_id),
+    flow_direction text NOT NULL,
+    flow_geom geometry(LINESTRING,4326) CHECK(flow_geom IS NULL OR ST_NPoints(flow_geom) = 2),
+    bin_size interval NOT NULL
+);
+
+GRANT SELECT ON ecocounter.flows TO bdit_humans;
+
+COMMENT ON COLUMN ecocounter.flows.bin_size
+IS 'temporal bins are either 15 or 30 minutes, depending on the sensor';
```
chmnata, line 12: is it now all in 15 min bin? Or does the 30 min bin still exist in the old data

volumes/ecocounter/tables/counts.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE ecocounter.counts (
+    flow_id numeric REFERENCES ecocounter.locations (flow_id),
+    datetime_bin timestamp NOT NULL,
+    volume smallint,
+    UNIQUE(flow_id, datetime_bin)
+);
+
+CREATE INDEX ON ecocounter.counts (datetime_bin);
```
chmnata, line 8: probably an index on flow_id as well

volumes/ecocounter/tables/counts.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE ecocounter.counts (
```
chmnata, line 1: We would ideally partition this in years but it is good for meow

volumes/ecocounter/tables/counts.sql
```diff
@@ -0,0 +1,13 @@
...
+    flow_id numeric REFERENCES ecocounter.locations (flow_id),
```
chmnata, line 2: Do you find yourself joining to the flows table to get the site_id a lot? Wondering if it would be helpful to include site_id here to reduce the need of joining tablels

volumes/ecocounter/tables/sites.sql
chmnata: is it an old location replaced by a new location ?

volumes/ecocounter/pull-data-from-api.py
```diff
@@ -0,0 +1,123 @@
...
+def getChannelData(channel_id,firstData='2015-01-01T00:00:00'):
+    requestChunkSize = timedelta(days=100)
+    requestStartDate = datetime.strptime(firstData,"%Y-%m-%dT%H:%M:%S%z")
+    data = []
+    while requestStartDate.timestamp() < datetime.now().timestamp():
+        response = requests.get(
+            f'{endpoint}/api/data/site/{channel_id}',
+            headers={'Authorization': f'Bearer {token}'},
+            params={
+                'begin': requestStartDate.isoformat(timespec='seconds'),
```
chmnata, line 48: ideally the `begin` and `end` would be a param that we can call but this is great for meow

chmnata (commented):
@Nate-Wessel  I think this is great for meow, just some comments on adding more comments in the table but your readme is great. I don't think you need to fix anything right meow, I can make those changes later if this works well for you at the moment

volumes/ecocounter/pull-data-from-api.py
Nate-Wessel: Yeah, I mostly just wanted to demo for you how the basic API calls work and get something that would be functional for pulling all the data.

volumes/ecocounter/tables/counts.sql
Nate-Wessel: Only a couple million records at the moment though. Is there a threshold where partitioning starts to make sense?

volumes/ecocounter/tables/counts.sql
Nate-Wessel: Not sure it makes much difference (in terms of speed) as the sites and flows tables will always be very small. I think I'd actually kind of rather have the extra join just to make clear how the concepts relate (i.e. that each site has 1+ flows)

volumes/ecocounter/tables/counts.sql
Nate-Wessel: Good catch. I kinda assumed this was happening implicitly already like for a primary key, but I guess not:

> Since a DELETE of a row from the referenced table or an UPDATE of a referenced column will require a scan of the referencing table for rows matching the old value, it is often a good idea to index the referencing columns too. Because this is not always needed, and there are many choices available on how to index, declaration of a foreign key constraint does not automatically create an index on the referencing columns.

from [here](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK)

volumes/ecocounter/tables/sites.sql
Nate-Wessel: I've changed the name of the column to try to be more clear. The `replaced_by_site_id` is the `site_id` of the new replacement site, if there is one. Some new / replacement sensors got installed in a few sites and instead of having the same `site_id`, they have a new one even though they're in the same physical location.

Nate-Wessel: I added a comment trying to explain what a site is. If it still seems a bit vague, that may be because it is a bit.

volumes/ecocounter/tables/flows.sql
Nate-Wessel: I'm actually not seeing the half hour bins now, but the hourly bins are still there for sure:

```sql
SELECT *
FROM ecocounter.counts
WHERE flow_id = 101052525
ORDER BY datetime_bin ASC
LIMIT 1000
```

Nate-Wessel:
@chmnata It's been a couple month since this branch was last updated. This data is now being updated semiregularly (by me, manually) and relied upon by other projects. Do you think you could approve this so I can merge it in and at least get the docs into the main branch?

Is there anything more that needs to be done before that can happen? I know that ideally this would be a whole automated pipeline first, but is that being prioritized? As it is, the data is there, in use, and not documented in the main branch.

chmnata (approved):
Great start of the doc, merging!

volumes/ecocounter/tables/sites.sql
```diff
@@ -0,0 +1,23 @@
...
+    site_id numeric PRIMARY KEY,
+    site_description text NOT NULL,
+    geom geometry(POINT, 4326),
```
chmnata, line 4: huhhh im not sure what this fluff error is

Nate-Wessel: I think `geometry` is a keyword in standard SQL, not a datatype like in PostGIS

PR #662, opened 2023-08-02
Add context to stale cycling unit dataset

Nate-Wessel:

- adds a tiny bit of badly needed context to this readme stub
- renames the directory to clarify that this is not a directory for all data related to the mode of cycling but rather refers to the Cycling and Pedestrian Projects Unit as the originator of the dataset
- moves the seasonality model to the short-term counts folder (the context where the model would likely be applied)

## Issue(s) this solves:

adds to the documentation of this largely undocumented dataset

facts

NA


radumas (approved):
Thanks.

I edited to ensure the folders are referenced in their parent folder README. Please do so in the future.

PR #667, opened 2023-08-04
Update anomalous field name in docs

Nate-Wessel:

- updates an old field name that sneaked through review of #589 and made it into the docs


radumas (approved):
Good catch!

PR #674, opened 2023-08-15
I0669 aggregate volumes 15min mvt tries to insert duplicate values

### LINKED ISSUE #669: aggregate-volumes_15min_mvt tries to insert duplicate values
This longstanding issue with aggregate-volumes_15min_mvt trying to insert duplicate values will need to be fixed in order to simplify the process of adding new miovision intersections. 

> So there is indeed a bug in the aggregation function. Because of the 0-padding of null-volumes it will try to insert a 0 where there might already be 0 data.
> 
> https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/miovision/sql/function-aggregate-volumes_15min_mvt.sql#L47-L49
> 
> That is because the aggregation query starts with the cross product of intersection X time-bins (from `generate_series`) for the date(s) to aggregate. It then left joins on existing volumes excluding `A.volume_15min_mvt_uid IS NULL`, but there are two conditions in which that field may be `NULL`:
> 1. the row exists and the column is null
> 2. the row doesn't exist
> 
> There's a subsequent `HAVING` clause that should exclude 0-volumes from being inserted... except for the classifications we deem to be important.
> 
> I can think of two solutions:
> 1. add an anti-join to this insert query to exclude intersection X time-bins that have already been populated (and I guess stop relying on this `volume_15min_mvt_uid` foreign key relationship
> 2. Insert 0s into `volumes` where they've been zero-padded.
> 
> Deleting volumes for specific intersections to re-run the CLI won't work because this function still zero-pad the other intersections on that date

_Originally posted by @radumas in https://github.com/CityofToronto/bdit_data-sources/issues/429#issuecomment-1230706354_

gabrielwol:
- Adds new function `miovision_api.aggregate_15_min_mvt_single_intersection` to aggregate just a single intersection into `volumes_15min_mvt`.
- Improvements to existing `miovision_api.aggregate_15_min_mvt` function

## Issue(s) this solves:
- Closes #669 

- Fluffed and unnested the CTE in base function https://github.com/CityofToronto/bdit_data-sources/blob/09130027d9b2ab637dd4876663d804467757909d/volumes/miovision/sql/function-aggregate-volumes_15min_mvt.sql

- Duplicated new base function but added an intersection function argument and added an anti join to new function to prevent adding duplicate 0 padded values. [function-aggregate-volumes_15min_mvt_single_intersection.sql](https://github.com/CityofToronto/bdit_data-sources/blob/09130027d9b2ab637dd4876663d804467757909d/volumes/miovision/sql/function-aggregate-volumes_15min_mvt_single_intersection.sql#L38-L41)
https://github.com/CityofToronto/bdit_data-sources/blob/09130027d9b2ab637dd4876663d804467757909d/volumes/miovision/sql/function-aggregate-volumes_15min_mvt_single_intersection.sql#L38-L41
`WHERE...`
https://github.com/CityofToronto/bdit_data-sources/blob/09130027d9b2ab637dd4876663d804467757909d/volumes/miovision/sql/function-aggregate-volumes_15min_mvt_single_intersection.sql#L72-L73

- Updated https://github.com/CityofToronto/bdit_data-sources/compare/master...i0669-aggregate-volumes_15min_mvt-tries-to-insert-duplicate-values to use the new single intersection function if intersections are specified in function call. 

- Update definition of `miovision_api.aggregate_15_min_mvt` 
- Add new function `miovision_api.aggregate_15_min_mvt_single_intersection`


volumes/miovision/api/intersection_tmc.py
```diff
@@ -321,9 +321,19 @@ def process_data(conn, start_time, end_iteration_time):
...
-                update="SELECT miovision_api.aggregate_15_min_mvt(%s::date, %s::date)"
-                cur.execute(update, time_period)
-                logger.info('Aggregated to 15 minute movement bins')
+                #if intersections specified, aggregate this step one at a time
+                # with different single intersection function
+                if user_def_intersection:
+                    for c_intersec in intersections:
+                        update="""SELECT miovision_api.aggregate_15_min_mvt_single_intersection(
+                                    %s::date, %s::date, %s::int)"""
+                        cur.execute(update, time_period. c_intersec.uid)
```
radumas, line 330: there's a period here instead of a comma. I also think you might need to expand the `time_period` tuple and pass one large tuple of parameters.

volumes/miovision/sql/function-aggregate-volumes_15min_mvt_single_intersection.sql
```diff
@@ -0,0 +1,115 @@
...
+
+RAISE NOTICE '% Done aggregating to 15min MVT bin', timeofday();
+END;
+
+$BODY$;
+
+ALTER FUNCTION miovision_api.aggregate_15_min_mvt_single_intersection(date, date, int)
+OWNER TO miovision_admins;
+
+GRANT EXECUTE ON FUNCTION miovision_api.aggregate_15_min_mvt_single_intersection(date, date, int) TO public;
```
radumas, line 111: I think we could probably, actually, revoke this

volumes/miovision/sql/function-aggregate-volumes_15min_mvt.sql
```diff
@@ -1,76 +1,104 @@
...
 END;
 
 $BODY$;
 
 ALTER FUNCTION miovision_api.aggregate_15_min_mvt(date, date)
-    OWNER TO miovision_admins;
+OWNER TO miovision_admins;
 
-GRANT EXECUTE ON FUNCTION miovision_api.aggregate_15_min_mvt(date, date) TO PUBLIC;
+GRANT EXECUTE ON FUNCTION miovision_api.aggregate_15_min_mvt(date, date) TO public;
```
radumas, line 101: I think we should actually revoke this.

radumas (changes requested):
Great work! All the SQL looks good, just a minor issue in `cur.execute`

volumes/miovision/sql/function-aggregate-volumes_15min_mvt.sql
gabrielwol: Done

gabrielwol:
Ready for re-review!

radumas (approved):

chmnata:
Changes applied on postgres + airflow

PR #680, opened 2023-08-23
I0638 vds detector inventory

### LINKED ISSUE #638: New process to update table rescu.detector_inventory from ITSC
This table is out of date and we do not have a process for updating. 
Replace with a detector_inventory pulled from ITS Central.

gabrielwol:
This PR will make it easier to use the `vdsconfig` and `entity_locations` fact tables which previously had very complex and slow join conditions to the dimension tables. 

- Add foreign keys between dimension tables `vdsconfig` and `entity_locations` and fact tables: `raw_vdsdata`, `raw_vdsvehicledata`, `counts_15min`, `counts_15min_bylane`, `veh_speeds_15min`, `veh_length_15min`. 
- Don't summarize data without foreign keys (we don't know details of those detectors).  
- Add view `detector_expected_bins` to dedup complex case statement in `counts_15min` and `counts_15min_bylane` queries. 
- Adjust `entity_locations` table to have a start and end timestamp instead of just start in order to make join conditions easier. 
- Add old `rescu.detector_inventory` table to new schema as `detector_inventory_cursed` in case some of the manual columns are needed.
- Updates to readme for changes in this PR. 
- Updated partition functions to be more generic for use on other tables in vds schema: replaced `create-function-partition_vdsdata.sql`, `create-function-partition_vdsvehicledata.sql` with `create-function-partition_vds_yyyy.sql`, `create-function-partition_vds_yyyymm.sql` and updated usage in check_partitions task. 
 
## Issue(s) this solves:

- Closes #638

- Use of trigger functions to add foreign keys. Tested this method using sample data in gwolofs schema: 
  - volumes/vds/sql/trigger/create-trigger-add_vds_fkeys.sql
- Many of the changes are duplicate; replacing vds_id / detector_id with `vdsconfig_uid` in multiple tables. Adding `entity_location_uid` to multiple tables. 

- @gabrielwol will need to make all the changes to the table definitions after the PR is approved. Will then need to update foreign keys columns for approximately 2B rows of existing data (already underway in gwolofs schema).


gabrielwol:
I tested the foreign key update trigger some more and it becomes very slow once there were several million rows in the test table. I'm going to go back to the drawing board for this one.

PR #686, opened 2023-08-30
Make all DAGs time-zone aware

### LINKED ISSUE #685: Make all DAGs Time Zone Aware
We should make our DAGs time zone aware by explicitly specifying the time zone of each DAG's `start_date`. Here's an example of the proposed change:

`start_date = datetime(2020, 4, 1)`
should be
`start_date = pendulum.datetime(2020, 4, 1, tz="America/Toronto")`

tahaislam:

- This PR makes our DAGs time-zone aware by explicitly specifying the time zone of each DAG's start_date

## Issue(s) this solves:

- resolves #685 

- In each DAG file (in `dags` folder), I imported `pendulum` and converted the DAG's `start_time` from `datetime` to `pendulum.datetime` with "America/Toronto" time zone.

- After merging the branch, we need to update `data_scripts` by running `git pull;`

### DIFF
--- a/dags/assets_pull.py
+++ b/dags/assets_pull.py
@@ -31,6 +31,7 @@ from airflow.models import Variable
 
 from dateutil.parser import parse
 from datetime import datetime
+import pendulum
 
 dag_name = 'traffic_signals_dag'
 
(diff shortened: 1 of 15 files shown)


radumas (approved):
Looks good to me. Slightly curious about the difference between "America/Toronto" and "Canada/Eastern" but I wasn't able to quickly find it

tahaislam:
> Looks good to me. Slightly curious about the difference between "America/Toronto" and "Canada/Eastern" but I wasn't able to quickly find it

According to [this](https://en.m.wikipedia.org/wiki/List_of_tz_database_time_zones), I think they are the same thing

Issue #687, opened 2023-08-30
WYS experienced a large drop in data volume beginning 2023-04-01. 


gabrielwol:
While working on #659 I noticed only a single mobile sign had data available in `wys.raw_data` for June 2023. 
Upon further investigation, there appears to have been a large drop in data volume beginning 2023-04-01 affecting both stationary and mobile signs:  

![graph_visualiser-1693430915043](https://github.com/CityofToronto/bdit_data-sources/assets/80077912/13f70a4b-9c90-4802-ab6a-1a96780e4628)

```sql
SELECT datetime_bin::date, count(*)
FROM wys.raw_data_2023
GROUP BY 1
ORDER BY 1
```

gabrielwol:
Number of distinct api_id with data:
![graph_visualiser-1693493287389](https://github.com/CityofToronto/bdit_data-sources/assets/80077912/ea08152c-a066-4f86-96ed-aae17a65f71a)
```sql
SELECT datetime_bin::date, count(distinct api_id)
FROM wys.raw_data_2023
GROUP BY 1
ORDER BY 1
```

gabrielwol:
It's from the locations endpoint: `/signs/statistics/location/`

radumas:
Confirming none of the signs with "Error: The location is not properly configured." Is reporting data after March 31st. 

```sql
SELECT --sign_name,
	date_trunc('month',datetime_bin) as mon,
	COUNT(DISTINCT datetime_bin::DATE) days_with_data, 
	COUNT(DISTINCT api_id)
	FROM wys.speed_counts_agg_5kph
-- 	INNER JOIN wys.locations USING (api_id)
-- 	WHERE sign_name LIKE 'F16'
	WHERE datetime_bin >= '2023-01-01' AND api_id IN 
		('2059',
'9632',
'8453',
'368',
'32875',
'27741',
'29977',
'26667',
'370',
'373',
'27707',
'29366',
'26666',
'32877',
'26665',
'26664',
'26663',
'4407',
'1745',
'26662',
'10042',
'28072',
'1750',
'28292',
'2060',
'26635',
'2786',
'8178',
'27430',
'2993',
'28293',
'27742',
'27238',
'2992',
'28294',
'26634',
'28075',
'4676',
'35681',
'4650',
'4659',
'4692',
'4657',
'4655',
'27239',
'28297',
'4656',
'26632',
'4658',
'4675',
'6866',
'4652',
'4666',
'4410',
[...]

PR #688, opened 2023-08-31
Add a Standard Slack Alert

### LINKED ISSUE #684: Add Standard Slack Alerts
Add a new function to serve as a standard Slack task-failure alert to be used by all DAGs.

tahaislam:

- Implement a standard task-failure slack alert (as a helper function) in a new file
- Update all DAGs to use the new slack alert function

## Issue(s) this solves:

- resolves #684 

- The new function in `dags/dag_functions.py` and all DAGs in `dags`

- Update the EC2 (in `/etc/airflow/data_scripts/`) and Morbius (in `/home/airflow/bdit_data-sources/`) with the new code.


gabrielwol:
I'm wondering if we can also have a standard slack message to use for warnings (or configure this one to work for that case as well), ie. to catch low data volumes or other anomalies that don't trigger an explicit failure, ie: #687 (and similar RESCU issue).

tahaislam:
> I'm wondering if we can also have a standard slack message to use for warnings (or configure this one to work for that case as well), ie. to catch low data volumes or other anomalies that don't trigger an explicit failure, ie: #687 (and similar RESCU issue).

I implemented something like what you're proposing in the VFH repo. I can add it to this repo as well

chmnata (approved):

PR #692, opened 2023-09-12
Handling Errors in WYS and VZ DAGs

### LINKED ISSUE #683: WYS `pull_from_sheet` fails with an uncaught error
If there's a problem with the connection to the spreadsheet (like in this case, because the proper URL hasn't been whitelisted), `values` is instantiated within the `try` block, but then checked later.
https://github.com/CityofToronto/bdit_data-sources/blob/42ed1cdee1520161c3811ddc4509dfe12115b5b3/wys/api/python/wys_google_sheet.py#L44-L60

 So if the number of attempts is exceeded, we the following error. Possibly not the desired outcome.

```
ERROR - Task failed with exception
Traceback (most recent call last):
  File "/data/airflow/airflow_venv/lib/python3.10/site-packages/airflow/operators/python.py", line 192, in execute
    return_value = self.execute_callable()
  File "/data/airflow/airflow_venv/lib/python3.10/site-packages/airflow/operators/python.py", line 209, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/etc/airflow/data_scripts/wys/api/python/wys_google_sheet.py", line 33, in read_masterlist
    pull_from_sheet(con, service, dict_table, dict_table[i+1])
  File "/etc/airflow/data_scripts/wys/api/python/wys_google_sheet.py", line 60, in pull_from_sheet
    if not values:
UnboundLocalError: local variable 'values' referenced before assignment
```

But also, if we were logging the `HttpError` here, we might know the right URL to whitelist.

tahaislam:

- Updates `pull_wys` DAG and the helper script `wys/api/python/wys_google_sheet.py`
- Updates `vz_google_sheets` DAG and the helper script `gis/school_safety_zones/schools.py`

## Issue(s) this solves:

- resolves #683 

- The two updated DAGs and how they handle the timeout error of Google sheets connection

Merge into `master` and update `data_scripts`

### DIFF
--- a/dags/pull_wys.py
+++ b/dags/pull_wys.py
@@ -4,12 +4,13 @@ A Slack notification is raised when the airflow process fails.
 """
 import os
 import sys
+from functools import partial
 import pendulum
+
 from airflow import DAG
-from datetime import datetime, timedelta
+from datetime import timedelta
 from airflow.operators.python_operator import PythonOperator
 from airflow.hooks.base_hook import BaseHook
-from airflow.contrib.operators.slack_webhook_operator import SlackWebhookOperator
 from airflow.providers.postgres.hooks.postgres import PostgresHook
 from airflow.models import Variable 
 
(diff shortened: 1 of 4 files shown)


gabrielwol (approved):
No comments, changes look good and it's working as per Airflow logs.

Issue #596, opened 2023-05-04
Automagically update MOVE collision views


scann0n:
We would like the MOVE collisions mat views to be updated every night after data from `ACC` is upserted into the bigdata `collisions_replicator` schema.

PR #694, opened 2023-09-13
Fix Collision DAG Name

tahaislam:

- Fix the collision DAG name
- Delete unused imports in the DAG file

## Issue(s) this solves:

- #596 

- The collision DAG file

- Update `data_scripts`


scann0n (approved):
Awesome! Thanks for removing all of those unnecessary packages!!!

PR #700, opened 2023-09-22
simplify check_partition to only check if Jan 1

gabrielwol:
- @tahaislam this is a small, non urgent PR. In a final review, I realized it would be much more elegant to simply create the partitions on Jan 1 than to query the database every day seeing if new partitions were needed. Reduced codebase by about 60 lines and I get to use a new operator 🥳 

## Issue(s) this solves:
- I was bad an did not create an issue #. 

- Is 4 sql statements in one PostgresOperator OK or do you think I need to split this into 4 tasks?

Pull updated branch into airflow.

### DIFF
--- a/dags/vds_pull_vdsdata.py
+++ b/dags/vds_pull_vdsdata.py
@@ -2,7 +2,7 @@ import os
 import sys
 from airflow import DAG
 from datetime import datetime, timedelta
-from airflow.operators.python import PythonOperator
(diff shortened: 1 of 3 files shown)


tahaislam (approved):
looks good. Have you tested this code?

gabrielwol:
> looks good. Have you tested this code?

I haven't tested the airflow changes, only the python function. What would you suggest?

PR #703, opened 2023-09-26
Dag here data home

chmnata:

- change bath operator path and add slack alert function

## Issue(s) this solves:

- #

- 

change symlink to use data_script


tahaislam (approved):

tahaislam:
I think we need to fix that `except` warning in all the DAGs in a separate PR

Issue #706, opened 2023-09-26
WYS `read_google_sheets` task is failing for certain wards trying to insert duplicate rows


gabrielwol:
Error Msg in `pull_wys.read_google_sheets` task:
```
ERROR - ON CONFLICT DO UPDATE command cannot affect row a second time
HINT:  Ensure that no rows proposed for insertion within the same command have duplicate constrained values.
```
This error has been occuring for a few months in Wards: 1, 7, 16, 22 (first silently failing and then loudly since #692 merged)

The `ON CONFLICT DO UPDATE` command occurs after removing duplicates in the `new_data`, but there apparently still duplicates??

gabrielwol:
These lines appears to be the problem
https://github.com/CityofToronto/bdit_data-sources/blob/f94c1d2ae2772266692cbb0c1deca596a96b5ccd/wys/api/python/wys_google_sheet.py#L137-L141
https://stackoverflow.com/questions/34708509/how-to-use-returning-with-on-conflict-in-postgresql

PR #709, opened 2023-09-28
Bug/i0706 wys read google sheets task is failing for certain wards trying to insert duplicate rows

gabrielwol:
- The task `read_google_sheets` of `pull_wys` was failing because certain duplicates were already present in the duplicates table, causing them to not be returned by `ON CONFLICT DO NOTHING` and then not be removed prior to the insert on wards tables. Changing to `ON CONFLICT DO UPDATE` should eliminate this issue. 

## Issue(s) this solves:
- #706 

- ON CONFLICT DO UPDATE logic

@radumas linked `pull_wys` to @gabrielwol home folder for testing purposes. We should revert that change.

### DIFF
--- a/wys/api/python/wys_google_sheet.py
+++ b/wys/api/python/wys_google_sheet.py
@@ -137,6 +137,8 @@ def pull_from_sheet(
             ON CONFLICT (location, from_street, to_street, direction, 
                         installation_date, removal_date, new_sign_number, 
                         comments)
+            /*do update required here because do nothing does not return dupes
+            on conflict. */
             DO UPDATE SET 
                 location=EXCLUDED.location,
                 from_street=EXCLUDED.from_street,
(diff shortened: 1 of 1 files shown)


gabrielwol:
@radumas once you take a look I can test it by re-running one of the failed `read_google_sheets` tasks.

gabrielwol:
I was thinking `wys.mobile_sign_installations_dupes` may warrant more attention. We never reference the contents of that table, only the new (and now modified) rows from RETURNING clause. Should we maybe truncate the table daily and make it contain only current duplicates? Or perhaps add first/last_updated date columns so we know if the values are current or outdated?

radumas:
I was expecting the work flow for that table would be:
1. Get notified of duplicates
2. Check table. 
3. Send duplicates to vision zero
4. truncate table

radumas:
I think let's merge this for now. If we find we need to make this more robust we can revisit.

radumas (approved):
Looks good.

PR #714, opened 2023-09-29
I0713 wys get schedules failing with typeerror

### LINKED ISSUE #713: WYS `get_schedules` failing with `TypeError`
Failing with following [error message](https://10.162.65.124/airflow/dags/pull_wys/grid?root=&dag_run_id=scheduled__2023-09-27T19%3A00%3A00%2B00%3A00&task_id=pull_wys&tab=logs): 
```
[2023-09-28, 23:55:06 EDT] {taskinstance.py:1943} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/data/airflow/airflow_venv/lib/python3.10/site-packages/airflow/operators/python.py", line 192, in execute
    return_value = self.execute_callable()
  File "/data/airflow/airflow_venv/lib/python3.10/site-packages/airflow/operators/python.py", line 209, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/data/airflow/dev_scripts/wys/api/python/wys_api.py", line 380, in api_main
    get_schedules(conn, api_key)
  File "/data/airflow/dev_scripts/wys/api/python/wys_api.py", line 475, in get_schedules
    rows = [(schedule['name'], api_id) for schedule in schedule_list if schedule['assigned_on_locations']
  File "/data/airflow/dev_scripts/wys/api/python/wys_api.py", line 475, in <listcomp>
    rows = [(schedule['name'], api_id) for schedule in schedule_list if schedule['assigned_on_locations']
TypeError: string indices must be integers
```
Seems to be related to a change in API response value. 
Related to #629, this small, standalone function is called is at the end of `pull_wys` task (1 hour run time), causing that task to fail. As part of this bug fix we should consider separating it into it's own Airflow task.

gabrielwol:
- Separates `get_schedules` into a separate Airflow task to prevent failure from impacting rest of `pull_wys` task.
- Adds try except block to `get_schedules` to catch TypeError
- Change all tasks to taskflow api format
- Move connections out of top level to improve DAG parsing performance

## Issue(s) this solves:
- Closes #713 

- First time using TaskFlow API!
- See error in #713

- I haven't tested parsing the changes to the DAG

### DIFF
--- a/dags/pull_wys.py
+++ b/dags/pull_wys.py
@@ -80,28 +80,26 @@ default_args = {'owner': ','.join(names),
                  'on_failure_callback': task_fail_slack_alert
                 }
 
-dag = DAG(dag_id = dag_name, default_args=default_args, schedule_interval='0 15 * * *')
+with DAG(dag_id = dag_name, default_args=default_args, schedule_interval='0 15 * * *') as dag:
 # Run at 3 PM local time every day
 
 with wys_postgres.get_conn() as con:
     task_pull_wys = PythonOperator(
             task_id = 'pull_wys',
             python_callable = api_main, 
-            dag = dag,
             op_kwargs = {'conn':con, 
                         'start_date':'{{ ds }}', 
                         'end_date':'{{ ds }}', 
                         'api_key':api_key}
                         )
     
-    @task
+    @task()
     def task_get_schedules():
-        get_schedules(wys_postgres, api_key)
+        get_schedules(wys_postgres, connection)
 
     task_read_google_sheets = PythonOperator(
             task_id = 'read_google_sheets',
             python_callable = read_masterlist,
-            dag = dag,
             op_args = [con, service],
             on_failure_callback = partial(
                 task_fail_slack_alert, extra_msg=custom_fail_slack_alert
--- a/dags/pull_wys.py
+++ b/dags/pull_wys.py
@@ -13,6 +13,7 @@ from airflow.operators.python_operator import PythonOperator
 from airflow.hooks.base_hook import BaseHook
 from airflow.providers.postgres.hooks.postgres import PostgresHook
 from airflow.models import Variable 
+from airflow.decorators import task
 
 from airflow.providers.google.common.hooks.base_google import GoogleBaseHook
 from googleapiclient.discovery import build
@@ -21,7 +22,7 @@ from dateutil.relativedelta import relativedelta
 try:
     repo_path = os.path.abspath(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
     sys.path.insert(0, repo_path)
-    from wys.api.python.wys_api import api_main
+    from wys.api.python.wys_api import api_main, get_schedules
     from wys.api.python.wys_google_sheet import read_masterlist
     from dags.dag_functions import task_fail_slack_alert
 except:
@@ -83,7 +84,7 @@ dag = DAG(dag_id = dag_name, default_args=default_args, schedule_interval='0 15 
 # Run at 3 PM local time every day
 
 with wys_postgres.get_conn() as con:
-    t1 = PythonOperator(
+    task_pull_wys = PythonOperator(
             task_id = 'pull_wys',
             python_callable = api_main, 
             dag = dag,
@@ -93,7 +94,11 @@ with wys_postgres.get_conn() as con:
                         'api_key':api_key}
                         )
     
-    t2 = PythonOperator(
+    @task
+    def task_get_schedules():
+        get_schedules(wys_postgres, api_key)
+
+    task_read_google_sheets = PythonOperator(
             task_id = 'read_google_sheets',
             python_callable = read_masterlist,
             dag = dag,
@@ -102,3 +107,7 @@ with wys_postgres.get_conn() as con:
                 task_fail_slack_alert, extra_msg=custom_fail_slack_alert
             )
     )
+
+    task_pull_wys
+    task_get_schedules()
+    task_read_google_sheets
--- a/dags/pull_wys.py
+++ b/dags/pull_wys.py
@@ -17,7 +17,6 @@ from airflow.decorators import task
 
 from airflow.providers.google.common.hooks.base_google import GoogleBaseHook
 from googleapiclient.discovery import build
-from dateutil.relativedelta import relativedelta
 
 try:
     repo_path = os.path.abspath(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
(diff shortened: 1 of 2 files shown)


wys/api/python/wys_api.py
```diff
@@ -478,9 +467,15 @@ def get_schedules(conn, api_key):
                          headers=headers)
     schedule_list = response.json()
```
radumas, line 468: My comment was a bit slow.

And realizing, given the response you output, that it wouldn't handle this specific error. But I think we should also add the `raise_for_status()` here. https://github.com/CityofToronto/bdit_data-sources/issues/713#issuecomment-1741418875

dags/pull_wys.py
```diff
@@ -83,7 +84,7 @@ def custom_fail_slack_alert(context: dict) -> str:
 # Run at 3 PM local time every day
 
 with wys_postgres.get_conn() as con:
```
radumas, line 86: this is a bit of an unideal pattern that we should've gotten rid of before because this connection gets opened every time the schedule refreshes this file. 

might as well move fully towards the TaskFlow API paradigm and only open the connection within the decorated functions

dags/pull_wys.py
```diff
@@ -93,7 +94,11 @@ def custom_fail_slack_alert(context: dict) -> str:
                         'api_key':api_key}
                         )
     
-    t2 = PythonOperator(
+    @task
```
radumas, line 97: You should add the on_failure_callback here

https://stackoverflow.com/a/76718651/4047679

wys/api/python/wys_api.py
```diff
@@ -463,16 +463,22 @@ def update_locations(conn, loc_table):
...
-    schedule_list = response.json()
     
+    try: 
+        response=session.get(url+signs_endpoint+schedule_endpoint,
+                         headers=headers)
+        response.raise_for_status()
+        schedule_list = response.json()
+    except RequestException as exc: 
+        logger.critical('Error querying API, %s', exc) 
+        sys.exit(2)
```
radumas, line 474: You don't need the `sys.exit` here if only this function is getting called.

dags/pull_wys.py
radumas: Let's just test opening the connection within the decorated function for the get_schedules task.

dags/pull_wys.py
```diff
@@ -93,7 +94,11 @@ def custom_fail_slack_alert(context: dict) -> str:
...
+    def task_get_schedules():
+        get_schedules(wys_postgres, api_key)
```
radumas, line 99: Same thing with pulling the `api_key` from the variable. That should happen inside the decorated function and not outside.

radumas (changes requested):
I think your implementation of the `@task` decorator is correct. Just adding some improvements

dags/pull_wys.py
```diff
@@ -80,28 +80,26 @@ def custom_fail_slack_alert(context: dict) -> str:
                  'on_failure_callback': task_fail_slack_alert
                 }
 
-dag = DAG(dag_id = dag_name, default_args=default_args, schedule_interval='0 15 * * *')
+with DAG(dag_id = dag_name, default_args=default_args, schedule_interval='0 15 * * *') as dag:
```
radumas, line 83: everything after this will need to be tabbed over

wys/api/python/wys_api.py
```diff
@@ -470,19 +471,19 @@ def get_schedules(conn, api_key):
...
                     for schedule in schedule_list if schedule['assigned_on_locations']
                         for api_id in schedule['assigned_on_locations'] if api_id]
     except TypeError as e:
         logger.critical('Error converting schedules response to values list.')
-        logger.critical('Return value: ', schedule_list)
+        logger.critical('Return value: %s', schedule_list)
         raise WYS_APIException(e)
 
-    with conn.cursor() as cur:
+    with conn.get_conn() as con, con.cursor() as cur:
```
radumas, line 486: this connection opening should be moved back to the dag file but inside the decorated task function

wys/api/python/wys_api.py
```diff
@@ -461,7 +461,8 @@ def update_locations(conn, loc_table):
         """
         cur.execute(update_locations_sql)
 
-def get_schedules(conn, api_key):
+def get_schedules(conn, api_connection):
+    api_key = api_connection.password
```
radumas, line 465: this should be moved back to the dag file but inside the decorated task function

radumas (changes requested):
Apologies I wasn't super clear on this, and I don't find the examples to be super explicit. My understanding is when a connection or a variable is defined in top-level code (fully untabbed), then Airflow is connecting to the metadata database to pull that. Though apparently there's caching now that can help with that.

https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#airflow-variables

In order to make it a bit easier to debug or play around with Python functions in an interactive session, I would recommend moving the Airflow specific calls to the dag-file so you don't have to construct an object with a `get_conn()` method or a `password` attribute in order to test the `get_schedules()` function.

So in the DAG you would have something like:

```python
    @task()
    def task_get_schedules():
        api_key = api_connection.password
        with wys_postgres.get_conn() as conn:
            get_schedules(wys_postgres, api_key)
```

dags/pull_wys.py
radumas: Yes you are correct.

dags/pull_wys.py
```diff
@@ -79,26 +67,56 @@ def custom_fail_slack_alert(context: dict) -> str:
...
+        #to connect to pgadmin bot
+        wys_postgres = PostgresHook("wys_bot")
+        
+        #api connection
+        api_conn = BaseHook.get_connection('wys_api_key')
+        api_key = api_conn.password
+
+        with wys_postgres.get_conn() as conn:
+            api_main(start_date = kwargs['ds'],
+                     end_date = kwargs['ds'],
```
radumas, line 87: I think this can be simplified by replacing the `**kwargs` argument with `(ds=None)` in the function definition and then using the `ds` variable directly. 

No need to make the modification. I've made it and will commit and push after it runs as scheduled at 15:00.

radumas:
Hmmm

```python
[2023-10-02, 15:22:14 EDT] {base.py:73} INFO - Using connection ID 'vz_api_google' for task execution.
[2023-10-02, 15:22:14 EDT] {base.py:73} INFO - Using connection ID 'wys_bot' for task execution.
[2023-10-02, 15:22:14 EDT] {taskinstance.py:1943} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/data/airflow/airflow_venv/lib/python3.10/site-packages/airflow/decorators/base.py", line 221, in execute
    return_value = super().execute(context)
  File "/data/airflow/airflow_venv/lib/python3.10/site-packages/airflow/operators/python.py", line 192, in execute
    return_value = self.execute_callable()
  File "/data/airflow/airflow_venv/lib/python3.10/site-packages/airflow/operators/python.py", line 209, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/data/airflow/dags/pull_wys.py", line 116, in read_google_sheets
    read_masterlist(con, service)
  File "/data/home/rdumas/git/bdit_data-sources/wys/api/python/wys_google_sheet.py", line 27, in read_masterlist
    with con:
psycopg2.ProgrammingError: the connection cannot be re-entered recursively
```

radumas (approved):
Thanks Gabe! Now we have a shining example of the new `@dag/@task` decoration

gabrielwol:
Thanks for all your comments/fixes!

Issue #716, opened 2023-10-03
Add a data volume check to pull_wys task


gabrielwol:
pull_wys.pull_wys sometimes unpredictably pulls very low data volumes, including on different retries for the same date. Add a check to ensure expected data volume is being pulled, otherwise task should fail. 

eg: 

- 1063 locations, 1,184,048 rows on retry 4
- 12 locations, 6,604 rows on retry 5

![image](https://github.com/CityofToronto/bdit_data-sources/assets/80077912/6cab5b55-9526-4dcc-bdee-62447cd3278b)
![image (1)](https://github.com/CityofToronto/bdit_data-sources/assets/80077912/1d6a36c9-5c41-4836-80ad-be35e7e0deb5)

Issue #702, opened 2023-09-26
Port over `check_rescu` functionality to `vds`


gabrielwol:
Port over functionality from `check_rescu` DAG to new `vds` schema.

- Change to rolling lookback rather than fixed value. 
- Implement on `raw_vdsdata` and `raw_vdsvehicledata`

radumas:
Noting some occasional performance issues with this (e.g. yesterday). Possibly related to when `VACUUM`ing happens, related to https://github.com/CityofToronto/bdit_data-sources/issues/469 ?

gabrielwol:
> Noting some occasional performance issues with this (e.g. yesterday). Possibly related to when `VACUUM`ing happens, related to #469 ?

Note that there was no auto-vacuum on the queried tables in the intervening time between the slow attempts and the next day's normal speed.

gabrielwol:
Was closed by #705

PR #659, opened 2023-07-31
Add Long-term Mobile WYS Signs

tahaislam:
## What this pull request accomplishes:

- Pulls the `work_order` of each sign and it to `wys.mobile_installations`
- Allows mobile signs with `removal_date=None` to be inserted into `wys.mobile_installations` in addition to `mobile_detailed` and `mobile_summary` views and their corresponding Open Data views.
- Updates filters on `wys.mobile_summary`, `open_data.wys_mobile_detailed`, `open_data.wys_mobile_summary`

## Issue(s) this solves:

- #534, #726

## What, in particular, needs to reviewed:

- `wys_google_sheet` file, which is responsible of Airflow task: `read_google_sheets`
- SQL scripts of tables `wys.mobile_sign_installations_dupes` and `wys.mobile_sign_installations`
- SQL scripts of Open Data views `wys_mobile_detailed` and `wys_mobile_summary`

## What needs to be done by a sysadmin after this PR is merged

- [x] Add the new column `work_order` to tables `wys.mobile_sign_installations_dupes` and `wys.mobile_sign_installations`
- [x] Clear a `read_google_sheets` task
- [x] Update permissions:  
    ```sql
    REVOKE ALL ON TABLE wys.mobile_sign_installations FROM jchew;
    REVOKE ALL ON TABLE open_data.wys_mobile_detailed FROM rdumas;
    REVOKE DELETE ON TABLE wys.mobile_sign_installations FROM rdumas;
    REVOKE ALL ON TABLE open_data.wys_stationary_summary FROM rdumas;
    REVOKE ALL ON TABLE open_data.wys_stationary_detailed FROM rdumas;
    
    ALTER FUNCTION wys.aggregate_speed_counts_one_hour_5kph(date) OWNER TO wys_admins;
    ALTER TABLE open_data.wys_stationary_summary OWNER TO wys_admins;
    ALTER TABLE open_data.wys_stationary_detailed OWNER TO wys_admins;
    ALTER TABLE open_data.wys_mobile_detailed OWNER TO wys_admins;
    ALTER TABLE wys.speed_counts_agg_5kph OWNER TO wys_admins;
    ALTER TABLE wys.raw_data_2017 OWNER to wys_admins;
    ALTER TABLE wys.raw_data_2018 OWNER to wys_admins;
    ALTER TABLE wys.raw_data_2019 OWNER to wys_admins;
    ALTER TABLE wys.raw_data_2020 OWNER to wys_admins;
[...]

PR #725, opened 2023-10-20
716 add a data volume check to pull wys task

gabrielwol:
- Adds a row count check to the daily pull_wys DAG
- Adds a distinct sign count check to the daily pull_wys DAG (DISTINCT api_id) 
- This uses the same sql/airflow template as merged ~~#702~~ **Updated:** #730 . 

## Issue(s) this solves:
- #716 

- The params used (60 day lookback, 0.7 threshold)

- It may be easier to wait to merge #659 first, as we are already running significant modifications to the DAG off that branch.


radumas (approved):
This looks great! Really solid emphasis on generalizable building blocks we can apply to other pipelines.

We could generalize the api_id count lookback to receive something like `sensor_id_col` as a param and then that could be potentially applied to allll (many?) of the sensor based data streams.

gabrielwol:
Thanks! Note that `SQLCheckOperatorWithReturnValue` was first added by Islam for VFH. 
Good idea, I'll add `sensor_id_col` to params.

gabrielwol:
Noting the potential for issue #730 to manifest on this DAG as well as they share similar SQL.

gabrielwol:
Fixed the failures on this task from Saturday/Sunday by adding template search path. Also adjusted the sqls to match #732: query aggregate table instead of raw. `[Tested](https://10.162.65.124/airflow/dags/pull_wys/grid?base_date=2023-10-28T23%3A59%3A59.999Z&dag_run_id=scheduled__2023-10-27T19%3A00%3A00%2B00%3A00&task_id=data_checks.check_row_count&tab=logs)` ✅ .

Issue #730, opened 2023-10-26
check_rows_vdsdata_* tasks unpredictable duration


gabrielwol:
These two task instances ran for > 5 hours before being cancelled.
[data_checks.check_rows_vdsdata_div2 for 2023-10-25](https://morbius.corp.toronto.ca:8080/airflow/dags/vds_pull_vdsdata/grid?base_date=2023-10-25T23%3A59%3A59.999Z&dag_run_id=scheduled__2023-10-25T08%3A00%3A00%2B00%3A00&task_id=data_checks.check_rows_vdsdata_div2)
[data_checks.check_rows_vdsdata_div8001 for 2023-10-23](https://morbius.corp.toronto.ca:8080/airflow/dags/vds_pull_vdsdata/grid?base_date=2023-10-25T23%3A59%3A59.999Z&dag_run_id=scheduled__2023-10-23T08%3A00%3A00%2B00%3A00&task_id=data_checks.check_rows_vdsdata_div8001)
- The other task instances ran in between 8-35s.
- There were no auto/manual vacuums on these tables between slow and fast task executions.
- The explain plan for the slow queries show it is doing a sequence scan over the correct, relatively small partitions, totalling ~10GB. 
- A similar data check method was implemented for WYS in #716, which has not yet been merged.
- Task execution_timeout did not work as expected. I think [statement_timeout](https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html#the-complete-postgres-operator-dag) is the correct way to cancel execution of a postgres operator.

gabrielwol:
I don't see the same variation in [past task instances](https://morbius.corp.toronto.ca:8080/airflow/taskinstance/list/?_flt_3_operator=SQLCheckOperatorWithReturnValue&_oc_TaskInstanceModelView=duration&_od_TaskInstanceModelView=desc&page_TaskInstanceModelView=11) of the SQLCheckOperatorWithReturnValue used in VFH.

PR #732, opened 2023-10-27
730 check rows vdsdata tasks unpredictable duration

gabrielwol:
- Performance improvements for the check data tasks of VDS DAGs. 
- I guess there was some instability to the method `FROM large_table, LATERAL (SELECT X from small_CTE)` since it ran alternately in 10s/>24hrs. I changed this to a cross join between two very small CTE and it is running in <1s.  

## Issue(s) this solves:
- #730

- I removed the check on div_8001 (Non rescu part of the VDS network) because it is not summarized ever. Alternatively I could use the method mentioned in 2nd bullet point above but go back to querying raw table - or have one query type for checks on raw tables and one for summary tables. 

Nothing


dags/vds_pull_vdsdata.py
```diff
@@ -132,19 +132,18 @@
...
     with TaskGroup(group_id='data_checks') as data_checks:
-        divisions = [2, 8001]
+        divisions = [2, ] #div8001 is never summarized and the query on the view is not optimized
         for divid in divisions:
             check_avg_rows = SQLCheckOperatorWithReturnValue(
                 task_id=f"check_rows_vdsdata_div{divid}",
                 sql="select/select-row_count_lookback.sql",
                 conn_id='vds_bot',
-                params={"table": 'vds.raw_vdsdata',
+                params={"table": f'vds.counts_15min_div{divid}',
```
tahaislam, line 141: Not specifically for this PR, but I think it's better to use either double or single quotes but not both. Most linters/standards prefer double quotes though

volumes/vds/sql/select/select-row_count_lookback.sql
```diff
@@ -5,31 +5,33 @@
...
-        SELECT AVG(lookback_count) lookback_avg FROM lookback
-    ) AS lb
-WHERE
-    a.{{ params.dt_col }} >= '{{ ds }} 00:00:00'::timestamp
-    AND a.{{ params.dt_col }} < '{{ ds }} 00:00:00'::timestamp + interval '1 day'
-    AND a.division_id = {{ params.div_id }}::int
-GROUP BY lb.lookback_avg
\ No newline at end of file
+FROM today AS a,
+    lookback AS lb
```
tahaislam, line 36: `lb` is used in line 32. From where did this warning come? Have you tested the code? Could it be misspelled above, e.g., `1` instead of `l`?

tahaislam (commented):
Looks good to me. Just one comment on that sqlfluff warning

gabrielwol:
Small change: I moved the file to a common location to be used by other dags under `dags/sql` and updated the template_searchpath. I'd say this is ready to merge!

tahaislam (approved):

Issue #717, opened 2023-10-04
Partition aggregated Miovision volumes tables?


Nate-Wessel:
I'm getting pretty bad performance while trying to query these tables (e.g. `miovision_api.volumes_15min_mvt` and `miovision_api.volumes_15min`) across all available dates. Even for a query that subsets the data to a single intersection and classification, which is indexed and should constitute a small subset of all data, I still see the query planner trying to do a full sequential scan in many (but not all) cases. I'm assuming this is because the data are inserted more or less sequentially, so the intersections and classifications aren't really clustered together on the disk and it knows it will have to read all the pages anyway. 

Wondering if a partition by classification and/or intersection would help performance here?

My thought it that a typical query would only hit a subset of classifications and intersections (and to be fair would likely have a limited date range as well).

Current unpartitioned tables sizes:
* `miovision_api.volumes_15min_mvt`: ~350m rows
* `miovision_api.volumes_15min`: ~278m rows

Sample query that seems like it really ought to be faster:
```sql
EXPLAIN SELECT 
    datetime_bin::date AS date,
    SUM(volume) AS count
FROM miovision_api.volumes_15min_mvt
WHERE
    intersection_uid = 2 -- indexed
    AND classification_uid = 2 -- also indexed
GROUP BY date
ORDER BY date
```

Or is there a better way of structuring this query?

gabrielwol:
The fact that your query is not making use of the indices likely has to do with the volume of data you are selecting. Adding a date filter makes use of the unique constraint and runs in about 2s:
```sql
AND datetime_bin >= '2023-01-01'::date
```
If we partitioned by intersection (x68 distinct) and/or classification (x9 distinct), it would make queries that cut across those fields slower, so we'd have to weigh the different types of queries we make. Perhaps by classifications but grouped: cars, bikes, else?

Nate-Wessel:
I think you're right. But here's some strangeness:

```sql
SELECT 
    datetime_bin::date AS date,
    SUM(volume) AS count
FROM miovision_api.volumes_15min_mvt
WHERE
    intersection_uid = 1 -- indexed
    AND classification_uid = 2 -- also indexed
    AND datetime_bin < '2021-01-01'
GROUP BY date
--ORDER BY date

UNION

SELECT 
    datetime_bin::date AS date,
    SUM(volume) AS count
FROM miovision_api.volumes_15min_mvt
WHERE
    intersection_uid = 1 -- indexed
[...]

PR #735, opened 2023-10-31
717 miovision partition volumes and daily summary

gabrielwol:
1. Partitions `miovision_api.volumes_15min` and `miovision_api.volumes_15min_mvt` by year  
    - Both were around 20GB and growing.
    - This will help with many of our common DRs as they query a small date slice of one of these tables.
3. Adding `volumes_daily` table grouped by date + intersection, with classification volumes spread into columns. 
    - This will help a lot on any long time range queries including sensor health checks (ie. are they working, are they reporting bike volumes). 
4. Change volumes table from inheritance to declarative partitioning. 
    - This will clean up the schema a bit when browsing and comes with some other improvements, like faster insert performance. 
    - Simultaneously partitions by year -> month as yearly partitions hover around 20GB for a year and data volume is only growing.
5. Added the partition create statements to the pull_miovision DAG instead of eoy_table_create.
    - I think this is preferable to have the pull data task fail if the partition create task fails. (this could also be done with an external task sensor, but they would be on different schedules (yearly/daily) which seems trickier).  
    - I tried something a little new which is creating a partition each year and each month! Main benefit is less database clutter in the form of empty partitions. 

## Issue(s) this solves:
- #717

- 26,355 records from miovision_api.volumes with classification_uid NULL excluded between: "2021-07-21 15:02:00", "2022-05-06 11:39:00" (violates new Primary key).
- Should we use ["save and drop dependencies"](https://github.com/CityofToronto/bdit_pgutils/pull/9) function.

**Update DAGs:**
- [ ] dags/eoy_create_tables.py
- [ ] dags/pull_miovision.py

**Create new tables/functions:**
- [x] volumes/miovision/sql/create-table-volumes_daily.sql
- [x] volumes/miovision/sql/function-aggregate-volumes_daily.sql
- [x] volumes/miovision/sql/function-create_mm_nested_volumes_partitions.sql
- [x] volumes/miovision/sql/function-create_yyyy_volumes_15min_partition.sql
- [x] volumes/miovision/sql/function-create_yyyy_volumes_partition.sql

**Run second half of these 3 scripts to complete changeover:** 
- [ ] volumes/miovision/sql/update-mio_volumes15min_mvt_partitioning.sql
- [ ] volumes/miovision/sql/update-mio_volumes15min_partitioning.sql
- [x] volumes/miovision/sql/update-mio_volumes_partitioning.sql

**Delete functions:** 
- [x] volumes/miovision/sql/function_create_volume_table.sql
- [x] volumes/miovision/sql/trigger-volumes_insert_trigger.sql

## To Do:
- [x] Update eoy_create_tables DAG
- [x] Add create partition scripts to miovision DAG
- [x] daily insert script for miovision.volumes_daily
- [x] Confirm ok to exclude 2018-12-31 11pm - midnight?


gabrielwol:
Note I excluded these rows from the new volumes table.

```sql
SELECT MIN(datetime_bin), MAX(datetime_bin), COUNT(*)
FROM miovision_api.volumes
WHERE classification_uid IS NULL --These rows violated new primary key.
```

"min"	"max"	"count"
"2021-07-21 15:02:00"	"2022-05-06 11:39:00"	26355

radumas:
> Note I excluded these rows from the new volumes table.
> 
> ```sql
> SELECT MIN(datetime_bin), MAX(datetime_bin), COUNT(*)
> FROM miovision_api.volumes
> WHERE classification_uid IS NULL --These rows violated new primary key.
> ```
> 
> "min" "max" "count" "2021-07-21 15:02:00" "2022-05-06 11:39:00" 26355

I think these might be what we were referring to as pedestrians in roadway but are actually cyclist approach volumes per a recent email.

the response from the API for that final minute is

```json
[
  {
    "timestamp": "2022-05-06T11:39:00.000-04:00",
    "class": "Bicycle",
    "entrance": "N",
    "exit": "S",
    "qty": 2
  },
  {
    "timestamp": "2022-05-06T11:39:00.000-04:00",
    "class": "Bicycle",
    "entrance": "N",
    "exit": "UNDEFINED",
    "qty": 1
  },
  {
    "timestamp": "2022-05-06T11:39:00.000-04:00",
    "class": "Light",
    "entrance": "W",
    "exit": "N",
    "qty": 7
  },
  {
    "timestamp": "2022-05-06T11:39:00.000-04:00",
    "class": "Pedestrian",
    "entrance": "N",
    "exit": "UNDEFINED",
    "qty": 1
  }
]
```

And what we have in our DB
```sql
SELECT volume_uid, intersection_uid, datetime_bin, classification_uid, leg, movement_uid, volume, volume_15min_mvt_uid
	FROM miovision_api.volumes
	WHERE intersection_uid = 65 AND datetime_bin = '2022-05-06 11:39:00'
```

image

@chmnata, could they have been inserted when Jedwin was backfilling that intersection because of the pedestrian error. I'm kind of puzzled they slipped through given the entire pipeline had been failing for the roadway peds.

gabrielwol:
Just re-posting my finding from Friday that all the null classification_uid values are from intersection_uid 65, movement_uid 7. 
image

volumes/miovision/sql/create-table-volumes.sql
```diff
@@ -1,129 +1,60 @@
...
-
-DO $do$
-DECLARE
-	startdate DATE;
-	yyyy TEXT;
-	basetablename TEXT := 'volumes_';
-	tablename TEXT;
-BEGIN
+ON miovision_api.volumes
+USING btree(volume_15min_mvt_uid);
```
chmnata, line 51: kinda wondering if we need this 🤔

volumes/miovision/sql/create-table-volumes_15min.sql
```diff
@@ -3,66 +3,70 @@
...
 );
-ALTER TABLE miovision_api.volumes_15min
-  OWNER TO miovision_admins;
+
+ALTER TABLE miovision_api.volumes_15min OWNER TO miovision_admins;
 GRANT ALL ON TABLE miovision_api.volumes_15min TO rds_superuser WITH GRANT OPTION;
 GRANT ALL ON TABLE miovision_api.volumes_15min TO dbadmin;
+REVOKE ALL ON TABLE miovision_api.volumes_15min FROM bdit_humans;
 GRANT SELECT, REFERENCES, TRIGGER ON TABLE miovision_api.volumes_15min TO bdit_humans WITH GRANT OPTION;
+REVOKE ALL ON TABLE miovision_api.volumes_15min FROM covid_admins;
```
chmnata, line 26: I think we can get rid of this meow, cause all covid_admins are humans so they should be able to have select reference and trigger already

volumes/miovision/sql/create-table-volumes_15min.sql
```diff
@@ -3,66 +3,70 @@
...
+GRANT SELECT ON TABLE miovision_api.volumes_15min TO covid_admins;
+GRANT ALL ON TABLE miovision_api.volumes_15min TO bdit_bots;
```
chmnata, line 28: grant all to bots is kinda sus, maybe we just need to for miovision_bot?

gabrielwol:
There were a lot of seemingly unnecessary permissions! Let's axe them!

volumes/miovision/sql/create-table-volumes_15min.sql
```diff
@@ -3,66 +3,70 @@
...
 
--- Index: miovision_api.volumes_15min_classification_uid_idx
+COMMENT ON COLUMN miovision_api.volumes_15min.leg
+IS 'leg location, e.g. E leg means both entry and exit traffic across the east side of the intersection. ';
```
chmnata, line 31: such an important comment

volumes/miovision/sql/create-table-volumes.sql
gabrielwol: Will remove.

volumes/miovision/sql/function-create_mm_nested_volumes_partitions.sql
```diff
@@ -0,0 +1,54 @@
...
+    );
+END;
+$BODY$;
+
+COMMENT ON FUNCTION miovision_api.create_mm_nested_volumes_partitions(text, integer, integer) IS
+'Create a new month partition under the parent year table `base_table`.
+Only to be used for miovision_api `volumes_15min` and `volumes_15min_mvt` tables. 
+Example: SELECT miovision_api.create_yyyy_volumes_partition(''volumes_15min'', 2023)';
+
+ALTER FUNCTION miovision_api.create_mm_nested_volumes_partitions(text, integer, integer) OWNER TO vds_admins;
```
chmnata, line 53: vds alert!

volumes/miovision/sql/update-mio_volumes15min_mvt_partitioning.sql
```diff
@@ -0,0 +1,155 @@
...
+ON mio_staging.volumes_15min_mvt USING btree(volume_15min_mvt_uid ASC nulls last);
+    
+--create partitions
+--note this partition create function has been altered since.
+DO $do$
+DECLARE
+	yyyy TEXT;
+BEGIN
+	FOR yyyy IN 2019..2023 LOOP
+        EXECUTE FORMAT($$SELECT gwolofs.create_yyyy_partition(
```
chmnata, line 54: should this be another schema?

volumes/miovision/sql/function-create_mm_nested_volumes_partitions.sql
```diff
@@ -0,0 +1,54 @@
...
+    end_mm:= start_mm + INTERVAL '1 month';
+    month_table:= year_table||mm_pad;
+    EXECUTE FORMAT($$
+            CREATE TABLE IF NOT EXISTS miovision_api.%I
+            PARTITION OF miovision_api.%I
+            FOR VALUES FROM (%L) TO (%L);
+            ALTER TABLE IF EXISTS miovision_api.%I OWNER TO miovision_admins;
+            REVOKE ALL ON TABLE miovision_api.%I FROM bdit_humans;
+            GRANT TRIGGER, SELECT, REFERENCES ON TABLE miovision_api.%I TO bdit_humans WITH GRANT OPTION;
+            GRANT ALL ON TABLE miovision_api.%I TO bdit_bots;
```
chmnata, line 32: remove dis plz

volumes/miovision/sql/function-create_mm_nested_volumes_partitions.sql
```diff
@@ -0,0 +1,54 @@
...
+            GRANT ALL ON TABLE miovision_api.%I TO rds_superuser WITH GRANT OPTION;
```
chmnata, line 33: remove dis plz

volumes/miovision/sql/function-create_yyyy_volumes_15min_partition.sql
```diff
@@ -0,0 +1,50 @@
...
+BEGIN
+
+    EXECUTE FORMAT($$
+        CREATE TABLE IF NOT EXISTS miovision_api.%I
+        PARTITION OF miovision_api.%I
+        FOR VALUES FROM (%L) TO (%L);
+        ALTER TABLE IF EXISTS miovision_api.%I OWNER TO miovision_admins;
+        REVOKE ALL ON TABLE miovision_api.%I FROM bdit_humans;
+        GRANT TRIGGER, SELECT, REFERENCES ON TABLE miovision_api.%I TO bdit_humans WITH GRANT OPTION;
+        GRANT ALL ON TABLE miovision_api.%I TO bdit_bots;
```
chmnata, line 27: remove dis plz

volumes/miovision/sql/function-create_yyyy_volumes_15min_partition.sql
```diff
@@ -0,0 +1,50 @@
...
+        GRANT ALL ON TABLE miovision_api.%I TO rds_superuser WITH GRANT OPTION;
```
chmnata, line 28: remove dis plz

volumes/miovision/sql/update-mio_volumes15min_mvt_partitioning.sql
gabrielwol: This is fine, it was just the temp schema of the function at the I ran it.

volumes/miovision/sql/function-create_yyyy_volumes_partition.sql
```diff
@@ -0,0 +1,49 @@
...
+BEGIN
+
+    EXECUTE FORMAT($$
+        CREATE TABLE IF NOT EXISTS miovision_api.%I
+        PARTITION OF miovision_api.%I
+        FOR VALUES FROM (%L) TO (%L)
+        PARTITION BY RANGE (%I);
+        ALTER TABLE IF EXISTS miovision_api.%I OWNER TO miovision_admins;
+        GRANT TRIGGER, SELECT, INSERT ON TABLE miovision_api.%I TO miovision_api_bot;
+        GRANT TRIGGER, SELECT, REFERENCES ON TABLE mio_staging.%I TO bdit_humans WITH GRANT OPTION;
```
chmnata, line 27: this should be miovision_api schema

volumes/miovision/sql/create-table-volumes_15min.sql
```diff
@@ -3,66 +3,68 @@
...
 WITH (
   OIDS=FALSE
 );
-ALTER TABLE miovision_api.volumes_15min
-  OWNER TO miovision_admins;
+
+ALTER TABLE miovision_api.volumes_15min OWNER TO miovision_admins;
 GRANT ALL ON TABLE miovision_api.volumes_15min TO rds_superuser WITH GRANT OPTION;
 GRANT ALL ON TABLE miovision_api.volumes_15min TO dbadmin;
+REVOKE ALL ON TABLE miovision_api.volumes_15min FROM bdit_humans;
```
chmnata, line 24: lets remove this as well

volumes/miovision/sql/create-table-volumes_15min_mvt.sql
```diff
@@ -1,67 +1,65 @@
...
-  CONSTRAINT volumes_15min_mvt_intersection_uid_datetime_bin_classificat_key UNIQUE (intersection_uid, datetime_bin, classification_uid, leg, movement_uid)
+    volume_15min_mvt_uid serial NOT NULL,
+    intersection_uid integer,
+    datetime_bin timestamp without time zone,
+    classification_uid integer,
+    leg text,
+    movement_uid integer,
+    volume numeric,
+    processed boolean,
+    CONSTRAINT volumes_15min_mvt_intersection_uid_datetime_bin_classificat_pk
```
chmnata, line 11: lets keep the naming consistent and use `pkey` instead of `pk`

volumes/miovision/sql/create-table-volumes_15min_mvt.sql
```diff
@@ -1,67 +1,65 @@
...
-ALTER TABLE miovision_api.volumes_15min_mvt
-  OWNER TO miovision_admins;
+
+ALTER TABLE miovision_api.volumes_15min_mvt OWNER TO miovision_admins;
 GRANT ALL ON TABLE miovision_api.volumes_15min_mvt TO rds_superuser WITH GRANT OPTION;
 GRANT ALL ON TABLE miovision_api.volumes_15min_mvt TO dbadmin;
+REVOKE ALL ON TABLE miovision_api.volumes_15min_mvt FROM bdit_humans;
 GRANT SELECT, REFERENCES, TRIGGER ON TABLE miovision_api.volumes_15min_mvt TO bdit_humans WITH GRANT OPTION;
+GRANT SELECT, INSERT, TRIGGER ON TABLE miovision_api.volumes_15min_mvt TO miovision_api_bot;
+GRANT ALL ON TABLE miovision_api.volumes_15min_mvt TO rds_superuser WITH GRANT OPTION;
```
chmnata, line 25: dupppped

volumes/miovision/sql/create-table-volumes_15min_mvt.sql
```diff
@@ -1,67 +1,65 @@
...
 CREATE INDEX volumes_15min_mvt_volume_15min_mvt_uid_idx
-  ON miovision_api.volumes_15min_mvt
-  USING btree
-  (volume_15min_mvt_uid);
-
+ON miovision_api.volumes_15min_mvt
+USING btree(volume_15min_mvt_uid ASC nulls last);
+
+--create yearly partitions
+SELECT miovision_api.create_yyyy_volumes_15min_partition('volumes_15min_mvt', 2019::int)
```
chmnata, line 61: i think either get rid of this block or add `;` at the end of each select

volumes/miovision/sql/create-table-volumes_daily.sql
```diff
@@ -0,0 +1,49 @@
...
+    volume_8 integer, 
+    volume_9 integer, 
+    volume_10 integer, 
+    volume_total integer,
+    CONSTRAINT volumes_daily_pkey
+    PRIMARY KEY (intersection_uid, dt)
+);
+
+CREATE INDEX volumes_intersection_idx
+ON miovision_api.volumes_daily
```
chmnata, line 22: should we test out hash on this table ? Run a test with btree and hash and see if hash is faster 🤔

gabrielwol: I  just tested a hash index and there was no difference in performance on this small table. Perhaps we should experiment on a larger one.

volumes/miovision/sql/update-mio_volumes15min_mvt_partitioning.sql
```diff
@@ -0,0 +1,153 @@
...
+ALTER SEQUENCE IF EXISTS miovision_api.volumes_15min_mvt_volume_15min_mvt_uid_seq SET SCHEMA mio_staging;
+ALTER SEQUENCE IF EXISTS mio_staging.volumes_15min_mvt_volume_15min_mvt_uid_seq OWNED BY volumes_15min_mvt.volume_15min_mvt_uid;
+ALTER SEQUENCE IF EXISTS mio_staging.volumes_15min_mvt_volume_15min_mvt_uid_seq OWNER TO miovision_admins;
+
+/*Careful with these!!!
+
+--save and drop dependencies
+SELECT public.deps_save_and_drop_dependencies('miovision_api','volumes_15min_mvt');
+--truncate and drop old volumes_15min_mvt
+TRUNCATE miovision_api.volumes_15min_mvt;
```
chmnata, line 122:
```
ERROR:  cannot truncate a table referenced in a foreign key constraint
DETAIL:  Table "volumes_mvt_atr_xover" references "volumes_15min_mvt".
```

chmnata: have to run
```ALTER TABLE miovision_api.volumes_mvt_atr_xover DROP CONSTRAINT atr_mvt_uid_volume_15min_mvt_uid_fkey;```

File: volumes/miovision/sql/create-table-volumes.sql [hunk 18]
```diff
@@ -1,129 +1,60 @@
...
+    intersection_uid integer,
+    datetime_bin timestamp without time zone,
+    classification_uid integer,
+    leg text,
+    movement_uid integer,
+    volume integer,
+    volume_15min_mvt_uid integer,
+    CONSTRAINT volumes_intersection_uid_datetime_bin_classification_pkey 
+    PRIMARY KEY (intersection_uid, datetime_bin, classification_uid, leg, movement_uid),
+    CONSTRAINT volumes_volumes_15min_mvt_uid_fkey FOREIGN KEY (volume_15min_mvt_uid)
```
chmnata, line 15: `volumes_volume_15min_mvt_uid_fkey`

volumes/miovision/sql/update-mio_volumes15min_mvt_partitioning.sql
chmnata: and this one
```
ERROR:  cannot truncate a table referenced in a foreign key constraint
DETAIL:  Table "volumes" references "volumes_15min_mvt".
```
ran 
ALTER TABLE miovision_api.volumes DROP CONSTRAINT volumes_volume_15min_mvt_uid_fkey;

volumes/miovision/sql/update-mio_volumes15min_mvt_partitioning.sql
```diff
@@ -0,0 +1,153 @@
...
+DROP miovision_api.volumes_15min_mvt;
```
chmnata, line 123: `DROP TABLE miovision_api.volumes_15min_mvt;`

volumes/miovision/sql/create-table-volumes_15min_mvt.sql
```diff
@@ -1,67 +1,61 @@
...
-  CONSTRAINT volumes_15min_mvt_intersection_uid_datetime_bin_classificat_key UNIQUE (intersection_uid, datetime_bin, classification_uid, leg, movement_uid)
+    volume_15min_mvt_uid serial NOT NULL,
+    intersection_uid integer,
+    datetime_bin timestamp without time zone,
+    classification_uid integer,
+    leg text,
+    movement_uid integer,
+    volume numeric,
+    processed boolean,
+    CONSTRAINT volumes_15min_mvt_int_uid_dt_bin_class_leg_mvmt_uid_pkey
```
chmnata, line 11:
```
ERROR:  there is no unique constraint matching given keys for referenced table "volumes_15min_mvt"
```
when creating it again after dropping when transferring from mio_staging to miovision_api. We will just need to make volumes_15_min_mvt_uid unique

volumes/miovision/sql/update-mio_volumes15min_mvt_partitioning.sql
gabrielwol: > have to run `ALTER TABLE miovision_api.volumes_mvt_atr_xover DROP CONSTRAINT atr_mvt_uid_volume_15min_mvt_uid_fkey;`

For this one we need to re-add the constraint after moving over the staging table: 

```
ALTER TABLE miovision_api.volumes_mvt_atr_xover ADD CONSTRAINT atr_mvt_uid_volume_15min_mvt_uid_fkey FOREIGN KEY (volume_15min_mvt_uid)
        REFERENCES miovision_api.volumes_15min_mvt (volume_15min_mvt_uid) MATCH SIMPLE
        ON UPDATE RESTRICT
        ON DELETE CASCADE;
```

gabrielwol:
@chmnata 
I just updated volumes_daily table to use sum volumes_15min_mvt instead of volumes_15min following all my docs review from last week...

I need your help to:

```sql
TRUNCATE TABLE miovision_api.volumes_daily;
--Update function definition from: bdit_data-sources/volumes/miovision/sql/function-aggregate-volumes_daily.sql
SELECT miovision_api.aggregate_volumes_daily('2019-01-01'::date, '2023-11-13'::date);
```

chmnata:
I should update the aggregate_volumes_daily to the latest version as well right? 

> @chmnata I just updated volumes_daily table to use sum volumes_15min_mvt instead of volumes_15min following all my docs review from last week...
> 
> I need your help to:
> 
> ```sql
> TRUNCATE TABLE miovision_api.volumes_daily;
> --Update function definition from: bdit_data-sources/volumes/miovision/sql/function-aggregate-volumes_daily.sql
> SELECT miovision_api.aggregate_volumes_daily('2019-01-01'::date, '2023-11-13'::date);
> ```

chmnata:
`miovision_api.volumes_15min_mvt` successfully moved over last week, I was moving `miovision_api.volumes_15min` today with the following  code, but when select from public.deps_saved_ddl to check what was inserted, there was only one covid view that got inserted, which is unexpected but we ran before with dry run and it returned 50 tables.  
```
SELECT public.deps_save_and_drop_dependencies('miovision_api'::text COLLATE pg_catalog."C",'volumes_15min'::text COLLATE pg_catalog."C");

SELECT * FROM public.deps_saved_ddl 
where deps_view_schema = 'miovision_api'

ALTER TABLE mio_staging.volumes_15min SET SCHEMA miovision_api;
SELECT public.deps_restore_dependencies('miovision_api'::text COLLATE pg_catalog."C",'volumes_15min'::text COLLATE pg_catalog."C");
```

Either way, managed to find a list of dependencies from Gabe and my DM and they are the following:
Noting that there are less than 50 because before this operation a couple of users dropped their dependencies. The ones that got commented out are the ones that was successfully recreated using code from github, mainly all activeto and rapidto views + matviews. The data_analysis ones looks old, and I doubt we will still need the covid ones.... though I will still recreate them. 

Unsure about the reason why the following dependencies were not capture in using the function, but we will investigate more. Minimal impact cause the important ones was well documented. 
```
--"activeto"	"miovision_reporting_summary"	"v"
--"activeto"	"miovision_volumes_15min_adj"	"m"
--"activeto"	"miovision_comparison_summary"	"v"
--"activeto"	"miovision_comparison_hourly"	"v"


--"rapidto"	"miovision_segments_comparison_summary"	"v"
--"rapidto"	"miovision_segments_reporting_summary"	"v"
--"rapidto"	"miovision_outage_dates"	"v"
--"rapidto"	"miovision_segments_reporting_info"	"v"
--"rapidto"	"miovision_segments_comparison_daily"	"v"
--"rapidto"	"miovision_segments_comparison_hourly"	"v"


"data_analysis"	"miovision_daily_volumes_covid"	"v"
"data_analysis"	"miovision_veh_data"	"m"
"data_analysis"	"wys_miov_comparison_15min_withspd"	"v"

"covid"	"miovision_index_bl_tps"	"v"
"covid"	"miovision_index_14day"	"v"
"covid"	"miovision_summary_stats"	"v"
"covid"	"miovision_peak_all_intersections_dow"	"v"
"covid"	"miovision_summary"	"m"
"covid"	"miovision_all_intersections"	"v"
"covid"	"miovision_peak_all_intersections"	"v"
"covid"	"miovision_daily_summary_temp"	"m"
"covid"	"miovision_hourly_veh_movements"	"m"
"covid"	"miovision_compare"	"m"
"covid"	"miovision_baselines"	"v"
"covid"	"miovision_peak_summary"	"m"
"covid"	"miovision_daily_summary"	"m"
"covid"	"miovision_daily_summary_new"	"m"
"covid"	"miovision_hourly"	"v"
"covid"	"miovision_hourly_temp"	"v"
"covid"	"miovision_jarvis_screenline"	"v"
```

gabrielwol:
`mio_staging` dropped!

dags/pull_miovision.py
```diff
@@ -33,11 +32,62 @@
...
+            params={"year": YEAR,
+                    "month": MONTH},
+            autocommit=True
+        )
+
+        check_annual_partition() >> create_annual_partition
+        check_month_partition() >> create_month_partition
+
+    t1 = BashOperator(
         task_id = 'pull_miovision',
```
chmnata, line 85: Can you update the directory to 
`/data/airflow/airflow_venv/bin/python3 /data/airflow/data_scripts/volumes/miovision/api/intersection_tmc.py run-api --path /data/airflow/data_scripts/volumes/miovision/api/config.cfg`

volumes/miovision/sql/create-table-volumes_15min.sql
```diff
@@ -3,66 +3,70 @@
...
-  CONSTRAINT volumes_15min_pkey PRIMARY KEY (volume_15min_uid),
-  CONSTRAINT volumes_15min_intersection_uid_datetime_bin_classification__key UNIQUE (intersection_uid, datetime_bin, classification_uid, leg, dir)
+    volume_15min_uid serial NOT NULL,
+    intersection_uid integer,
+    datetime_bin timestamp without time zone,
+    classification_uid integer,
+    leg text,
+    dir text,
+    volume numeric,
+    CONSTRAINT volumes_15min_intersection_uid_datetime_bin_classification_pkey
```
chmnata, line 13: Do you think we should make volume_15min_uid the primary key, and make this chunk a unique constraint instead?

chmnata: to make foreign references we need a the column (volumes_15min_uid or volumes_15min_mvt) to have a unique constraint/ as the primary key. I think our original table structure has volumes_15min_uid as the primary key and a unique constraint on the list of columns.

gabrielwol: "To create a unique or primary key constraint on a partitioned table, the partition keys must not include any expressions or function calls and the **constraint's columns must include all of the partition key columns**." [Source](https://www.postgresql.org/docs/current/ddl-partitioning.html#DDL-PARTITIONING-DECLARATIVE-LIMITATIONS)
The other option would be to add the uid to the current Pkey.

chmnata: Ok lets add it to the primary key then. I was initially thinking we can make it a unique constraint since we would not want duplicate uids, as serial column could produce duplicate uid if we manually specify a value in the serial column (thought now that i think about it, we probably will never do this)

gabrielwol: I just added these lines: 
```sql
--add volume_15min_mvt_uid to pkey so it can be used in fkey
ALTER TABLE miovision_api.volumes_15min_mvt DROP CONSTRAINT volumes_15min_mvt_int_uid_dt_bin_class_leg_mvmt_uid_pkey;
ALTER TABLE miovision_api.volumes_15min_mvt ADD CONSTRAINT volumes_15min_mvt_int_uid_dt_bin_class_leg_mvmt_uid_pkey
PRIMARY KEY (volume_15min_mvt_uid, intersection_uid, datetime_bin, classification_uid, leg, movement_uid);
```

chmnata: thanks! running

chmnata: ok so im still running into the same error 
`ERROR:  there is no unique constraint matching given keys for referenced table "volumes_15min_mvt"`
I believe its because we are referencing `volume_15min_mvt_uid` and not `volumes_15min_mvt_int_uid_dt_bin_class_leg_mvmt_uid_pkey`, where the first one technically on itself can be not unique, it just needs to be unique with combination of another columns. Can we make that one column a unique constraint + cascade to the rest of the partition?

gabrielwol: Let's try again! I think the declarative partitioning takes care of cascading to the partitions. Looks to be the case for the change we made above in this thread.
```sql
ALTER TABLE miovision_api.volumes_15min_mvt ADD CONSTRAINT volume_15min_mvt_uid_unique UNIQUE (volume_15min_mvt_uid);
```

chmnata: sigh
```
ERROR:  unique constraint on partitioned table must include all partitioning columns
DETAIL:  UNIQUE constraint on table "volumes_15min_mvt" lacks column "datetime_bin" which is part of the partition key.
```

chmnata: seems like we cannot have a foreign key referencing the column `volumes_15min_mvt` alone in this case, as it has to include partition key. Unless we reference datetime_bin + volumes_15min_mvt, however this wouldn't work cause  datetime_bin in volumes_15min_mvt wont be the same as the datetime_bin in `volumes`.  
I think not having foreign key is not too big of a deal, and saving a little space from not creating unique constraint+ index on this table is not a bad thing. 
Maybe in the future we can revisit the use of volumes_15min_mvt_uid(?) 
I'm leaning towards dropping the foreign key issue and merge this PR. Thoughts?

gabrielwol: I agree removing the foreign key relationship on `volumes` referencing `volumes_15min_mvt` is fine since it's incompatible with the new partition design. The new replacement index will help with this condition: 
https://github.com/CityofToronto/bdit_data-sources/blob/99a0e2c1861a76b0650111b2d3e925be5950e5d7/volumes/miovision/sql/function-aggregate-volumes_15min_mvt.sql#L62C1-L63C43

chmnata (approved):

volumes/miovision/sql/create-table-volumes.sql
```diff
@@ -1,129 +1,60 @@
...
+    intersection_uid integer,
+    datetime_bin timestamp without time zone,
+    classification_uid integer,
+    leg text,
+    movement_uid integer,
+    volume integer,
+    volume_15min_mvt_uid integer,
+    CONSTRAINT volumes_intersection_uid_datetime_bin_classification_pkey 
+    PRIMARY KEY (intersection_uid, datetime_bin, classification_uid, leg, movement_uid),
+    CONSTRAINT volumes_volume_15min_mvt_uid_fkey FOREIGN KEY (volume_15min_mvt_uid)
```
chmnata, line 15: we will have to get of this foreign key then :cry_cat:

gabrielwol: done!

PR #678, opened 2023-08-21
I0673 miovision find_gaps fixes

gabrielwol:
## What this pull request accomplishes:
- Completely rewrote miovision `find_gaps` function to fix many issues described in #673 including:
- Speed (<1 min from 40 mins), readability improvements
- Catch gaps for detectors that ~~reported no data or~~ had an outage starting midday by performing a cross join of intersections and start/end of day
- ~~Removed second date parameter, since based on a 60 day lookback it should be running one date at a time. Made necessary changes to run_api to support this change.~~ Used window function to correctly reference a rolling lookback when run for multiple days rather than use a single fixed lookback. See: https://github.com/CityofToronto/bdit_data-sources/pull/678#issuecomment-1858066978
- Considered `ref.holidays` when determining weekday/weekend.
- Fixed a minor bug where gap sizes are exaggerated by 1 minute since the gap_start was incorrectly treated as inclusive. 
- ~~Fixed a bug where gaps sometimes ended on the hour, causing the whole following hour to be excluded.~~ Cross join gaps to the 15 minute periods they overlap with, which fixes some incorrect edge cases in gap expansion to hour periods. 
  - Adjust the join when aggregating into `volumes_15min_mvt` to reflect this change. 
- ~~Changed gaps to more explicitly reference the beginning and end of the hour instead of the gap start and end since, the gap list was not complete anyway (due to DISTINCT ON) and could cause confusion if used for other purposes.~~
- Updated readme to reflect changes to `find_gaps`.

**Latest update (2023-12-13):**
- Separate `gapsize_lookup` (previously a CTE) into a table so that it can also be used downstream by new `daily_volumes` table (#780).
- In `gapsize_lookup`, include both total volumes to be used for gap threshold determination, and individual classification volumes to be used for imputing the missing volume from gaps. 

## Issue(s) this solves:
- Closes #673 
- Closes #419 

## What, in particular, needs to reviewed:
- volumes/miovision/sql/function-find_gaps.sql: completely rewrote find_gaps
- volumes/miovision/api/intersection_tmc.py: Changes to reflect removal of end_date from find_gaps. 

## What needs to be done by a sysadmin after this PR is merged
See https://github.com/CityofToronto/bdit_data-sources/pull/678#issuecomment-1892761571 for implementation steps.

Issue #419, opened 2021-06-28
Miovision Gap Finding SQL function doesn't correctly identify entire missing days 


radumas:
Not only does it not alert us to these, but it does also fill in 0s for them instead of `NULL`

radumas:
another one!
```sql
SELECT intersection_name , DATE(datetime_bin) dt, COUNT(DISTINCT datetime_bin), classification, SUM(volume)
	FROM miovision_api.volumes_15min
	NATURAL JOIN miovision_api.intersections
	NATURAL JOIN miovision_api.classifications
	WHERE Datetime_bin >= '2021-11-30' AND datetime_bin < '2021-12-03'
	GROUP BY intersection_name, dt, classification
	ORDER BY dt, intersection_name,classification
```

radumas:
Related
https://github.com/CityofToronto/bdit_data-sources/issues/370

gabrielwol:
Related to this, find_gaps doesn't identify gaps that start before midnight or end after midnight next day.
For example:
```sql
--this sensor stopped working at 2023-04-24 04:55:00. 
SELECT MAX(datetime_bin) 
FROM miovision_api.volumes
WHERE
    intersection_uid = 45
    AND datetime_bin >= '2023-04-24 00:00:00'
    AND datetime_bin < '2023-04-25 00:00:00'
```
```sql
--However last gap listed for sensor is "2023-04-24 04:19:00" to "2023-04-24 04:46:00"
SELECT *
FROM miovision_api.unacceptable_gaps
WHERE
    intersection_uid = 45
ORDER BY gap_end DESC
LIMIT 1

```

gabrielwol:
Planning to solve this by adding time bins corresponding to start_date and end_date for each intersectoin, if they don't already exist.

PR #738, opened 2023-11-03
737 pull miovision dag data pull checks

### LINKED ISSUE #370: determine_working_machine improperly handles mass data outages
On 2020-12-14 there was a partial outage of Miovision data. From 18:30-18:45 some stations went offline, and from 18:45-19:00 all stations were offline. This can be seen using

```sql
SELECT datetime_bin,
	   SUM(volume) total_vol
FROM miovision_api.volumes
WHERE datetime_bin BETWEEN '2020-12-14 18:30:00' AND '2020-12-14 19:15:00'
GROUP BY datetime_bin
ORDER BY 1
```

This was confusingly reported by the `check_miovision` Airflow process:

image

The contradictory notices come from [`determine_working_machine`](https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/miovision/sql/function-determine_working_machine.sql) - the first is [from `tot`](https://github.com/CityofToronto/bdit_data-sources/blob/520498173da5e24ca7546195699aff4014f77051/volumes/miovision/sql/function-determine_working_machine.sql#L65), which checks if any stations have invalid gaps, while the second is from the [contents of `mis`](https://github.com/CityofToronto/bdit_data-sources/blob/520498173da5e24ca7546195699aff4014f77051/volumes/miovision/sql/function-determine_working_machine.sql#L16), which includes invalid gaps for any stations. When every station is down, however, `determine_working_machine` associates the corresponding timestamps with `NULL` intersection and volume UIDs, and inserts a row with `intersection_uid = NULL` into `mis`.

Ideally we would figure out a way to distinguish lines with `intersection_uid = NULL` after they're generated in [this CTE](https://github.com/CityofToronto/bdit_data-sources/blob/520498173da5e24ca7546195699aff4014f77051/volumes/miovision/sql/function-determine_working_machine.sql#L25-34). One simple thing would be to simply replace `NULL` with `-1`, `999` or `'ALL'` (if we're willing to return a string rather than an int in `mis`). We'd also need to figure out the error message handling for `tot`, since it shouldn't report `All cameras are working fine` in those cases.

I have no idea why the script didn't report the partial outages from 18:30 - 18:45.

### LINKED ISSUE #737: pull_miovision DAG missing basic data pull checks
- row count
- distinct intersection_uid count
- distinct classification_uid count

gabrielwol:
- unifies the data check sqls between miovision and wys and moves to a common location under `bdit_data-sources/dags`.
  - Note `wys/api/sql/select-row_count_lookback.sql` was moved to the common source under another PR that got merged first so it appears to be deleted here.
- cc @tahaislam Copied the latest SQLCheckOperatorWithReturnValue from VFH, edited to push xcom value if data check fails. 
- cc @tahaislam Edited the common `task_fail_slack_alert` to separate the `extra_msg` arg with new lines and include as an attachment instead of part of regular message. 
- Adds 4 data checks to pull_miovision
  - row count
  - distinct ids: classification_uid
  - distinct ids: intersection_uid
  - find gaps (now with a generic script)
- Put some data checks in pull_* DAGs and some in *check* DAGs depending on which are red/yellow card checks.
  - New skeleton `wys_check` DAG for 1 data check.
- Improvements to the `select-sensor_id_count_lookback.sql` sql to include a list of unique ids which occurred during look-back but not today to logs!! example wys response;   

| "check" | "ds_count"         | "lookback_avg"              | "passing_value"       | "id_diff"                                                                                                                                                                                                                                         |   |
|---------|--------------------|-----------------------------|-----------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---|
| true    | "Daily count: 719" | "60 Days Lookback Avg: 682" | "Pass threshold: 613" | "The following api_id were present within the last 60 days, but absent today:{4651,4661,4692,6867,6870,6873,6885,7702,7913,8067,8112,8175,8286,8332,8427,8428,8434,8726,23337,32606,58444,58922,68317,68318,68319,68320,68806,71148,71344,71541}" |   |

- New generic find gaps data check sql which can work for multiple volumes data-sources. Sample response for miovision:
<html><body>
<!--StartFragment-->

"check" | "summ" | "gaps"
-- | -- | --
"False" | "There was 1 gaps larger than 2 hours." | "{""(7,\""2023-11-07 22:00:00--2023-11-08 19:38:00\"")""}"

<!--EndFragment-->
</body>
</html>
  - Will work with any id column.


## Issue(s) this solves:
- Closes #737 
- Closes #370 (via `check_distinct_intersection_uid`, which identifies intersections with no values today)

- Thresholds
  - I've suggested 0.999 for miovision classification / intersection, which would cause us to get failure notifications every time there is 1 id missing. 

Test these two DAGs.


gabrielwol:
I only realized after creating this PR that there is an existing DAG `check_miovision` which has some overlap with the miovision intersection_uid check task added here. 
- The existing `miovision_api.determine_working_machine` function used in `check_miovision` finds intersections with 4 hour gaps. Intersections with no data are not highlighted. 
- The new task I proposed here finds intersections that appeared within the last X days (currently X=60) but not today. Does not identify intersections that appeared but with gaps.  

We could keep both tasks but I would propose we scrap that DAG and add that task to this DAG. @radumas, @chmnata thoughts?

radumas:
I'm on board with combining DAGs. 

I think the identifying non-functioning intersections is a critical function until the alert pipeline can supersede it.

radumas:
This should also solve https://github.com/CityofToronto/bdit_data-sources/issues/370 and https://github.com/CityofToronto/bdit_data-sources/issues/419

(maybe including them in a comment will make them actually show up in the "link an issue" sidebar)

gabrielwol:
Note: [This section](https://github.com/CityofToronto/bdit_data-sources/blob/158b5cb943da73598e59c8f45da33b5d4d397b0d/volumes/miovision/README.md#3-finding-gaps-and-malfunctioning-camera) of the readme should be updated as part of this PR

dags/sql/select-find_gaps.sql
```diff
@@ -0,0 +1,73 @@
...
+bin_times AS (
+    SELECT 
+        sensor_id_col,
+        dt_col AS gap_start,
+        LEAD(dt_col, 1) OVER (PARTITION BY sensor_id_col ORDER BY dt_col) AS gap_end,
+        LEAD(dt_col, 1) OVER (PARTITION BY sensor_id_col ORDER BY dt_col)
+            - dt_col
+            --sum works because between 0 (synthetic) and 1 (real), we want 1 (implies real data)
+            - SUM(gap_adjustment)
+        AS bin_gap
```
gabrielwol, line 51: I didn't agree with these sqlfluff recommendations so I ignored them 🗡️

gabrielwol:
@chmnata these changes are working as intended! see [todays dag run](https://trans-bdit.intra.prod-toronto.ca/airflow/dags/pull_miovision/grid?root=&dag_run_id=scheduled__2023-11-20T08%3A00%3A00%2B00%3A00&task_id=data_checks) and the example [slack messages](https://bditto.slack.com/archives/C04BWRWVB3L/p1700582498151649) which include the data check results

gabrielwol:
> This should also solve https://github.com/CityofToronto/bdit_data-sources/issues/370 and https://github.com/CityofToronto/bdit_data-sources/issues/419

#370, yes. 
#419 is more correctly tagged in [#678 - find_gaps improvements](https://github.com/CityofToronto/bdit_data-sources/pull/678)

gabrielwol:
Note to self: can make use `extra_msg_from_task` implemented [here](https://github.com/CityofToronto/bdit_data-sources/pull/787#pullrequestreview-1760586506) instead of `get_return_value`.

volumes/miovision/README.md
```diff
@@ -366,7 +366,8 @@ The following process is used to determine the gap sizes assigned to an intersec
...
+The following process is to determine if a Miovision camera is still working. It is different from the process above because the gap sizes used above are small and do not say much about whether a camera is still 
+working. We roughly define a camera to be malfunctioning if that camera/intersection has a gap greater than 4 hours OR do not have any data after '23:00:00'. This is checked daily by `pull_miovision` DAG, task `data_checks.check_gaps`. A slack notification will be sent if there's at least 1 camera that is not working. The function also returns a list of intersections that are not working and from what time to what time that the gaps happen which is helpful in figuring out what has happened.
```
chmnata, line 370: is the `do not have any data after 23:00:00` still true after we switch to a day aggregation?

dags/sql/select-find_gaps.sql
```diff
@@ -0,0 +1,77 @@
+--a generic find_gaps function to find gaps in sensor data
```
chmnata, line 1: I think this is a great base for a general find_gaps functions! I have some general thoughts:
1)  Can we make this a function? I think this will be really helpful created as a function in postgres, I can think of a couple of use cases, e.g. checking the number of 1 day gap for an individual bluetooth route with in a week, checking a set of link_dir's gaps within peak hours, etc. 
2)  Adding an end_datetime param instead of using ds + 1 day interval. I think this add flexibility to the usage and is a relatively small change
3)  Add a [array] param for selecting sensor_id. default can be all, but sometimes when debugging and investigating gaps for sensors we might be interested in one specific sensor.

I think for point 2 and 3, we could make it into another issue if we want to keep this query as simple as possible, open for discussion

gabrielwol: Thanks! I'll implement at least 1+2 in this PR, maybe 3. 
I'm also considering a param for returning the list of outages as opposed to the summary if we think this could have wider use.

dags/sql/select-find_gaps.sql
```diff
@@ -0,0 +1,77 @@
...
+            ('{{ ds }} 00:00:00'::timestamp + '1 day'::interval)
+        ) AS bins(dt_col)
+    
+    UNION ALL
+    
+    SELECT
+        sensor_id_col,
+        dt_col,
+        --need to reduce gap length by 1 minute for real data since that minute contains data
+        '{{ params.default_bin }}'::interval AS gap_adjustment
```
chmnata, line 37: is there a use case for this to not be 1 min?

gabrielwol: For different volumes datasets with different default reporting intervals! Admittedly I haven't tested on anything except for the 1-minute miovision data.

dags/sql/select-find_gaps.sql
gabrielwol: 🥳  You can test the new functions in my schema with:

```sql
SELECT * FROM gwolofs.generic_find_gaps(
    '2023-11-20'::date,
    '2023-11-20'::date,
    'intersection_uid'::text,
    'datetime_bin'::text,
    'miovision_api'::text,
    'volumes'::text,
    '10 minutes'::interval,
    '1 minute'::interval
)

SELECT * FROM gwolofs.summarize_gaps_data_check(
    '2023-11-20'::date,
    '2023-11-20'::date,
    'intersection_uid'::text,
    'datetime_bin'::text,
    'miovision_api'::text,
    'volumes'::text,
    '10 minutes'::interval,
    '1 minute'::interval
)
```

A dbadmin will have to create the functions in public schema.

volumes/miovision/README.md
gabrielwol: Editing.

dags/sql/select-find_gaps.sql
chmnata: splitting them into two function is so helpful! generic_find_gaps is definitely gonna be useful during investigation

chmnata: just tested it with bluetooth, returns an error with data type cause analysis_id is a big int and it was expecting the column to be an integer. We can probs just cast the column to integer explicitly 

```
SELECT * FROM gwolofs.generic_find_gaps(
    '2023-11-20'::date,
    '2023-11-20'::date,
    'analysis_id'::text,
    'datetime_bin'::text,
    'bluetooth'::text,
    'aggr_5min'::text,
    '10 minutes'::interval,
    '1 minute'::interval
)

```

chmnata: actuallllllllllllllly mebbe a text will be better, using it on `link_dir` doesnt work cause its a string

chmnata: I wonder if its possible to do dynamic column type return

chmnata: seems possible! 
https://www.postgresql.org/docs/current/datatype-pseudo.html

gabrielwol: good call! 
works now with an extra parameter for datatype (null::datatype) 
'''sql
SELECT * FROM gwolofs.generic_find_gaps(
    '2023-11-20'::date,
    '2023-11-20'::date,
    'analysis_id'::text,
    'datetime_bin'::text,
    'bluetooth'::text,
    'aggr_5min'::text,
    '20 minutes'::interval,
    '1 minute'::interval,
    null::bigint
)
'''

chmnata:
Moving the convo from platonic pipeline to here. We can try to distinguish red card and yellow card checks? Anything below that if failed we will want to run the dag again? @radumas @gabrielwol 
- check_row_count
   - Compare the row count today with the average row count from the lookback period.
- check_distinct_classification_uid 
   - Compare the count of classification_uids appearing in today's pull vs the lookback period
- check_distinct_intersection_uid 
   - Identify intersections which appeared within the lookback period that did not appear today.
- check_gaps 
   - Identify gaps larger than gap_threshold in intersections with values today.

gabrielwol:
Maybe:
pull_miovision:
- check_row_count
- check_distinct_classification_uid
check_miovision:
- check_distinct_intersection_uid
- check_gaps

radumas:
I support Gabe's proposal

wys/api/sql/select-row_count_lookback.sql
gabrielwol, line 1: This is now located in `dags/sql`. It was created in [another PR](https://github.com/CityofToronto/bdit_data-sources/pull/732/files#diff-dfa910c9ee4f288a2c50ee1c9ba45f1106c5ca5f2635d1da092ce018e1613b64).

dags/sql/select-find_gaps.sql
chmnata: ahhhh make sense

dags/sql/create-function-generic_find_gaps.sql
```diff
@@ -0,0 +1,109 @@
...
+CREATE OR REPLACE FUNCTION public.generic_find_gaps(
+    start_date date,
+    end_date date,
+    id_col text,
+    dt_col text,
+    sch_name text,
+    tbl_name text,
+    gap_threshold interval,
+    default_bin interval,
+    id_col_dtype anyelement DEFAULT NULL::int
```
chmnata, line 11: kinda frustrating that it needs to know the datatype at call and we have to add another input to specify the datatype.... I tried some other things but I think this is the only way at the moment.

dags/sql/create-function-generic_find_gaps.sql
```diff
@@ -0,0 +1,109 @@
...
+VOLATILE
+
+AS $BODY$
+DECLARE
+   resulttype regtype := pg_typeof(id_col_dtype);
+BEGIN
+    RETURN QUERY EXECUTE FORMAT($$
+        WITH raw AS (
+            SELECT DISTINCT
+                CAST(%I AS %s) AS sensor_id_col, --id_col, resulttype
```
chmnata, line 27: I feel like we dont need to re-cast the column, as it should already has that column type. is this here so we can declare something?

dags/sql/create-function-generic_find_gaps.sql
```diff
@@ -0,0 +1,109 @@
...
+COST 100
+VOLATILE
+
+AS $BODY$
+DECLARE
+   resulttype regtype := pg_typeof(id_col_dtype);
+BEGIN
+    RETURN QUERY EXECUTE FORMAT($$
+        WITH raw AS (
+            SELECT DISTINCT
```
chmnata, line 26: I guess one "caveat" for this type of query is if the sensor has a gap larger than [start_date, end_date), then it won't show up at all eh?

dags/sql/create-function-generic_find_gaps.sql
```diff
@@ -0,0 +1,109 @@
...
+OWNER TO gwolofs;
+
+GRANT EXECUTE ON FUNCTION public.generic_find_gaps TO bdit_humans;
+GRANT EXECUTE ON FUNCTION public.generic_find_gaps TO bdit_bots;
+
+COMMENT ON FUNCTION public.generic_find_gaps
+IS '''A generic find_gaps function to find gaps in sensor data.
+Includes a lookback of `gap_threshold` so it catches gaps overlapping the start of the interval.
+Includes synthetic start/end time points for all sensors so we catch gaps that don''t start or
+don''t end in the interval. Doesn''t catch sensors with no data in the interval.''';
```
chmnata, line 109: ah its already in the comments :P

dags/sql/create-function-generic_find_gaps.sql
```diff
@@ -0,0 +1,109 @@
...
+END;
+$BODY$;
+
+ALTER FUNCTION public.generic_find_gaps
+OWNER TO gwolofs;
+
+GRANT EXECUTE ON FUNCTION public.generic_find_gaps TO bdit_humans;
+GRANT EXECUTE ON FUNCTION public.generic_find_gaps TO bdit_bots;
+
+COMMENT ON FUNCTION public.generic_find_gaps
```
chmnata, line 105: can we add a comment to mention whether `gap_end` and `bin_gap` is inclusive or naut

dags/sql/create-function-summarize_gaps_data_check.sql
```diff
@@ -0,0 +1,60 @@
...
+                %L::text, --tbl_name
+                %L::interval, --gap_threshold
+                %L::interval, --default_bin
+                %L::regtype --id_col_dtype
+            )
+        )
+
+        SELECT
+            NOT(COUNT(summarized_gaps.*) > 0) AS check,
+            CASE WHEN COUNT(summarized_gaps.*) = 1 THEN 'There was ' ELSE 'There were ' END ||
```
chmnata, line 40: grammar is important!

dags/sql/create-function-generic_find_gaps.sql
radumas: could we do function overloading?

gabrielwol: I think `id_col_dtype` is more elegant than function overloading!

dags/check_miovision.py
```diff
@@ -42,18 +47,24 @@ def check_miovision(con, start_date, end_date):
...
                  'retry_delay': timedelta(minutes=5),
                  'on_failure_callback': task_fail_slack_alert
                 }
 
-dag = DAG(dag_id = dag_name, default_args=default_args, schedule='0 7 * * *', catchup=False)
+dag = DAG(dag_id = dag_name,
+          default_args=default_args,
+          schedule='0 7 * * *',
+          catchup=False,
+          tags = ['miovision', "data_checks", "archived"],
```
chmnata, line 65: no longer archived!

dags/check_miovision.py
```diff
@@ -1,3 +1,8 @@
+"""
```
chmnata, line 1: can we change this dag's name to be `miovision_check`?

chmnata: WAIT there is another miovision_check!??!

gabrielwol: I figured since the new `miovision_check` has none of the same tasks as existing `check_miovision`, it would be best to duplicate and rename rather than update existing.

volumes/miovision/api/readme.md
```diff
@@ -237,9 +241,15 @@ Since there is this function on Airflow script to send slack notifications when
 
 The Airflow uses BashOperator and run one task named `pull_miovision` using a bash command that looks something like this bash_command = `'/etc/airflow/.../intersection_tmc.py run-api --path /etc/airflow/.../config.cfg --dupes'`. `--dupes` is used to catch duplicates and fail the script when that happen.
```
chmnata, line 242: i know its just an example but the bash command's has been updated with a new path file, if you can update it here as well it would be greeeat

volumes/miovision/api/readme.md
```diff
@@ -237,9 +241,15 @@ Since there is this function on Airflow script to send slack notifications when
...
+- `check_distinct_intersection_uid` checks the count of distinct values in `intersection_uid` column.  
+- `check_gaps` identifes gaps larger than 4 hours for active miovision cameras.  
+
 ### **`check_miovision`**
 
-There is another Airflow process related to Miovision named `check_miovision` which is to check if all Miovision cameras are working and send notifications when there is at least one malfunctioning camera. More information can be found at [this part of the README.](https://github.com/CityofToronto/bdit_data-sources/tree/miovision_api_bugfix/volumes/miovision#3-finding-gaps-and-malfunctioning-camera)
+The `check_miovision` DAG is deprecated by the addition of the `data_checks` TaskGroup to the main `pull_miovision` DAG, in particular `data_checks.check_gaps` which directly replaces `check_miovision.check_miovision`. This DAG previously was used to check if any Miovision camera had a gap of at least 4 hours. More information can be found at [this part of the README.](https://github.com/CityofToronto/bdit_data-sources/tree/miovision_api_bugfix/volumes/miovision#3-finding-gaps-and-malfunctioning-camera)
```
chmnata, line 252: We should also mention miovision_check here as well

dags/sql/create-function-generic_find_gaps.sql
gabrielwol: I don't fully understand this, but when I remove this CAST, `generic_find_gaps` still works, but `summarize_gaps_data_check` gives this error: 

> ERROR:  Returned type integer does not match expected type regtype in column 1.structure of query does not match function result type 
> 
> ERROR:  structure of query does not match function result type
> SQL state: 42804
> Detail: Returned type integer does not match expected type regtype in column 1.
>

dags/miovision_check.py
```diff
@@ -0,0 +1,100 @@
...
+    catchup=False,
+    tags=["miovision", "data_checks"],
+    doc_md=__doc__
+)
+def miovision_check_dag():
+
+    t_upstream_done = ExternalTaskSensor(
+        task_id="starting_point",
+        external_dag_id="pull_miovision",
+        external_task_id="done",
```
chmnata, line 55: I think this is supposed to reference the task_id aka `pull_miovision` once more. DAG name and task name should really be distinct, I'm glad we are changing the dag name in another dag :meow_dio:

chmnata: nvm I saw your externaltaskmarker :meow_dio:

dags/sql/select-sensor_id_count_lookback.sql
```diff
@@ -16,11 +17,24 @@ WITH lookback AS ( --noqa: L045
...
     FROM {{ params.table }}
     WHERE
         {{ params.dt_col }} >= '{{ ds }} 00:00:00'::timestamp
         AND {{ params.dt_col }} < '{{ ds }} 00:00:00'::timestamp + interval '1 day'
+),
+
+ids_dif AS (
+    SELECT ARRAY_AGG(ids_diff) AS ids_diff
+    FROM (
+        SELECT DISTINCT UNNEST(lb.daily_ids) AS ids_diff
```
chmnata, line 31: if we are comparing the distinct ids across the entire period here, I think we can omit creating an array for all days within the date range in the lookback cte

chmnata: nvm we use this calculate the lookback avg 🤔

dags/sql/select-sensor_id_count_lookback.sql
```diff
@@ -31,7 +45,13 @@ SELECT
     'Pass threshold: ' || to_char(
             FLOOR({{ params.threshold }}::numeric * AVG(lb.lookback_count)),
```
chmnata, line 46: so technically the passing value could change overtime ?

dags/sql/select-sensor_id_count_lookback.sql
```diff
@@ -31,7 +45,13 @@ SELECT
...
             'FM9,999,999,999'
-            ) AS passing_value
+            ) AS passing_value,
```
chmnata, line 48: the to_char function does a round up to 0 sig fig, i think we should include at least 1 decimal

dags/miovision_check.py
```diff
@@ -0,0 +1,100 @@
...
+LOGGER = logging.getLogger(__name__)
+logging.basicConfig(level=logging.DEBUG)
+
+DAG_NAME = 'miovision_check'
+DAG_OWNERS = Variable.get('dag_owners', deserialize_json=True).get(DAG_NAME, ["Unknown"])
+
+default_args = {
+    'owner': ','.join(DAG_OWNERS),
+    'depends_on_past':False,
+    'start_date': pendulum.datetime(2023, 12, 21, tz="America/Toronto"),
```
chmnata, line 34: can you change the date to like Dec 19 so we can test it?

dags/pull_miovision.py
```diff
@@ -11,39 +13,45 @@
...
-@dag(dag_id=dag_name,
-     default_args=default_args,
-     schedule='0 3 * * *',
-     catchup=False)
+@dag(
+    dag_id=DAG_NAME,
+    default_args=default_args,
+    # Add 3 hours to ensure that the data are at least 2 hours old
+    schedule='0 3 * * *',
+    template_searchpath=os.path.join(repo_path,'dags/sql'),
```
chmnata, line 46: need to add this to miovision_check

gabrielwol: Done!

radumas:
Is the row count just looking at the last 60 days? Has there been some analysis of the seasonal variation in counts? In the absence of data I'm not sure if 0.7 is low enough. 

Otherwise things look good. 

> I feel like for cases like classification_uid, the threshold should be a set value, maybe we can add another param for that? Also feel like we should have that classification_uid check per intersection, but that should use a passing value of threshold * avg(past 60 days value)

I think ultimately for multi-modal cameras we're going to want to be looking at volumes at an approach-classification volumes to identify anomalies.

gabrielwol:
> Is the row count just looking at the last 60 days? Has there been some analysis of the seasonal variation in counts? In the absence of data I'm not sure if 0.7 is low enough.

Yes. No. Maybe? I can add weekday/weekend+holiday grouping as a bare minimum. 

> I think ultimately for multi-modal cameras we're going to want to be looking at volumes at an approach-classification volumes to identify anomalies.

Yes, for future data quality checks, but too detailed for a data pipeline check.

gabrielwol:
A short visual analysis: 

WYS row counts with 70% threshold: (oragne=alert, blue = no alert). Identified 57 days last year. 46 if using 60% threshold. 
Only 1 extra bad day was identified when I tried partitioning by weekday/weekend.
image

Miovision: (Query was much slower, only did 100 days). Identified only 1 bad day (Thanksgiving) whether or not partitioning by weekday/weekend. 60% = 0 days. 
image

SQL (+ graphing in Excel):
```sql
WITH lookback AS ( --noqa: L045
    SELECT
        date_trunc('day', raw.datetime_15min) AS _dt, --noqa: L039
        NOT(date_part('isodow', raw.datetime_15min) <= 5 AND hol.holiday IS NULL) AS weekend,
        SUM(raw.count) AS lookback_count
    FROM vds.veh_speeds_15min AS raw
    LEFT JOIN ref.holiday AS hol ON
        raw.datetime_15min >= hol.dt
        AND raw.datetime_15min < hol.dt + interval '1 day'
    WHERE
        raw.datetime_15min >= '2023-12-16 00:00:00'::timestamp - interval '365 days'
        AND raw.datetime_15min < '2023-12-16 00:00:00'::timestamp
    --group by day then avg excludes missing days.
    GROUP BY
        _dt, --noqa: L003
        hol.holiday,
        date_part('isodow', raw.datetime_15min)
    ORDER BY _dt
)

SELECT
    _dt,
    lookback_count,
    weekend,
    AVG(lookback_count) OVER w
FROM lookback
WINDOW w AS (
    --60 day lookback for same intersection, classification, day type, hour
    PARTITION BY weekend
    ORDER BY _dt RANGE BETWEEN
        interval '60 days' PRECEDING AND interval '1 days' PRECEDING
)
ORDER BY _dt
```

dags/sql/select-sensor_id_count_lookback.sql
gabrielwol: Yeah! I think it makes sense, if we add/remove detectors. As long as there's no sudden downwards spike.

dags/sql/select-sensor_id_count_lookback.sql
gabrielwol: I floored all the thresholds because the `today_count` will always be an integer (Row count / count distinct ids) so I thought it makes more sense to compare int/int.

chmnata: make sense :meow_salute:

dags/pull_wys.py
```diff
@@ -117,29 +119,25 @@ def pull_wys(ds=None):
...
-            "dt_col": 'datetime_bin',
-            "threshold": 0.7
-        }
         check_row_count = SQLCheckOperatorWithReturnValue(
             task_id="check_row_count",
             sql="select-row_count_lookback.sql",
             conn_id="wys_bot",
-            params=data_check_params | {"col_to_sum": 'volume'},
-            retries=2
+            params={
```
chmnata, line 126: Can we print the error msg to slack if fails as well?

radumas:
Using `ref.holidays` would be a good inclusion to the checking... maybe buffer the threshold extra

gabrielwol:
I updated the holiday threshold to be 50% of weekend day avg lookback row count.
Also got the [attachment](https://bditto.slack.com/archives/C04BWRWVB3L/p1704237996377569) working again!

dags/pull_wys.py
gabrielwol: Check out this change:
https://github.com/CityofToronto/bdit_data-sources/blob/6ac6b1344ea62b9956eeac6f2ae3d3ab1cf534a1/dags/custom_operators.py#L50-L59
Plus sample messages: https://bditto.slack.com/archives/C04BWRWVB3L/p1704297198015299

chmnata:
Currently when the row count check in miovision pull fails, all other data checks in miovision checks does not get triggered to run. I feel like we should still run the check gaps + check distinct intersection uid even if there were less row counts. What do you think?

gabrielwol:
> Currently when the row count check in miovision pull fails, all other data checks in miovision checks does not get triggered to run. I feel like we should still run the check gaps + check distinct intersection uid even if there were less row counts. What do you think?

I agree! I actually made that change last week here: https://github.com/CityofToronto/bdit_data-sources/pull/738/files#diff-1b2bc68a1c008f3fa5cef31de30d2b5eeb5541ca252a8b6da352213c5258bac0R126

dags/pull_miovision.py
```diff
@@ -70,7 +78,51 @@ def check_partitions():
...
+            "table": "miovision_api.volumes_15min_mvt",
+            "lookback": '60 days',
+            "dt_col": 'datetime_bin',
+            "threshold": 0.7
+        }
+        check_row_count = SQLCheckOperatorWithReturnValue(
+            task_id="check_row_count",
+            sql="select-row_count_lookback.sql",
+            conn_id="miovision_api_bot",
+            retries=1,
```
chmnata, line 95: I don't think we need retries for tasks in data checks group.

gabrielwol: Fixed!

dags/sql/select-sensor_id_count_lookback.sql
```diff
@@ -0,0 +1,60 @@
...
+        ARRAY_AGG(DISTINCT {{ params.id_col }}) AS daily_ids --noqa: L039
+    FROM {{ params.table }}
+    WHERE
+        {{ params.dt_col }} >= '{{ ds }} 00:00:00'::timestamp - interval '{{ params.lookback }}'
+        AND {{ params.dt_col }} < '{{ ds }} 00:00:00'::timestamp
+    --group by day then avg excludes missing days.
+    GROUP BY _dt --noqa: L003
+),
+
+today AS (
```
chmnata, line 18: Can we change this to selecting from lookback as well? It is ever so slighter faster this way.

dags/sql/select-sensor_id_count_lookback.sql
```diff
@@ -0,0 +1,60 @@
...
+    AS lookback_avg,
+    'Pass threshold: ' || to_char(
+        FLOOR({{ params.threshold }}::numeric * AVG(lb.lookback_count)),
+        'FM9,999,999,999'
+    )
+    AS passing_value,
+    CASE WHEN array_length(c.ids_diff, 1) > 0
+        THEN 'The following ' || '{{ params.id_col }}' || ' were present within the last '
+            || '{{ params.lookback }}' || ', but absent today:' || c.ids_diff::text
+        ELSE 'All ids present.'
```
chmnata, line 55: All ids present might not be true, Its more like all ids presents within the last X are present

volumes/miovision/api/readme.md
chmnata: can we change the `etc` to `data`?

chmnata:
Confirming logic before merging:
row_counts
- weekday (threshold 0.7 * averages of weekday within past 60 days)
- weekend (threshold 0.7 * averages of weekend or holidays within past 60 days)
- holiday (threshold 0.5 * averages of weekend or holidays within past 60 days)

id counts
- distinct sensor id of today * 0.9 >= distinct sensors of the past 60 days

radumas:
> Confirming logic before merging: row_counts
> 
>     * weekday (threshold 0.7 * averages of weekday within past 60 days)
> 
>     * weekend (threshold 0.7 * averages of weekend or holidays within past 60 days)
> 
>     * holiday (threshold 0.5 * averages of weekend or holidays within past 60 days)
> 

good

> 
> id counts
> 
>     * miovision: distinct sensor id of today * 0.9999 >= distinct intersection_uid of the past 60 days ****
> As @gabrielwol noted, miovision intersection uid count check will fail for 60 days straight if one camera is down. 
> Q1: Do we want to be bombarded alert messages everyday for one down camera? 

This could join with the `anomalous_ranges` table and omit anything with an open-ended range, which would imply the issue has been identified and hopefully being acted upon. But maybe that should be for another issue?

> Q2: We will not get an alert for the down camera after 60 days, is that ok?

Yes, hopefully logged and acted upon and we will get an email when it's back online.

>     * wys : distinct sensor id of today * 0.9999 >= distinct api_ids of the past 60 days
> 
> 

I feel like this could be liable to be screaming constantly when I think Vision Zero ought to be getting email alerts every time a sensor goes down. I think there may already be enough of a backlog of issues to resolve with sensors. Tempted to lower the threshold to.... 0.98% or something.

gabrielwol:
> This could join with the anomalous_ranges table and omit anything with an open-ended range, which would imply the issue has been identified and hopefully being acted upon. But maybe that should be for another issue?  

I'll open a new issue about this! Right now it's a generic function so modifying it for this specific Miovision use case is a slightly larger piece of work.

> I feel like this could be liable to be screaming constantly when I think Vision Zero ought to be getting email alerts every time a sensor goes down. I think there may already be enough of a backlog of issues to resolve with sensors. Tempted to lower the threshold to.... 0.98% or something.

I'm going to change this to 0.9 for now, it looks like there's more variability in WYS api_id counts.

gabrielwol:
> I'm going to change this to 0.9 for now, it looks like there's more variability in WYS api_id counts.

It already is 0.9 🤦‍♂️

chmnata (approved):
Lets merge this!!!

Issue #742, opened 2023-11-07
Update the Miovision Getting Started.md re intersection volume code


scann0n:
Add a code snippet to the Miovision Getting Started doc that specifies how to handle requests for, say, volumes on King Street.

PR #743, opened 2023-11-07
Update getting_started.md

scann0n:
Still waiting to see if the code snippet runs, and what data it should refer to...

- updates miovision's getting started.md with a code snippet and some more discussion on approach vs exit volumes
- 

## Issue(s) this solves:

- #742

- does the code run, and should the reference to `mio_staging` be changed to `miovision_api`?

Nothing


scann0n:
Okay this is ready for review and I nominated @gabrielwol !!!

gabrielwol:
@scann0n I haven't had a chance to fully wrap my head around this and I have to call it quits for this afternoon.
I do have 1 main comment so far which is that there this is missing high level description of which of `volumes_15min` / `volumes_15min_mvt` to use for which purposes. There is some comparison of TMC / ATR but these acronyms don't correspond to the names of the two main aggregate tables which just makes things more confusing. Maybe we need to rename the 15min tables??? 

Having the new code example in [Calculating Volumes on a Segment](https://github.com/CityofToronto/bdit_data-sources/blob/158b5cb943da73598e59c8f45da33b5d4d397b0d/volumes/miovision/getting_started.md#calculating-volumes-on-a-segment) is nice - maybe a corresponding diagram would drive home that point even more.

scann0n:
There's a description of `volumes_15min` and `volumes_15min_mvt` in the main readme and I will certainly refer back to it... I will hold off on any edits (including the diagram, which I have and totally agree with including) until your review is complete!!!

volumes/miovision/getting_started.md
```diff
@@ -98,3 +99,47 @@ The `leg` represents the side of the intersection that the pedestrian is crossin
...
+        AND dv.datetime_bin >= '2023-10-01'::date
+        AND dv.datetime_bin < '2023-11-01'::date
+        AND date_part('isodow', datetime_bin) <= 5 -- weekdays
+        AND ((dv.leg = 'E' AND dv.dir = 'WB')
+            OR (dv.leg = 'W' AND dv.dir = 'EB'))
+    GROUP BY
+        i.intersection_uid,
+        i.intersection_name,
+        dt
+	HAVING COUNT(DISTINCT datetime_bin) > 92
```
radumas, line 133: volume will be `NULL` for time bins that have been processed but where we determined the data to be dubious/gaps.

gabrielwol:
I think rather than the DR-like code we added from Wednesday, as a new user I would prefer a sampling of scenarios we may have to calculate: 

Confused about whether to use `volumes_15min` ("ATR" counts) or `volumes_15min_mvt` ("TMC" counts)? Here are some scenarios in which each may be useful: 
1. Do you need to know specific movements, eg. illegal thru movements on King St?
Use **`miovision_api.volumes_15min_mvt`**!
```sql
SUM(volume) 
FROM miovision_api.volumes_15min_mvt
WHERE movement_uid = 1 --1=thru movement
AND leg IN ('EB', 'WB') --entering the intersection on King
```

2. Do you want to know the total traffic passing through an intersection regardless of movement?
Use **`miovision_api.volumes_15min_mvt`**
```sql
SUM(volume) FROM miovision_api.volumes_15min_mvt
```
OR `miovision_api.volumes_15min` but be careful not to double count entrances and exits!
```sql
SUM(volume) FROM miovision_api.volumes_15min
WHERE leg <> dir --count only entrances
```
Which should equal:
```sql
SUM(volume) FROM miovision_api.volumes_15min
WHERE leg = dir --count only exits
```

3. Do you want to find a screenline? ie. King, east of Bathurst - use ATR (**`miovision_api.volumes_15min`**)!
```sql
SUM(volume) FILTER (WHERE dir = 'EB') AS 'EB volume'
SUM(volume) FILTER (WHERE dir = 'WB') AS 'WB volume'
FROM miovision_api.volumes_15min
WHERE leg = 'EB'
```

4. Any vehicle using King St? 
```sql
SUM(volume) FROM miovision_api.volumes_15min_mvt
WHERE
leg IN ('EB', 'WB') --these vehicles entered intersection on King
OR (leg IN ('SB', 'NB') AND movement_uid IN (2,3)) --Plus, vehicle entering from North or South turning left or right on to King. 
```

gabrielwol (changes requested):
Whoops! I commented last week but did not mark it as "reviewed". 
Let me know what you think of my suggestions.

gabrielwol (approved):
Approve! These new diagrams/examples would have been so helpful when I was getting started - much simpler.

PR #740, opened 2023-11-07
Replicate MOVE's collisions mat views and tables

tahaislam:
## What this pull request accomplishes:

- a new DAG that truncates collision tables under `collisions` and copy new data from `move_staging` into `collisions` tables
- the new DAG is triggered externally by the upstream DAG running on one of MOVE's on-prem servers

## Issue(s) this solves:

- closes #734 

## What, in particular, needs to reviewed:

- The DAG file (the DAG was tested and ran successfully)

## What needs to be done by a sysadmin after this PR is merged

- Update the `data_scripts` folder

## Notes

I updated Airflow configurations to allow external API calls (added `airflow.api.auth.backend.basic_auth` to `api.auth_backends`), and added a new Airflow user for MOVE

PR #748, opened 2023-11-08
Replicate MOVE's counts mat views and tables

### LINKED ISSUE #747: Add a new replicator for MOVE's counts
Add a new counts replicator DAG to save MOVE's tables and materialized views in the `traffic` schema. This DAG should truncate and insert records of transferred objects from the `move_staging` into `traffic` schema every day after the MOVE replicator DAG finishes its daily run.

tahaislam:

- A new DAG that truncates some counts tables under `traffic` and copy new data into them from `move_staging`
- The new DAG is triggered externally by the upstream DAG running on one of MOVE's on-prem servers

## Issue(s) this solves:

- closes #747 

- the DAG file `dags/pull_counts.py`

- Update `data_scripts` folder

## Notes

- Only review after merging #740 as this branch reuses code from that branch
- Out of 11 tables: 8 copied successfully and 3 failed (`arteries_centreline`, `arterydata`, `cnt_det`). They either have different column types or order, or exist as views. They also have a lot of dependencies that need to be saved, dropped, and restored after updating these 3 tables.
  
image


dags/pull_collisions.py
```diff
@@ -0,0 +1,69 @@
+#!/data/airflow/airflow_venv/bin/python3
+# -*- coding: utf-8 -*-
+# noqa: D415
+r"""### The Daily Collision Replicator DAG
+
+This DAG runs daily to update the following collisions tables from the tables
+staged in the database by the MOVE's ``bigdata_replicator`` DAG:
```
gabrielwol, line 7: Can you please add a note about how these DAGs are triggered by the `bigdata_replicator` DAG? It is not obvious at first when you see the schedule interval.

dags/pull_counts.py
```diff
@@ -0,0 +1,90 @@
...
+    max_active_runs=5,
+    schedule_interval=None,
+    doc_md=__doc__,
+    tags=["counts"]
+)
+def counts_replicator():
+    """The main function of the counts DAG."""
+    from dags.common_tasks import wait_for_external_trigger, copy_table
+
+    tables = [
```
gabrielwol, line 73: I do wonder if dynamic task mapping is the best choice for this when the tables to be replicated are hard coded in the DAG. In this case it may be more clear to generate them with a for loop (the old dynamic way), perhaps adding start/end dates to the list. If the tables are changing every day, then it would make more sense to use dynamic task mapping and pass the tables to be replicated with the xcom value from the external trigger. 
Obviously this approach is working fine, but curious to hear your thoughts!

dags/pull_counts.py
```diff
@@ -0,0 +1,90 @@
...
+    "retries": 3,
+    "retry_delay": timedelta(minutes=60),
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+@dag(
+    dag_id=DAG_NAME,
+    default_args=default_args,
+    catchup=False,
+    max_active_runs=5,
```
gabrielwol, line 64: Perhaps you should also limit concurrency (max active tasks).

gabrielwol (changes requested):
Good example with the common tasks!
A few minor comments.

dags/pull_counts.py
tahaislam: you're right. As per Airflow documentation, this's the simplest dynamic task mapping. I was going to read the table names from an Airflow variable to avoid hard-coding these names.
I also second your proposal to get the list of tables to be copied from the external trigger. But I considered this as a future work as it requires some more work to send this list of names using Airflow APIs between the two DAGs and to create these tables if they don't already exist. To keep this PR simpler and get the DAGs ready sooner, I'll create a new issue for this *"improvements"* to be done in the future.

tahaislam:
@gabrielwol: ready for re-review

gabrielwol (approved):
Nice changes!

PR #749, opened 2023-11-08
744 miovision DST fix

### LINKED ISSUE #744: Miovision pulls duplicate records over EDT->EST transition
[Logs](https://trans-bdit.intra.prod-toronto.ca/airflow/dags/pull_miovision/grid?base_date=2023-11-07T23%3A59%3A59.999Z&dag_run_id=scheduled__2023-11-05T08%3A00%3A00%2B00%3A00&task_id=pull_miovision&tab=logs)

The duplicate records, via a manual API call:
```
(68, '2023-11-05T01:59:00.000-04:00', '1', 'N', '2', 2),
(68, '2023-11-05T01:59:00.000-05:00', '1', 'N', '2', 1),
```

gabrielwol:
A fairly small PR to drop the 2nd 1AM on EDT->EST transition in pull_miovision. 

## Issue(s) this solves:
- Closes #744 

- I tested locally, everything seems to be working as intended.

Test this DAG by re-running Nov 5, 2023.


chmnata (approved):
Looks good!

chmnata:
I will test it with Nov 5 :meow_salute:

chmnata:
Looks like its working! :meow_wow:
Kinda resulting in quite a loooooong log file, we could remove all `1AM, UTC-05:00` right before inserting technically, but I think since it only happens once a year for one hour only its fine 👍🏼 
image

tahaislam:
I know you already researched this a lot, but what do you think of averaging the two hours instead of dropping one of them? Is it easy to implement? Would it make any difference?
Similarly, we might interpolate the missing hour in March as well 🤔

gabrielwol:
because it's an overnight hour with low volume which will never be very useful for analysis, I think this lower effort solution is good enough. 
For the March time change, I do wonder if we need to implement some special cases to make sure that day is not labelled as a "bad day" bc of missing hour, ie. in Miovision.

PR #760, opened 2023-11-14
746 reorganize miovision DAG

### LINKED ISSUE #746: Reorganize pull_miovision DAG
We should split up the `pull_miovision` task in the `pull_miovision` DAG, particularly the data pull / aggregations to make it easier to re-run certain parts of the pipeline. 
- Change from BashOperator to python/task flow. 
- Separate tasks for each aggregation
- Clear raw table/aggregations prior to inserts.

gabrielwol:
**PR REPLACED BY #822** 

  
- New taskflow refactored Miovision DAG as `miovision_pull` 
  - Keeps old `pull_miovision` with end date of 2023-12-01. 
- Updates to `intersection_tmc.py`:
  - Allows the aggregations to be called individually so they can be separated out into Airflow tasks
  - clear each of the miovision tables prior to inserting
    - `miovision_api.volumes`, `miovision_api.api_log` and `miovision_api.volumes_15min_mvt` need options to clear only an individual intersection and so were given SQL functions.
    - Other tables can be cleared and re-run in their entirety even if just a single intersection is pulled, so I used more simple DELETE queries directly in `intersection_tmc.py` 
- Improved the single intersection option; 
  - Added an airflow parameter so it can be used in run config
- Note: 31 sql files were organized into folders with no changes.
  - Updates to readmes to fix links to reorganized sqls.

## Issue(s) this solves:
- Closes #746 

- I waffled between clearing all tables at once, or clearing them individually before inserting. I went with the latter.
  - Potential downside is if someone inadvertently runs the `pull_miovision` task without the downstream aggregations, then there could be a mismatch between the raw `volumes` table and the downstream aggregate tables.

4 new sql functions to be added for clearing aggregate tables:
- [x] volumes/miovision/sql/function-clear-api_log.sql
- [x] volumes/miovision/sql/function-clear-volumes_15min.sql
- [x] volumes/miovision/sql/function-clear-volumes_15min_mvt.sql
- [x] volumes/miovision/sql/function-clear-volumes.sql

One edited sql function (drop original (date, date) and replace with (date, date, integer):
- [ ] volumes/miovision/sql/function-api_log.sql

- [ ] Adjust the start/end dates of the two DAGs when this PR gets reviewed.

- [ ] Assuming #738 gets merged before this one, we will need to manually add the data check tasks to the new `miovision_pull` DAG, as the file name has changed and it won't get merged cleanly.


volumes/miovision/sql/function/function-clear-api_log.sql
```diff
@@ -0,0 +1,31 @@
+﻿CREATE OR REPLACE FUNCTION miovision_api.clear_api_log(
+    _start_date date,
+    _end_date date,
+    intersection integer DEFAULT NULL)
```
chmnata, line 4: maybe good if it takes an array, in case we want to get rid of a couple ones that fails (? 
just a thought 🧠

volumes/miovision/sql/function/function-clear-api_log.sql
```diff
@@ -0,0 +1,31 @@
+﻿CREATE OR REPLACE FUNCTION miovision_api.clear_api_log(
```
chmnata, line 1: fyi created in the bigdata

volumes/miovision/sql/function/function-clear-volumes.sql
```diff
@@ -0,0 +1,31 @@
+﻿CREATE OR REPLACE FUNCTION miovision_api.clear_volumes(
```
chmnata, line 1: created in bigdata

volumes/miovision/sql/function/function-clear-volumes_15min.sql
```diff
@@ -0,0 +1,28 @@
+CREATE OR REPLACE FUNCTION miovision_api.clear_volumes_15min(
```
chmnata, line 1: created!
no specifying for intersection this time?

volumes/miovision/sql/function/function-clear-volumes_15min_mvt.sql
```diff
@@ -0,0 +1,64 @@
+CREATE OR REPLACE FUNCTION miovision_api.clear_15_min_mvt(
```
chmnata, line 1: created!

volumes/miovision/sql/function/function-clear-api_log.sql
```diff
@@ -0,0 +1,31 @@
...
+    _start_date date,
+    _end_date date,
+    intersection integer DEFAULT NULL)
+RETURNS void
```
chmnata, line 5: not like an important feature but it will be helpful if we can see how many rows got deleted, or if anything got deleted at all

volumes/miovision/sql/function/function-clear-volumes_15min.sql
gabrielwol: IIRC it's because this function is much faster than the 15min_mvt one so it's not as big a deal to re-run the whole date. 
~2 minutes vs ~20s
image

PR #764, opened 2023-11-15
746 add clear miovision task

gabrielwol:
## What this pull request accomplishes:

- 
- 

## Issue(s) this solves:
- Closes #746 

## What, in particular, needs to reviewed:
-

PR #763, opened 2023-11-15
752 miovision standardize definition of day

gabrielwol:
## What this pull request accomplishes:
- Standardizes definition of a day. No more 11pm-11pm aggregations. 
  - `volumes_15min` and `volumes_15min_mvt` were straightforward to update, just removing `- interval '1 hour'` references. Most of the changed lines come from fluffing these old scripts. 
  - `volumes_daily` was able to be simplified after changing the time period, removing two now unnecessary columns. 

## Issue(s) this solves:

- Closes #752 

## What, in particular, needs to reviewed:
- 

## What needs to be done by a sysadmin after this PR is merged
- [x] Update definitions of four modified functions in bigdata.

- Once this PR is merged and we switch from 11pm-11pm aggregation to 12am-12am aggregation, there will be 1 hour that needs to be manually aggregated for `volumes_15min` and `volumes_15min_mvt`:  
- [x] Update `volumes_15min_mvt`
```sql
    WITH aggregate_delete AS (
        --delete existing aggregation.
        DELETE FROM miovision_api.volumes_15min_mvt
        WHERE
            datetime_bin >= start_date
            AND datetime_bin < end_date
        RETURNING volume_15min_mvt_uid
    )

    --To update foreign key for 1min bin table
    UPDATE miovision_api.volumes AS a
    SET volume_15min_mvt_uid = NULL
    FROM aggregate_delete AS b
    WHERE
        a.volume_15min_mvt_uid = b.volume_15min_mvt_uid
        AND a.datetime_bin >= start_date
        AND a.datetime_bin < end_date;

--insert new aggregation
SELECT miovision_api.aggregate_15_min_mvt(start_date, end_date);
```
- [x] Update `volumes_15min`
```sql
    --delete existing records
    DELETE FROM miovision_api.volumes_15min
    WHERE
        datetime_bin >= start_date
        AND datetime_bin < end_date; 

    --remove processed flag in volumes_15min_mvt
    UPDATE miovision_api.volumes_15min_mvt
    SET processed = NULL
    WHERE 
        datetime_bin >= start_date
        AND datetime_bin < end_date;

--insert new aggregation 
SELECT miovision_api.aggregate_15_min(start_date, end_date);
```
- And for `volumes_daily` we need to drop these two columns, replace function and replace the table contents: 
- [x] Update `volumes_daily` table: 
```sql
ALTER TABLE miovision_api.volumes_daily DROP COLUMN period_start;
[...]

Issue #630, opened 2023-06-08
Detect Unusual Miovision Volume Counts Automagically 


scann0n:
Through issue #570, weekly volume counts for all Miovision cameras were graphed and weeks with unusual volumes were visually identified (the ranges of bad dates per camera are stored in `miovision_api.bad_data_ranges`).

Is there a way to identify these bad data ranges without the time consuming visual inspection piece? Let's find out!

gabrielwol:
I've opened #777 which would start to address this issue by automatically including zero count periods in `miovision_api.anomalous_ranges`. Further work will be required for periods with non-zero low/high volumes.

Nate-Wessel:
I'd like a chance to review this before anything gets added automatically to `anomalous_ranges`. I'm off for the next couple of days though. 

I'm a little anxious about automation letting this table quickly get very big, beyond our abilities to manually check... though it's definitely the right place to flag these issues.

gabrielwol:
No rush. Idea: we could keep manual and automated gap ID in separate tables and combine with a view if you prefer.

Nate-Wessel:
> No rush. Idea: we could keep manual and automated gap ID in separate tables and combine with a view if you prefer.

There's actually already a way to mark ranges not yet checked by a human. See `miovision_api.anomaly_investigation_levels`'s `auto-flagged` value.

PR #777, opened 2023-11-22
#630 detect zero miovision volume counts automagically

gabrielwol:
- Automatically add/appends to `anomalous_ranges` table zero volume days for classifications 1,2,6,10. Includes special case for when all these classifications are zero (classification_uid becomes NULL as per existing convention). 
  - **Major change from original PR:** The new gaps are now automatically appended to existing gaps.
- Adds a new task to `miovision_pull` DAG to run this aggregation daily. 
- Add missing `miovision_api.anomaly_problem_levels`, `miovision_api.anomaly_investigation_levels` table definitions to data-sources github (unchanged).  
- Updates readme to reflect changes to anomalous_ranges table.

## Issue(s) this solves:
- A start to tackling #630

~~The new format of automated daily inserts into `anomalous_ranges` is not a breaking change, but the format is slightly different than the existing entries which are generally multi-day. I think a future improvement could involve aggregating new entries to anomalous_ranges table with existing records using https://pgxn.org/dist/range_agg/.~~

- [x] ~~**Note:** #763 should be merged first (consistent definition of a day)~~
- [ ] **Note:** #764 should be merged first (reorganize DAG)
- [ ] Add new sql function to miovision_api. 
- [ ] Update definition of `miovision_api.anomalous_ranges`:
```sql
ALTER TABLE miovision_api.anomalous_ranges
ADD COLUMN range_start timestamp DEFAULT null,
ADD COLUMN range_end timestamp DEFAULT null;

UPDATE miovision_api.anomalous_ranges
SET
    range_start = 
        CASE
            --if not inclusive, move forward to make inclusive
            WHEN lower_inc(time_range) IS FALSE
                THEN LOWER(time_range) + interval '1 day'
            ELSE LOWER(time_range)
        END, 
    range_end =
        CASE
            --if inclusive, move forward to make non-inclusive
            WHEN upper_inc(time_range) IS TRUE
                THEN UPPER(time_range) + interval '1 day'
            ELSE UPPER(time_range)
        END;

ALTER TABLE miovision_api.anomalous_ranges
ADD CONSTRAINT anomalous_ranges_range_order
CHECK (
    --results for check constraints of true or null are allowed
    range_start < range_end
);

COMMENT ON TABLE miovision_api.anomalous_ranges
IS 'Table to store anomalous ranges of Miovision data. 
Values are added to this table both manually by miovision_admins/miovision_data_detectives and
automatically through a daily Airflow process that runs function miovision_api.identify_zero_counts(date).';

COMMENT ON COLUMN miovision_api.intersection_uid
IS 'NULLs are allowed in the column intersection_uid to denote
anomalous that affect all intersections.';

COMMENT ON COLUMN miovision_api.classification_uid
IS 'NULLs are allowed in the column classification_uid to denote
anomalous that affect all classifications.';

COMMENT ON COLUMN miovision_api.anomalous_ranges.range_start
IS 'Start of anomalous range. Inclusive. NULL values represent an unbounded range_start.';

COMMENT ON COLUMN miovision_api.anomalous_ranges.range_start
IS 'End of anomalous range. Exclusive. NULL values represent an unbounded range_end.';

--drop and recreate the constraint for range_start/range_end instead of time_range.
ALTER TABLE miovision_api.anomalous_ranges
DROP CONSTRAINT miovision_qc_intersection_uid_classification_uid_time_range_key;

ALTER TABLE miovision_api.anomalous_ranges
ADD CONSTRAINT miovision_qc_intersection_uid_classification_uid_time_range_key
UNIQUE NULLS NOT DISTINCT (intersection_uid, classification_uid, range_start, range_end);

--drop deprecated time_range column
ALTER TABLE miovision_api.anomalous_ranges
DROP COLUMN time_range;
```


volumes/miovision/sql/function/function-identify-zero-counts.sql
```diff
@@ -0,0 +1,86 @@
+CREATE OR REPLACE FUNCTION miovision_api.identify_zero_counts(
+    start_date timestamp,
+    end_date timestamp,
```
Nate-Wessel, line 3: What kind of `start_date`, `end_date` values will this function be run with? A one-day interval? Something else?

Nate-Wessel: Should these be called `_date` when they are actually timestamps?

volumes/miovision/sql/function/function-identify-zero-counts.sql
```diff
@@ -0,0 +1,86 @@
...
+            intersection_uid,
+            null::integer AS classification_uid,
+            tsrange(MIN(datetime_bin), MAX(datetime_bin) + interval '15 minutes', '[)') AS time_range,
+            'zero counts' AS notes,
+            'auto_flagged' AS investigation_level,
+            'do-not-use' AS problem_level
+        FROM miovision_api.volumes_15min_mvt
+        WHERE
+            classification_uid IN (1,2,6,10) 
+            AND datetime_bin > start_date - min_gap --needed to detect gaps overlapping daily intervals
```
Nate-Wessel, line 25: should be greater than _or equal_ (`>=`)?

volumes/miovision/sql/function/function-identify-zero-counts.sql
gabrielwol: Yes, a one day interval run by the daily Airflow DAG included in this PR. 
I'll consider changing the variables names, but this is pretty consistent with the other Miovision function conventions.

Nate-Wessel: It makes sense if they'll always be date values, essentially.

Nate-Wessel:
So for each day with zero data (for however many cameras are down at any given moment) this will create a record for that day, and possibly per mode if just a classification or two are malfunctioning. I'm very interested in 

>  I think a future improvement could involve aggregating new entries to anomalous_ranges table with existing records using https://pgxn.org/dist/range_agg/.

But I suppose we can always do that retroactively. If this starts running and filling up the table, I may be motivated to come up with that aggregation query myself :-)

...

I'm not sure I understand the role of the `min_gap` variable here, since it's just aggregating to the supplied date/time range, 1 day. If there were a 1 hour camera outage, this function would not detect it, no?

gabrielwol:
> I'm not sure I understand the role of the `min_gap` variable here, since it's just aggregating to the supplied date/time range, 1 day. If there were a 1 hour camera outage, this function would not detect it, no?

🤦 In my mind this PR was ready, but clearly it was not. My intention was to capture gaps in intervals of `min_gap` using a gaps and island method.

gabrielwol:
An example of how I'm testing the new script (with a temp copy of anomalous_ranges in gwolofs schema): 
<details><summary>(long sql)</summary>

```sql
DROP TABLE gwolofs.anomalous_ranges_test;
CREATE TABLE gwolofs.anomalous_ranges_test (
    uid serial PRIMARY KEY,
    intersection_uid smallint REFERENCES miovision_api.intersections (intersection_uid),
    classification_uid smallint REFERENCES miovision_api.classifications (classification_uid),
    time_range tsrange NOT NULL,
    range_start timestamp NOT NULL,
    range_end timestamp,
    notes text NOT NULL,
    investigation_level text NOT NULL,
    problem_level text NOT NULL,
    UNIQUE (intersection_uid, classification_uid, range_start, range_end)
);

--some fake test values, some of which overlap and some of which don't. 
INSERT INTO gwolofs.anomalous_ranges_test(intersection_uid, classification_uid, time_range, range_start, range_end, notes, investigation_level, problem_level)
( VALUES
    (NULL, 10, tsrange('2023-10-31 00:00:00'::timestamp,'2023-11-01 00:00:00'::timestamp, '[)'), '2023-10-31 00:00:00'::timestamp,'2023-11-01 00:00:00'::timestamp,	'Zero counts, identified by a daily airflow process running function miovision_api.identify_zero_counts', 'auto_flagged',	'do-not-use'),
    (45,NULL, tsrange('2023-10-31 00:00:00'::timestamp,'2023-11-01 00:00:00'::timestamp, '[)'), '2023-10-31 00:00:00'::timestamp,'2023-11-01 00:00:00'::timestamp,	'Zero counts, identified by a daily airflow process running function miovision_api.identify_zero_counts', 'auto_flagged',	'do-not-use'),
    (45,NULL, tsrange('2023-11-02 00:00:00'::timestamp,'2023-11-03 00:00:00'::timestamp, '[)'), '2023-11-02 00:00:00'::timestamp,'2023-11-03 00:00:00'::timestamp,	'Zero counts, identified by a daily airflow process running function miovision_api.identify_zero_counts', 'auto_flagged',	'do-not-use'),
    (20, 10, tsrange('2023-10-31 00:00:00'::timestamp,'2023-11-01 00:00:00'::timestamp, '[)'), '2023-10-31 00:00:00'::timestamp,'2023-11-01 00:00:00'::timestamp,	'Zero counts, identified by a daily airflow process running function miovision_api.identify_zero_counts',	'auto_flagged',	'do-not-use'),
    (7,	NULL, tsrange('2023-10-31 00:00:00'::timestamp,'2023-11-01 00:00:00'::timestamp, '[)'), '2023-10-31 00:00:00'::timestamp,'2023-11-01 00:00:00'::timestamp,	'Zero counts, identified by a daily airflow process running function miovision_api.identify_zero_counts',	'auto_flagged',	'do-not-use'),    --this test value should prevent an outage from intersection 63 on 11-01 from being added.
    (63,NULL, tsrange('2023-10-01 00:00:00'::timestamp,NULL, '[)'), '2023-10-01 00:00:00'::timestamp, NULL, 'Zero counts, identified by a daily airflow process running function miovision_api.identify_zero_counts',	'auto_flagged',	'do-not-use')
);

    WITH zero_intersections AS (
        SELECT
            intersection_uid,
            MIN(datetime_bin) AS range_start,
            MAX(datetime_bin) + interval '15 minutes' AS range_end
        FROM miovision_api.volumes_15min_mvt
        WHERE
            datetime_bin >= '2023-11-01'::date
            AND datetime_bin < '2023-11-02'::date
        GROUP BY intersection_uid
        HAVING SUM(volume) = 0
    ),
    
    new_gaps AS (
        SELECT
            v15.intersection_uid,
            v15.classification_uid,
            MIN(v15.datetime_bin) AS range_start,
            MAX(v15.datetime_bin) + interval '15 minutes' AS range_end
        FROM miovision_api.volumes_15min_mvt AS v15
        --anti join intersections with periods of zero volume for all modes
        LEFT JOIN zero_intersections AS zi ON 
            zi.intersection_uid = v15.intersection_uid
            AND v15.datetime_bin >= zi.range_start
            AND v15.datetime_bin < zi.range_end
        WHERE
            zi.intersection_uid IS NULL
            AND datetime_bin >= '2023-11-01'::date
            AND datetime_bin < '2023-11-02'::date
            --this script will only catch zeros for classification_uid 1,2,6,10
            --since those are the ones that are zero padded. Filter for speed.
            AND v15.classification_uid IN (1,2,6,10)
        GROUP BY
            v15.classification_uid,
            v15.intersection_uid
        HAVING SUM(volume) = 0

        UNION

        SELECT
            intersection_uid,
            null::integer AS classification_uid,
            range_start,
            range_end
        FROM zero_intersections
    ),
    
    updated_values AS (
        --update new outages which are contiguous with old outages
        UPDATE gwolofs.anomalous_ranges_test AS existing
        SET
            time_range = tsrange(existing.range_start, new_.range_end, '[)'),
            range_end = new_.range_end
        FROM new_gaps AS new_
        WHERE
            --deal with null values in intersection, classification
            COALESCE(existing.intersection_uid, 0) = COALESCE(new_.intersection_uid, 0)
            AND COALESCE(existing.classification_uid, 0) = COALESCE(new_.classification_uid, 0)
            --only merging ranges that overlap in (0)->(new) direction
            --otherwise if days run out of order we have to deal with
            --recursive situation of collapsing (0)<-(new)->(2)
            AND existing.range_end = new_.range_start
            AND existing.notes = 'Zero counts, identified by a daily airflow process running function miovision_api.identify_zero_counts'
            AND existing.investigation_level = 'auto_flagged'
        --returns the rows from `new_gaps` which were used to modify existing.
        RETURNING new_.*
    )
    
    INSERT INTO gwolofs.anomalous_ranges_test(
        intersection_uid, classification_uid, time_range, range_start, range_end, notes, investigation_level, problem_level)
    SELECT
        new_gaps.intersection_uid,
        new_gaps.classification_uid,
        tsrange(new_gaps.range_start, new_gaps.range_end, '[)') AS time_range,
        new_gaps.range_start,
        new_gaps.range_end,
        'Zero counts, identified by a daily airflow process running function miovision_api.identify_zero_counts' AS notes,
        'auto_flagged' AS investigation_level,
        'do-not-use' AS problem_level
    FROM new_gaps
    --anti join values which were already used for above updates
    LEFT JOIN updated_values ON
        COALESCE(new_gaps.intersection_uid, 0) = COALESCE(updated_values.intersection_uid, 0)
        AND COALESCE(new_gaps.classification_uid, 0) = COALESCE(updated_values.classification_uid, 0)
        AND new_gaps.range_start = updated_values.range_start
    --anti join existing open ended gaps
    LEFT JOIN gwolofs.anomalous_ranges_test AS existing ON
        COALESCE(existing.intersection_uid, 0) = COALESCE(new_gaps.intersection_uid, 0)
        AND COALESCE(existing.classification_uid, 0) = COALESCE(new_gaps.classification_uid, 0)
        AND existing.range_end IS NULL
    WHERE
        existing.uid IS NULL
        AND updated_values.range_start IS NULL;
    
--check results
SELECT * FROM gwolofs.anomalous_ranges_test;
```
</details>

gabrielwol:
@Nate-Wessel major updates to this PR. Back to focusing only on date ranges and not sub-day ranges (those are covered by find_gaps). 
Also takes care of appending new ranges to existing which was a major concern with the first version. 
Re-review when you're ready!

volumes/miovision/sql/table/create_anomalous_ranges.sql
```diff
@@ -30,6 +31,8 @@ CREATE TABLE miovision_api.anomalous_ranges (
     intersection_uid smallint REFERENCES miovision_api.intersections (intersection_uid),
     classification_uid smallint REFERENCES miovision_api.classifications (classification_uid),
     time_range tsrange NOT NULL,
+    range_start timestamp NOT NULL,
+    range_end timestamp,
```
Nate-Wessel, line 35: Aren't these redundant to the `time_range` field? You could get them with 

```sql
lower(time_range),
upper(time_range)
```

https://www.postgresql.org/docs/current/functions-range.html

volumes/miovision/sql/function/function-identify-zero-counts.sql
```diff
@@ -0,0 +1,119 @@
...
+    )
+    
+    INSERT INTO miovision_api.anomalous_ranges(
+        intersection_uid, classification_uid, time_range, range_start, range_end, notes, investigation_level, problem_level)
+    SELECT
+        new_gaps.intersection_uid,
+        new_gaps.classification_uid,
+        tsrange(new_gaps.range_start, new_gaps.range_end, '[)') AS time_range,
+        new_gaps.range_start,
+        new_gaps.range_end,
```
Nate-Wessel, line 88: Yeah, this is not ideal for the manual update process I've been doing. Why the duplication? It also won't enforce that start is before end, doesn't indicate whether the values are inclusive or exclusive, etc. Why reinvent the range type?

Nate-Wessel (changes requested):
See inline comments.

volumes/miovision/sql/table/create_anomalous_ranges.sql
Nate-Wessel: What was the problem with indexing?

gabrielwol: I created a small test case and in-line with Raph's suggestion, the explicit range_start and range_end was able to use the volumes datetime_bin index and ran about 50% faster. 
Explain Plan using `time_range @> datetime_bin`:
image
vs
Explain plan using `datetime_bin >= LOWER(time_range) AND datetime_bin < UPPER(time_range)` (and similar for `datetime_bin >= range_start AND datetime_bin < range_end`:
image

volumes/miovision/sql/function/function-identify-zero-counts.sql
gabrielwol: I will add a check constraint on the column order, and column comments.

volumes/miovision/sql/table/create_anomalous_ranges.sql
Nate-Wessel: Sorry, I'm not sure I'm totally following. @gabrielwol are you saying that creating an index on certain fields on the `anomalous_ranges` table speeds up the join with one or more of the volumes tables? 

I don't know why you would ever do `datetime_bin >= LOWER(time_range) AND datetime_bin < UPPER(time_range)` rather than `time_range @> datetime_bin`. It's not surprising that's less efficient, right?

gabrielwol: @Nate-Wessel other way around: `datetime_bin >= LOWER(time_range) AND datetime_bin < UPPER(time_range)` was 50% faster than `time_range @> datetime_bin`.
The above screenshots were showing that the ts_range method was NOT able to use the index on the datetime_bin column of volumes, while the explicit method was.

Nate-Wessel: Oh, huh. Interesting! I'm surprised Postgres hasn't optimized this yet for these built-in types and operators. I'd expect the planner to be smarter than that. 

If that works efficiently with the `lower` and `upper` functions in the query, can we remove the duplicate fields on the table? Alternatively we could remove the tsrange field, though there are already a number of existing things that rely on that - they would have to be refactored. 

I'd just really like to avoid having three columns all saying the same thing in slightly different ways.

radumas: >Oh, huh. Interesting! I'm surprised Postgres hasn't optimized this yet for these built-in types and operators. I'd expect the planner to be smarter than that.

I'm sure they have their reasons, but yeah I was also surprised by this. There are a lot of optimizations that seem intuitive for timestamps and yet... they don't exist.

I agree we shouldn't have duplicative columns (though we can ensure they are equivalent with triggers, but that's a real belt & suspender situation). My preference would be to use the timestamp columns going forward, so people don't get tripped up with the ranges and get sad when their queries take way longer than they expect. If you've got higher priorities than refactoring we could punt deleting the `tsrange` column until you can take that on?

gabrielwol: There's only two dependents in bigdata: `gwolofs.hourly_vehicle_volumes`, `monitoring.vi_daily_totals_miovision`
and one reference in bigdata (in this PR) that we'll have to update as far as I can tell.

Nate-Wessel: I'm ok with that. Sad that the thing that seems like it should be the clear solution isn't great though :-/ Oh well. 

May I suggest an interim check constraint rather than a trigger? I've found triggers to be fairly opaque. Something like:

```sql
CHECK(range_start = lower(time_range)),
CHECK(range_end = upper(time_range))
```
I'm not 100% sure how that works with `[)` vs `[]` etc but it shouldn't be too hard to figure out.

gabrielwol: Micro lightning talk: you can use `lower_inc(time_range), upper_inc(time_range)` to check if bounds are inclusive/exclusive! 
https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-INCLUSIVITY

Issue #673, opened 2023-08-15
miovision_api.find_gaps improvements


gabrielwol:
Looking at the pull_miovision airflow logs, this step alone takes 40 minutes to run. 

- Explain shows it includes multiple seq scans over `miovision_api.volumes` table
- 10+ CTEs
- long and not well commented
- use of `ful` table probably not necessary.
- needs fluffing for readability 
- Fails silently for intersections with no data at all
- For gaps overlapping multiple hours, only the start of the hour is used to determine the allowed_gap threshold.
- Holidays not considered when determining weekdays/weekends
- gap sizes are exaggerated by 1 minute, since both gap_start and gap_end are exclusive. 

@scann0n's gap and islands approach for rescu can be used for inspiration (runs in just a few seconds for whole year of rescu data): https://github.com/CityofToronto/bdit_data-sources/blob/d85291a5b76b65cc6348f5d30949a1887b35e236/volumes/rescu/date_evaluation/x2_rescu_gaps.sql

radumas:
Also relevant:
https://github.com/CityofToronto/bdit_data-sources/issues/3700
https://github.com/CityofToronto/bdit_data-sources/issues/419

gabrielwol:
@radumas Another issue with this function is it accepts `start_date` and `end_date` to be run for multiple days, however it uses a 60 day lookback to identify an acceptable threshold based only on end_date. In other words it may not produce the same results backfilling over a period as it would when run daily. 
Alternatives: 
1) Change to only accept start_date and run for a single date. I think we could still run multiple days at a time with a LATERAL command. 
2) Set new thresholds at the `intersections` table level which are not calculated based on end_date. Perhaps 1 single threshold across all sensors (15 minutes?). This is the current threshold case statement, where `avg(mio.vol)` is the avg volume for that hour over the previous 60 days.  
```
CASE
    WHEN avg(mio.vol) < 100::numeric THEN 20
    WHEN avg(mio.vol) >= 100::numeric AND avg(mio.vol) < 500::numeric THEN 15
    WHEN avg(mio.vol) >= 500::numeric AND avg(mio.vol) < 1500::numeric THEN 10
    WHEN avg(mio.vol) > 1500::numeric THEN 5
    ELSE NULL::integer
END AS gap_tolerance 
```

gabrielwol:
[...]

PR #783, opened 2023-11-30
780 create a daily miovision table for data collection

### LINKED ISSUE #780: Create a daily Miovision table for Data Collection
- Data Collection team is seeking daily volumes, so do not filter out entire hours because small gaps (5 minutes) are missing! This is the status quo in the current `volumes` -> `volumes_15min_mvt` -> `volumes_daily aggregation`.
- This should be fast to query, so it should be a table (not a mat view since it is updated daily). 
- Make it obvious that a day has unacceptable gaps, but do not exclude. Let end user come up with thresholds. 
- Exclude `anomalous_ranges`.

gabrielwol:
Refactor the Miovision daily summary table for use by Data Collection AADT purposes. 
- Switch `classification_uid` from wide to long format.
- No longer omits full hours containing unacceptable gaps - only omit the gap itself (extended zero volume periods, rounded up to 15 minute bins)
- Speed improvement for querying daily table by pre-summarizing and adding reporting view columns.

## Issue(s) this solves:
- Closes #780

- 

```sql
DROP TABLE miovision_api.volumes_daily;
--create new table: volumes/miovision/sql/table/create-table-volumes_daily.sql
--replace function: volumes/miovision/sql/function/function-aggregate_volumes_daily.sql
--create new view: volumes/miovision/sql/views/create-view-volumes_daily.sql

--backfill new table from staging + any new dates:
INSERT INTO miovision_api.volumes_daily SELECT * FROM gwolofs.volumes_daily;
SELECT miovision_api.aggregate_volumes_daily('2024-01-22'::date, current_date)
```
- remind Nate to switch https://github.com/Toronto-Big-Data-Innovation-Team/monitoring/pull/57 to use miovision_api schema in order to drop `gwolofs.volumes_daily`


gabrielwol:
Note to self to add `SUM(un.avg_historical_vol)` to this table after #673 is merged.

volumes/miovision/sql/create-table-daily_vehicle_volumes.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TABLE miovision_api.daily_vehicle_volumes(
+    intersection_uid integer,
+    dt date,
+    vehicle_volume integer,
+    isodow smallint,
+    holiday smallint,
```
Nate-Wessel, line 6: Why an int and not a boolean?

volumes/miovision/sql/function-daily_vehicle_volumes.sql
```diff
@@ -0,0 +1,86 @@
...
+        --only anomalous ranges for vehicles
+        ar.classification_uid = v.classification_uid
+        OR ar.classification_uid IS NULL
+    )
+    AND v.datetime_bin >= LOWER(ar.time_range)
+    AND v.datetime_bin < UPPER(ar.time_range)
+    AND ar.problem_level IN ('do-not-use', 'questionable')
+WHERE
+    ar.time_range IS NULL --anti join anomalous ranges
+    AND c.class_type = 'Vehicles' --all types
```
Nate-Wessel, line 66: FYI, this would exclude buses and streetcars which, bizarrely, are not considered "Vehicles" in `miovision_api.classifications`. Note also that streetcars became not-vehicles in August of 2019.

Nate-Wessel:
> ```sql
>     SUM(v.volume) FILTER (WHERE v.classification_uid IN (2, 10)) AS bike_volume,
> ```

We shouldn't add 2's and 10's; each is theoretically counting _all_ the bikes on its own, so this would be double counting. 

I would probably incline toward keeping the classifications separate if that's an option.

radumas (changes requested):
One smol change. Plus the larger changes we discussed yesterday (have classification be a column so we can have 1 daily row per classification

gabrielwol:
@ZahraGhayen @Nate-Wessel
Here is the latest sample table (You can see more at: `SELECT * FROM gwolofs.volumes_daily;`)
Thanks for your patience, this required some significant changes to another Miovision PR, #678. 

| "dt"         | "intersection_uid" | "classification_uid" | "daily_volume" | "isodow" | "holiday" | "datetime_bins_missing" | "unacceptable_gap_minutes" | "avg_historical_gap_vol" |
|--------------|--------------------|----------------------|----------------|----------|-----------|-------------------------|----------------------------|--------------------------|
| "2023-11-01" | 12                 | 1                    | 32217          | 3        | false     | 26                      | 15                         | 394                      |
| "2023-11-01" | 12                 | 2                    | 4514           | 3        | false     | 353                     | 15                         | 127                      |
| "2023-11-01" | 12                 | 3                    | 1780           | 3        | false     | 557                     | 15                         | 21                       |
| "2023-11-01" | 12                 | 4                    | 941            | 3        | false     | 885                     | 15                         | 9                        |
| "2023-11-01" | 12                 | 5                    | 30             | 3        | false     | 1410                    | 15                         | 0                        |
| "2023-11-01" | 12                 | 6                    | 27153          | 3        | false     | 112                     | 15                         | 710                      |
| "2023-11-01" | 12                 | 9                    | 56             | 3        | false     | 1387                    | 15                         | 0                        |
| "2023-11-01" | 12                 | 10                   | 2411           | 3        | false     | 543                     | 15                         | 68                       |
| "2023-11-01" | 15                 | 1                    | 26479          | 3        | false     | 18                      | 0                          |                          |
| "2023-11-01" | 15                 | 2                    | 4038           | 3        | false     | 406                     | 0                          |                          |
| "2023-11-01" | 15                 | 3                    | 939            | 3        | false     | 797                     | 0                          |                          |
| "2023-11-01" | 15                 | 4                    | 791            | 3        | false     | 922                     | 0                          |                          |
| "2023-11-01" | 15                 | 5                    | 21             | 3        | false     | 1419                    | 0                          |                          |
| "2023-11-01" | 15                 | 6                    | 30449          | 3        | false     | 147                     | 0                          |                          |
| "2023-11-01" | 15                 | 9                    | 17             | 3        | false     | 1423                    | 0                          |                          |
| "2023-11-01" | 15                 | 10                   | 2353           | 3        | false     | 587                     | 0                          |                          |

gabrielwol:
@Nate-Wessel this is particularly for you since the old version of `miovision_api.volumes_daily` is referenced by `monitoring.vi_daily_totals_miovision`. This PR was meant to overhaul that table based on comments from you and Zahra. Are you OK with dropping that Mat view, or do you want this proposal to exist side by side with the old `volumes_daily` under a new name?

Nate-Wessel:
> @Nate-Wessel this is particularly for you since the old version of `miovision_api.volumes_daily` is referenced by `monitoring.vi_daily_totals_miovision`. This PR was meant to overhaul that table based on comments from you and Zahra. Are you OK with dropping that Mat view, or do you want this proposal to exist side by side with the old `volumes_daily` under a new name?

It looks like the main change is that it went from a wide to a long format with the classifications? Is that right? Is so I'd be happy to rewrite that matview to use the new table... but it would also be great if I could have them both exist for a day or two while I do the switch. Otherwise a bunch of things will CASCADE out of existence and need to be recreated.

gabrielwol:
> It looks like the main change is that it went from a wide to a long format with the classifications? Is that right? Is so I'd be happy to rewrite that matview to use the new table... but it would also be great if I could have them both exist for a day or two while I do the switch. Otherwise a bunch of things will CASCADE out of existence and need to be recreated.

@Nate-Wessel sounds good - you can alter your mat view to use `gwolofs.volumes_daily`. It currently has values for 2023-2024 and I'll insert 2019-2022 over the next few hours.

Nate-Wessel:
> @Nate-Wessel sounds good - you can alter your mat view to use `gwolofs.volumes_daily`. It currently has values for 2023-2024 and I'll insert 2019-2022 over the next few hours.

Done! 

I'm assuming I'll need to switch this to `miovision_api.volumes_daily` at some point when this changes schemas. Just let me know when. 

For reference changes are here: https://github.com/Toronto-Big-Data-Innovation-Team/monitoring/pull/57

volumes/miovision/sql/function/function-aggregate_volumes_daily.sql
```diff
@@ -0,0 +1,126 @@
...
+        FROM miovision_api.volumes AS v
+        LEFT JOIN ref.holiday AS hol ON hol.dt = v.datetime_bin::date
+        --identify duration of unacceptable gaps (extended zero periods)
+        --and avg historical volumes by classification for those gaps
+        LEFT JOIN unacceptable_gaps_summarized AS un ON
+            un.intersection_uid = v.intersection_uid
+            AND v.datetime_bin::date = un.gap_date
+            AND un.classification_uid = v.classification_uid
+        --anti join anomalous_ranges table
+        LEFT JOIN miovision_api.anomalous_ranges AS ar ON
```
chmnata, line 79: so technically everytime we update anomalous range we will have to update this daily table? 🤔

gabrielwol: The only solution I can think of to that is have a view on top of this table that filters them out. 
Or... a function/trigger to use when inserting to anomalous ranges? :/

volumes/miovision/sql/table/create-table-volumes_daily.sql
```diff
@@ -1,44 +1,44 @@
...
+IS '''Daily volumes by intersection_uid, classification_uid.
+Excludes `anomalous_ranges` (use discouraged based on investigations)
+but does not exclude time around `unacceptable_gaps` (zero volume periods).''';
+
+COMMENT ON COLUMN miovision_api.volumes_daily.isodow
+IS 'Use `WHERE isodow <= 5 AND holiday is False` for non-holiday weekdays.';
+
+COMMENT ON COLUMN miovision_api.volumes_daily.unacceptable_gap_minutes
+IS 'Periods of consecutive zero volumes deemed unacceptable
+based on avg intersection volume in that hour.';
```
chmnata, line 37: hmmmm in which hour?

volumes/miovision/sql/function/function-aggregate_volumes_daily.sql
radumas: I would like it to be something that a user can't easily ignore, and not put it on them to be joining

chmnata: having a view on top of this table like gabe suggested sounds easier and less error prone. I think we should probably start announcing newly identified anomalous range and the potential changed on the view?

Like a dag task that looks at newly added anomalous ranges and :Cat_yell: in an alert message tagging daily table users?

gabrielwol: Two workable options: 
1. new view which takes into account `anomalous_ranges` table
2. I could add an optional intersection parameter to the daily agg function and then using an on insert / update / delete trigger on the `anomalous_ranges` table we could re-run the daily aggregation for any affected intersections. 
3. ~~Let the user exclude a_r~~

gabrielwol: How do you feel about table `volumes_daily_donotuse` and view `volumes_daily`, plus some comments?

Nate-Wessel: How about `volumes_daily` vs `volumes_daily_unfiltered`? I think there could be valid reasons for wanting to bypass the `anomalous_ranges` filter, such as for QA.

chmnata: _unfiltered sounds good. As long as we restrict bdit_humans select access to it + give investigators select access I think it should be good :meow_salute:

volumes/miovision/sql/table/create-table-volumes_daily.sql
gabrielwol: Made comment more specific

chmnata (approved):

radumas (approved):

PR #785, opened 2023-12-01
Here dag fix

chmnata:
-  make pull_here.py work
- tested in airflow, it actually works this time :cry_cat:

## Issue(s) this solves:

- #

- just updated the date formatting 

git pull in master

## Notes to future self
- During the attempt to make this also taskflowy, ran into some errors with externelsensors because we now have more tasks to sense than just one bashoperator before. Fixing this will need changes in all dependencies DAGs aka more work and out of scope for this PR. So I just did the minimal changing and make this work for meow.


gabrielwol:
Looks good. See you next time taskflow 👋

Issue #802, opened 2023-12-20
VDS Pipeline - remove top level connections using taskflow API


gabrielwol:
Update `vds_pull_vdsdata.py` and `vds_pull_vdsvehicledata.py` to use Taskflow API and remove the top level Airflow connections.

PR #804, opened 2023-12-20
802 vds pipelines taskflow api

gabrielwol:
A small PR to: 
- Update two VDS DAGs to use taskflow API in order to remove top level connections
- Add doc_md by extracting the relevant section from github readme.
- Add tags. 
- Updates readme with recent DAG updates + refresh DAG graph images. 

## Issue(s) this solves:
- #802 

- Nothing, these DAGs were linked to my home folder for testing and have run as expected the last two days. 

- Unlink the DAGs from Gabe's home folder.


dags/dag_functions.py
```diff
@@ -129,3 +129,10 @@ def task_fail_slack_alert(
         proxy=proxy,
     )
     return failed_alert.execute(context=context)
+
+def get_readme_docmd(readme_path, dag_name):
+    import re
+    contents = open(readme_path, 'r').read()
+    doc_md_key = '<!-- ' + dag_name + '_doc_md -->'
```
tahaislam, line 136: Can you update the [Contributing.md](https://github.com/CityofToronto/bdit_data-sources/blob/master/Contributing.md) file to instruct all of us to do this while writing READMEs for DAGs

dags/dag_functions.py
```diff
@@ -129,3 +129,10 @@ def task_fail_slack_alert(
         proxy=proxy,
     )
     return failed_alert.execute(context=context)
+
+def get_readme_docmd(readme_path, dag_name):
+    import re
```
tahaislam, line 134: This is debatable. This makes DAG loading faster, but not too organized or in a standard way. I'm not sure actually what's the best way to do it 🤔

dags/vds_pull_vdsdata.py
```diff
@@ -48,43 +37,52 @@
     'catchup': True,
 }
```
tahaislam, line 38: I think it might be beneficial to start packing this block of code (updating Python path, setting DAG's default args, getting the doc_md, etc.) which is shared in almost all DAGs in one initialization function for DAGs. and maybe also add it to the Contributing.md file

What do you think? Maybe a new issue and PR for this

dags/vds_pull_vdsdata.py
```diff
@@ -48,43 +37,52 @@
...
-        )
+@dag(
+    dag_id = DAG_NAME,
+    default_args=default_args,
+    max_active_runs=1,
+    template_searchpath=[
+        os.path.join(repo_path,'volumes/vds/sql'),
+        os.path.join(repo_path,'dags/sql')
+    ],
+    doc_md=DOC_MD,
```
tahaislam, line 48: This could be one of the default args for all DAGs

tahaislam (approved):
Looks good. We can merge it, if the DAGs run without any issues

I also left a few comments (not all of them need to be addressed here)

dags/vds_pull_vdsdata.py
gabrielwol: Created a new issue for this!

dags/dag_functions.py
gabrielwol: Added to #806

dags/dag_functions.py
gabrielwol: I added a detailed comment to make this more clear. I think it will make our DAGs more readable.

tahaislam: sorry I wasn't clear about this. I meant having import statements inside functions

gabrielwol: Ohhhh ok, I'll move it to the top.

Issue #812, opened 2024-01-02
End of Year Dag did not run


chmnata:
eoy_table_dag creates a bunch of yearly partition tables, including bt, here, and congestion. 
It failed this year cause it was not scheduled. 
Next execution date showed by running cli `airflow dags next-execution eoy_table_create` shows the next execution date is  2023-12-01 which is in the past.  

Also should probably change over to taskflow + it is still reading paths in `etc`.

chmnata:
I think execution date of 2023-12-01 is correct, as the actual run date would be execution date + 1 year. The below screenshot shows the logical date for next run is 2024-12-01, which is correct.

![image](https://github.com/CityofToronto/bdit_data-sources/assets/46324452/864f6c6e-4d25-466f-9108-fdaf3a8eeb52)

PR #818, opened 2024-01-02
812 eoy table

chmnata:

- taskflowify-ish the dag (definitely more things can be done on the bluetooth side, but can wait after declarative partitioning switch)
- update path from etc to data
- make it run again (?)

## Issue(s) this solves:

- #812 

- Theoretically the start time, interval time, date interval end are all correct so it should run at 2024-12-01. 
- reconfirm the interval time 🤔 

git pull to data_script


dags/eoy_create_tables.py
```diff
@@ -83,64 +55,67 @@ def congestion_create_table(dt):
...
+    schedule='0 0 1 12 *', # At 00:00 on 1st in December.
+    tags=["partition_create"],
+    catchup=False,
+    doc_md=__doc__
+    ) 
+def eoy_create_table_dag():
+    @task_group
+    def yearly_task():
+        """Task group to create yearly tables and triggers."""
+        bt_replace_trigger = PythonOperator(
```
gabrielwol, line 70: I think you forgot to taskflow-ify this one! 🤔

gabrielwol:
As discussed, [here](https://github.com/CityofToronto/bdit_data-sources/blob/545739c6b0fbd2e4bd482e1f6d68f2cea40a27ee/dags/pull_wys.py#L82C1-L102) is the template I've used within the ETL DAGs which uses [generic short circuit operators](https://github.com/CityofToronto/bdit_data-sources/blob/545739c6b0fbd2e4bd482e1f6d68f2cea40a27ee/dags/common_tasks.py#L103C1-L117) short circuit operators to check the date.
In the future we could wrap this all inside a generic TaskGroup which Islam has an [example](https://github.com/Toronto-Big-Data-Innovation-Team/vfh_data_mgmt/blob/2f848e31b2b28bfac5dc8c8007dabcc661100c64/custom_operators/task_group.py#L34) of in VFH!

dags/eoy_create_tables.py
```diff
@@ -1,79 +1,51 @@
...
-    task_msg = """All End of Year tables have been successfully created, 
-        {slack_name} please checkout the tables on the database and make sure 
-        they have been properly created. """.format(slack_name = list_names,)   
+from datetime import datetime, timedelta
+import logging
+LOGGER = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
 
-    return task_msg
+bt_bot = PostgresHook('bt_bot')
```
gabrielwol, line 28: This can go inside `bt_replace_trigger` once you taskflow that one.

gabrielwol:
I think the new @task.run_if decorator in 2.10 will be helpful for simplifying EOY/EOM tasks. Something like
`@task.run_if(lambda context: check_jan_1st(context["ds"]))`
https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html#conditionally-skipping-tasks

gabrielwol:
See #1071 for a cool `pre_execute` syntax for running tasks only on certain days!

gabrielwol (approved):
Implemented my comments around task-flowing remaining tasks. 
No need to change the paradigm around how we run yearly tasks rn, let's get this working before Dec 1!

Issue #820, opened 2024-01-03
Organize Miovision SQLs into folders


gabrielwol:
PR #760 is trying to do too much, including reorganizing SQL files into folders. Creating a separate branch which only moves the SQL files into folders should help tame that PR.

PR #821, opened 2024-01-03
#820 organize files into folders

gabrielwol:
- reorganizes Miovision sql files into folders
- fix references to sql files within the 3 miovision readmes
- deletes ` miovision_eoy_create_tables.py` which is deprecated by #735 
  
## Issue(s) this solves:

- #820 

- nothing


chmnata:
Thanks! This looks a lot more organized :meow_wow:. For   `miovision_eoy_create_tables.py` that is still in `sql/` I'm wondering if we should just delete it or move it to an archived folder or sth. Also I took a look at the list of tables you moved, I think none of them are referenced in a dag/automated process so changing file path should have no effect, can you confirm this?

gabrielwol:
I confirmed no sql scripts are referenced in `pull_miovision.py` and `intersection_tmc.py`. It looks like the one place is the readmes which I'll change now. Good catch!
Good to delete `miovision_eoy_create_tables.py`. Don't think we'll ever go back to using triggers like that!

chmnata (approved):
Looks good! Merging :meow_merge:

PR #823, opened 2024-01-05
#630 detect zero miovision volume counts automagically

gabrielwol:
Replaces #777 which was abandoned due to mega-merge conflict. 

- Automatically add/appends to `anomalous_ranges` table zero volume days for classifications 1,2,6,10. Includes special case for when all these classifications are zero (classification_uid becomes NULL as per existing convention). 
  - **Major change from original PR:** The new gaps are now automatically appended to existing gaps.
- Adds a new task to `miovision_pull` DAG to run this aggregation daily. 
- Add missing `miovision_api.anomaly_problem_levels`, `miovision_api.anomaly_investigation_levels` table definitions to data-sources github (unchanged).  
- Updates readme to reflect changes to anomalous_ranges table.

## Issue(s) this solves:
- A start to tackling #630

- New gaps are automatically appended to existing. Chose not to deal with rarer situation when a new gap should be prepended to an existing gap because of potential for recursive situation.  

- [x] ~~**Note:** #763 should be merged first (consistent definition of a day)~~
- [ ] Add new sql function to miovision_api. 
- [ ] Update definition of `miovision_api.anomalous_ranges`:
```sql
ALTER TABLE miovision_api.anomalous_ranges
ADD COLUMN range_start timestamp DEFAULT null,
ADD COLUMN range_end timestamp DEFAULT null;

UPDATE miovision_api.anomalous_ranges
SET
    range_start = 
        CASE
            --if not inclusive, move forward to make inclusive
            WHEN lower_inc(time_range) IS FALSE
                THEN LOWER(time_range) + interval '1 day'
            ELSE LOWER(time_range)
        END, 
    range_end =
        CASE
            --if inclusive, move forward to make non-inclusive
            WHEN upper_inc(time_range) IS TRUE
                THEN UPPER(time_range) + interval '1 day'
            ELSE UPPER(time_range)
        END;

ALTER TABLE miovision_api.anomalous_ranges
ADD CONSTRAINT anomalous_ranges_range_order
CHECK (
    --results for check constraints of true or null are allowed
    range_start < range_end
);

COMMENT ON TABLE miovision_api.anomalous_ranges
IS 'Table to store anomalous ranges of Miovision data. 
Values are added to this table both manually by miovision_admins/miovision_data_detectives and
automatically through a daily Airflow process that runs function miovision_api.identify_zero_counts(date).';

COMMENT ON COLUMN miovision_api.anomalous_ranges.intersection_uid
IS 'NULLs are allowed in the column intersection_uid to denote
anomalous that affect all intersections.';

COMMENT ON COLUMN miovision_api.anomalous_ranges.classification_uid
IS 'NULLs are allowed in the column classification_uid to denote
anomalous that affect all classifications.';

COMMENT ON COLUMN miovision_api.anomalous_ranges.range_start
IS 'Start of anomalous range. Inclusive. NULL values represent an unbounded range_start.';

COMMENT ON COLUMN miovision_api.anomalous_ranges.range_end
IS 'End of anomalous range. Exclusive. NULL values represent an unbounded range_end.';

--drop and recreate the constraint for range_start/range_end instead of time_range.
ALTER TABLE miovision_api.anomalous_ranges
DROP CONSTRAINT miovision_qc_intersection_uid_classification_uid_time_range_key;

ALTER TABLE miovision_api.anomalous_ranges
ADD CONSTRAINT miovision_qc_intersection_uid_classification_uid_time_range_key
UNIQUE NULLS NOT DISTINCT (intersection_uid, classification_uid, range_start, range_end);

--drop deprecated time_range column
--we will need to get Nate to alter dependent `monitoring.vi_daily_totals_miovision` before we can drop this column. 
ALTER TABLE miovision_api.anomalous_ranges
DROP COLUMN time_range;
```


volumes/miovision/api/intersection_tmc.py
```diff
@@ -408,7 +409,22 @@ def get_report_dates(conn, time_period, intersections):
...
+    
+def agg_zero_volume_anomalous_ranges(conn, time_period):
+    """Aggregate into miovision_api.anomalous_ranges.
+    Data is cleared from volumes_daily prior to insert.
+    """
+    try:
+        with conn:
+            with conn.cursor() as cur:
+                #this function includes a delete query preceeding the insert.
+                anomalous_range_sql="SELECT miovision_api.identify_zero_counts(%s::timestamp, %s::timestamp)"
```
gabrielwol, line 425: Need to modify this to run multiple days.

chmnata:
1) On approach: 

As discussed with @radumas, we should add a column in the anomalous_ranges for approach. 

Currently the table has:
- intersection_uid
- classification_uid
- time_Range
- notes
- uid
- investigation_level
- problem_level

2) On intersection_uid and classification_uid having NULLs 
Both intersection_uid and classification_uid could be `NULL` at the moment (NULL = all)  but I think we should be a little more explicit on defining that. We could change them to int array and include all intersection_uid/classification_uid in the array, as it would be easier to use `<@` to filter out data after. Since we are also adding new intersections in the future, using `NULL` could be confusing as these anomalous ranges would not apply to any new intersections.

gabrielwol:
I think both of these points warrant some more discussion. Do you think they need to resolved in this PR or could they be new issues?
1. Open to it, but do our investigations ever get to that level of detail? Are we aware of cases where one approach is labeled suspect but we accept the others? 
2. 
> Since we are also adding new intersections in the future, using NULL could be confusing as these anomalous ranges would not apply to any new intersections.  

But they also don't NOT apply to new intersections... Like if we add new intersections today and backfill them, I would assume any ongoing issues would also apply to the backfilled data, but we would have to go back and update the anomalous_range.

chmnata:
> I think both of these points warrant some more discussion. Do you think they need to resolved in this PR or could they be new issues?
> 
> 1. Open to it, but do our investigations ever get to that level of detail? Are we aware of cases where one approach is labeled suspect but we accept the others?
> 
> > Since we are also adding new intersections in the future, using NULL could be confusing as these anomalous ranges would not apply to any new intersections.
> 
> But they also don't NOT apply to new intersections... Like if we add new intersections today and backfill them, I would assume any ongoing issues would also apply to the backfilled data, but we would have to go back and update the anomalous_range.

Yea you are right, this anomalous ranges table should be its own issue. We can update this function to accommodate any changes afterwards.

For (2), I think its more like we won't have data for some anomalous ranges entries as they were not installed at the time, so the issue doesn't apply to them. This might impact our ability to do analysis on how reliable each camera is later down the road. But I feel like if they weren't installed they shouldn't be included in anomalous ranges :meow_dio: . 

There are only 2 ongoing issues for "all" intersections in the table, which are when streetcars were included in classification_uid (2 and 9) . Now that I think about it, this is not very `anomalous_range` fitting, but more of a `discontinuities`, similar to ecocounter's discontinuities table Nate created.

radumas:
Aye I think model changes are likely going to have to be their own table, not sure how to easily make them work with anomalous ranges, see https://github.com/CityofToronto/bdit_data-sources/issues/846

gabrielwol:
While QCing the new intersection I found; there are 5 intersection with complete outages and another 41 intersection_uid / classification_uid (6, 10) combos with no data right now 😵 This new task is going to be very helpful...

chmnata:
Can you pull master to this when you have time? We can test tmr

chmnata:
@Nate-Wessel can you alter dependent `monitoring.vi_daily_totals_miovision` ? We are planning to implement this change and would need to drop `time_range` column in `miovision_api.anomalous_ranges`

Nate-Wessel:
> @Nate-Wessel can you alter dependent `monitoring.vi_daily_totals_miovision` ? We are planning to implement this change and would need to drop `time_range` column in `miovision_api.anomalous_ranges`

Happy to. Though... should I still filter dates with anomalous ranges at all, or just trust whatever view to do that for me? If so, where is that view?

Also, I'm currently filtering out days where there is any temporal overlap between the day and an anomalous range where
```sql
anomalous_ranges.problem_level IN ('do-not-use', 'questionable')
```
Is this the same criteria being discussed/implemented here?

gabrielwol:
(1) hmmm. I see you are selecting from `FROM miovision_api.volumes_daily`. The new view with that name will exclude anomalous_ranges, but only after #783 is merged.

(2) Yes - 'do-not-use' label will be used on the auto flagged anomalous_ranges.

Nate-Wessel:
I'm actually pointing it at `gwolofs.volumes_daily` right now. The current version of that matview is here: https://github.com/Toronto-Big-Data-Innovation-Team/monitoring/pull/57

I just did a very minimal update to drop the dependency on `time_range`. Using `range_start` and `range_end` to do the same thing. 

I think I'll come back and clean things up more once this gets merged and the dust settles.

volumes/miovision/sql/function/function-identify-zero-counts.sql
```diff
@@ -0,0 +1,147 @@
...
+        AND range_start <> start_date
+        AND range_end = start_date + interval '1 day';
+    
+    RAISE INFO '% records in anomalous_ranges extended to include %.', n_updated, start_date;
+
+END;
+
+$BODY$;
+
+ALTER FUNCTION miovision_api.identify_zero_counts(timestamp, timestamp, interval)
```
chmnata, line 140: have to change to just date

volumes/miovision/sql/table/create-table-anomaly_investigation_levels.sql
```diff
@@ -0,0 +1,30 @@
...
+GRANT SELECT ON TABLE miovision_api.anomaly_investigation_levels TO bdit_humans;
+
+GRANT ALL ON TABLE miovision_api.anomaly_investigation_levels TO miovision_admins;
+
+GRANT REFERENCES ON TABLE miovision_api.anomaly_investigation_levels TO miovision_data_detectives;
+
+GRANT ALL ON TABLE miovision_api.anomaly_investigation_levels TO scannon;
+
+COMMENT ON TABLE miovision_api.anomaly_investigation_levels
+IS 'Indicates the furthest degree to which the movision QA issue has been investigated.
```
chmnata, line 29: typo miovision

volumes/miovision/sql/table/create-table-anomaly_problem_levels.sql
```diff
@@ -0,0 +1,29 @@
...
+REVOKE ALL ON TABLE miovision_api.anomaly_problem_levels FROM bdit_humans;
+REVOKE ALL ON TABLE miovision_api.anomaly_problem_levels FROM miovision_data_detectives;
+
+GRANT SELECT ON TABLE miovision_api.anomaly_problem_levels TO bdit_humans;
+
+GRANT ALL ON TABLE miovision_api.anomaly_problem_levels TO miovision_admins;
+
+GRANT REFERENCES ON TABLE miovision_api.anomaly_problem_levels TO miovision_data_detectives;
+
+GRANT ALL ON TABLE miovision_api.anomaly_problem_levels TO scannon;
```
chmnata, line 26: change scannon to miovision_data_detectives

volumes/miovision/sql/table/create-table-anomaly_investigation_levels.sql
```diff
@@ -0,0 +1,30 @@
...
+    CONSTRAINT qa_levels_pkey PRIMARY KEY (uid)
+)
+
+TABLESPACE pg_default;
+
+ALTER TABLE IF EXISTS miovision_api.anomaly_investigation_levels
+OWNER TO miovision_admins;
+
+REVOKE ALL ON TABLE miovision_api.anomaly_investigation_levels FROM bdit_humans;
+REVOKE ALL ON TABLE miovision_api.anomaly_investigation_levels FROM miovision_data_detectives;
```
chmnata, line 18: I'm guessing detectives should have access to this ?

volumes/miovision/sql/table/create_anomalous_ranges.sql
```diff
@@ -1,37 +1,84 @@
...
\ No newline at end of file
+REVOKE ALL ON TABLE miovision_api.anomalous_ranges FROM bdit_humans;
+
+GRANT SELECT ON TABLE miovision_api.anomalous_ranges TO bdit_humans;
+
+GRANT ALL ON TABLE miovision_api.anomalous_ranges TO miovision_admins;
+
+GRANT ALL ON TABLE miovision_api.anomalous_ranges TO miovision_data_detectives;
+
+GRANT ALL ON TABLE miovision_api.anomalous_ranges TO nwessel;
```
chmnata, line 45: nwessel is a part of detectives, 🔪 dis

volumes/miovision/sql/table/create_anomalous_ranges.sql
```diff
@@ -1,37 +1,84 @@
...
+    ON DELETE NO ACTION,
+    CONSTRAINT miovision_qc_problem_level_fkey FOREIGN KEY (problem_level)
+    REFERENCES miovision_api.anomaly_problem_levels (uid) MATCH SIMPLE
+    ON UPDATE NO ACTION
+    ON DELETE NO ACTION,
+    CONSTRAINT miovision_qc_qa_level_fkey FOREIGN KEY (investigation_level)
+    REFERENCES miovision_api.anomaly_investigation_levels (uid) MATCH SIMPLE
+    ON UPDATE NO ACTION
+    ON DELETE NO ACTION,
+    CONSTRAINT anomalous_ranges_range_order CHECK (
```
chmnata, line 29: can we add a `miovision` in front of the constraint name? We will have to add the same checks to other data sources later.

volumes/miovision/sql/table/create_anomalous_ranges.sql
```diff
@@ -1,37 +1,84 @@
...
+
+-- Trigger: audit_trigger_row
+
+-- DROP TRIGGER IF EXISTS audit_trigger_row ON miovision_api.anomalous_ranges;
+
+CREATE OR REPLACE TRIGGER audit_trigger_row
+AFTER INSERT OR DELETE OR UPDATE
+ON miovision_api.anomalous_ranges
+FOR EACH ROW
+EXECUTE FUNCTION miovision_api.if_modified_func('true');
```
chmnata, line 55: what is this if_modified_func? I don't see it in miovision_api

chmnata: NEVERMIND I FOND IT

volumes/miovision/sql/function/function-identify-zero-counts.sql
```diff
@@ -0,0 +1,154 @@
...
+        COALESCE(existing.intersection_uid, 0) = COALESCE(new_gaps.intersection_uid, 0)
+        AND COALESCE(existing.classification_uid, 0) = COALESCE(new_gaps.classification_uid, 0)
+        AND (
+                existing.range_start <= new_gaps.range_start
+                AND existing.end >= new_gaps.range_end)
+            OR (
+                existing.range_end IS NULL
+                AND existing.range_start <= new_gaps.range_start
+            )
+        )
```
chmnata, line 115: extra ) here

volumes/miovision/sql/function/function-identify-zero-counts.sql
```diff
@@ -0,0 +1,154 @@
...
+        COALESCE(new_gaps.intersection_uid, 0) = COALESCE(updated_values.intersection_uid, 0)
+        AND COALESCE(new_gaps.classification_uid, 0) = COALESCE(updated_values.classification_uid, 0)
+        AND new_gaps.range_start = updated_values.range_start
+    --anti join existing open ended/overlapping gaps
+    LEFT JOIN miovision_api.anomalous_ranges AS existing ON
+        COALESCE(existing.intersection_uid, 0) = COALESCE(new_gaps.intersection_uid, 0)
+        AND COALESCE(existing.classification_uid, 0) = COALESCE(new_gaps.classification_uid, 0)
+        AND (
+                existing.range_start <= new_gaps.range_start
+                AND existing.end >= new_gaps.range_end)
```
chmnata, line 110: range_end instead of end

chmnata (approved):
Thanks @gabrielwol!🌟  There is magic in this world. 🌟

PR #828, opened 2024-01-12
I810 move match

### LINKED ISSUE #810: Generalized MOVE Summary Stats Transform-nation
We did a request that required:
- peak hour counts from TMCs and speed volumes
- mean, 15th, 50th, 85th and 95th percentile speeds
- Daily traffic volumes (from speed volumes)

Now can we generalize this to make future requests (and work) easier?

I was thinking a `VIEW`, but it might be more appropriate to be calculating summary stats on a function basis, which should maybe go into the MOVE count loading and then get replicated over to us. 

cc: @peter-lyons @mkewins

scann0n:

- Closes #810
- Provides sqls for calculating peak hour tmc and speedvol counts, plus speedvol mean speed, daily counts as well as 15th, 50th, 85th and 95th percentile speeds

## Issue(s) this solves:

- #810 

- This code was reviewed as part of a request and was lightly edited to remove the request-specific parts.

- Nothing - at this point we are saving it here for reference

### DIFF
--- a/volumes/short_term_counting_program/move_match/speedvol_stats.sql
+++ b/volumes/short_term_counting_program/move_match/speedvol_stats.sql
@@ -65,7 +65,7 @@ In this CTE, two WINDOWs are used:
     cumulative: creates windows based on count_info_id and speed bin
     _all: creates windows based on count_info_id
 Here is an explanation of the first SUM(SUM()) line:
-    Part 1 - SUM(SUM(spd_vol.volume_15min)) OVER cumulative: 
+    Part 1 - SUM(SUM(spd_vol.volume_15min)) OVER cumulative:
         The inside SUM() adds up 15 minute volume data (resulting in daily volumes by speed bin).
         The outside SUM() adds up the bins over the cumulative window.
         In plain language: sum daily totals by speed bin, then calculate a cumulative sum for
@@ -79,8 +79,8 @@ Here is an explanation of the first SUM(SUM()) line:
 Here is an explanation of the second SUM(SUM()) line:
     SUM(spd_vol.volume_15min) / (SUM(SUM(spd_vol.volume_15min)) OVER _all + 0.00000001) AS pct:
     Calculate daily volumes by speed bin and divide it by...
-    Daily total volumes (15 minute volume data are aggregated to speed bins in the inside sum, then they
-    are totalled in the outside sum).
+    Daily total volumes (15 minute volume data are aggregated to speed bins in the inside sum, then
+    they are totalled in the outside sum).
     You end up with the proportion of daily volume in each speed bin.
 */
 
@@ -99,12 +99,12 @@ per_15 AS (
             - spd_bin_vol.cum_pct -- the cumulative % for that bin
             + spd_bin_vol.pct -- munis the non-cumulative % from that bin
         )
-            -- all dividded by the bin volume % plus that decimal to avoid a divide by zero error
+        -- all dividded by the bin volume % plus that decimal to avoid a divide by zero error
         / spd_bin_vol.pct AS pctile_speed_15
     FROM spd_bin_vol
     WHERE
         spd_bin_vol.cum_pct >= 0.15
-        AND pct <> 0
+        AND spd_bin_vol.pct <> 0
     ORDER BY
         spd_bin_vol.count_info_id,
         spd_bin_vol.cum_pct
@@ -126,7 +126,7 @@ per_50 AS (
     FROM spd_bin_vol
     WHERE
         spd_bin_vol.cum_pct >= 0.50
-        AND pct <> 0
+        AND spd_bin_vol.pct <> 0
     ORDER BY
         spd_bin_vol.count_info_id,
         spd_bin_vol.cum_pct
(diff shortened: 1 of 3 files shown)


volumes/short_term_counting_program/move_match/speedvol_stats.sql
```diff
@@ -0,0 +1,251 @@
...
+Here is an explanation of the second SUM(SUM()) line:
+    SUM(spd_vol.volume_15min) / (SUM(SUM(spd_vol.volume_15min)) OVER _all + 0.00000001) AS pct:
+    Calculate daily volumes by speed bin and divide it by...
+    Daily total volumes (15 minute volume data are aggregated to speed bins in the inside sum, then they are totalled in the outside sum).
+    You end up with the proportion of daily volume in each speed bin.
+*/
+
+-- Calculate 15th percentile speed
+per_15 AS (
+    SELECT DISTINCT ON (spd_bin_vol.count_info_id)
```
gabrielwol, line 73: You need to pair an ORDER BY with DISTINCT ON to make sure this always returns the correct results even if the source data is not sorted. See: https://www.postgresql.org/docs/current/sql-select.html

> The DISTINCT ON expressions are interpreted using the same rules as for ORDER BY (see above). Note that the “first row” of each set is unpredictable unless ORDER BY is used to ensure that the desired row appears first.

volumes/short_term_counting_program/move_match/speedvol_stats.sql
```diff
@@ -0,0 +1,251 @@
...
+        (
+            LOWER(spd_bin_vol.speed_kph) -- lower speed bound of bin
+            + (UPPER(spd_bin_vol.speed_kph) - LOWER(spd_bin_vol.speed_kph))
+            * (
+                0.15 -- multiplied by the desired percentile (for which you want speed)
+                - ( -- minus
+                    spd_bin_vol.cum_pct -- the cumulative % for that bin
+                    - spd_bin_vol.pct -- munis the non-cumulative % from that bin
+                )
+                ) / (spd_bin_vol.pct + 0.00000001 -- all dividded by the bin volume % plus that decimal to avoid a divide by zero error
```
gabrielwol, line 88: Filtering out zeros or using COALESCE would be better than adding to the denominator, which impacts the result ever so slightly.

volumes/short_term_counting_program/move_match/speedvol_stats.sql
```diff
@@ -0,0 +1,251 @@
...
+am_peak AS (
+    SELECT DISTINCT
+        spd_bin_vol.count_info_id,
+        spd_bin_vol.arterycode,
+        spd_bin_vol.dt,
+        FIRST_VALUE(hr_sums.hr) OVER (PARTITION BY spd_bin_vol.count_info_id ORDER BY hr_sums.hr_vol DESC) AS am_peak_hr,
+        FIRST_VALUE(hr_sums.hr_vol) OVER (PARTITION BY spd_bin_vol.count_info_id ORDER BY hr_sums.hr_vol DESC) AS am_peak_vol
+    FROM spd_bin_vol
+    LEFT JOIN hr_sums USING (count_info_id)
+    WHERE hr_sums.hr::time < '12:00'
```
gabrielwol, line 207: date_part('hour', timestamp) returns in integer which is easier for comparisons

volumes/short_term_counting_program/move_match/speedvol_stats.sql
```diff
@@ -0,0 +1,251 @@
...
+        SUM(spd_vol.volume_15min) / (SUM(SUM(spd_vol.volume_15min)) OVER _all + 0.00000001) AS pct -- SUM(SUM()) explained below
+    FROM spd_vol
+    GROUP BY 
+        spd_vol.arterycode,
+        spd_vol.count_info_id,
+        spd_vol.direction,
+        spd_vol.location,
+        spd_vol.datetime_bin::date,
+        spd_vol.speed_kph
+    WINDOW
```
gabrielwol, line 42: Good use of window functions

gabrielwol (changes requested):
Need to change the distinct on issue, and the 0.00000001 issue. 

Also may need some refactoring re: Raph's issue, but I need some clarification on that. 
> I was thinking a VIEW, but it might be more appropriate to be calculating summary stats on a function basis, which should maybe go into the MOVE count loading and then get replicated over to us.

volumes/short_term_counting_program/move_match/speedvol_stats.sql
```diff
@@ -0,0 +1,251 @@
+-- This code replicates statistics from the speed volume Speed Percentile Report: https://move.intra.prod-toronto.ca/view/location/s1:Ab_DAA/POINTS/reports/ATR_SPEED_VOLUME
+-- Note that am and pm peak hours as presented in this report are based on clock face hours (not 15 minute bins)
```
radumas, line 2: This should be changed to be consistent across count types (using a rolling hour window instead of hourly aggregation)

volumes/short_term_counting_program/move_match/speedvol_stats.sql
gabrielwol: Accomplished by excluding bins with pct = 0. These could not include the desired percentile anyway because they have the same upper bound as the previous bin which would be chosen instead.

volumes/short_term_counting_program/move_match/speedvol_stats.sql
gabrielwol: Resolved using
```
ORDER BY
    spd_bin_vol.count_info_id,
    spd_bin_vol.cum_pct
```

gabrielwol (approved):

Issue #827, opened 2024-01-12
Update miovision_api.aggregate_15_min_mvt to take optional intersection array


gabrielwol:
We can replace the use of two distinct functions `miovision_api.aggregate_15_min_mvt` + `miovision_api.aggregate_15_min_mvt_single_intersection` with a single updated function that takes an optional intersection array, similar to the method used for `miovision_api.aggregate_15_min(date, date, integer[])` implemented in #822. 
Since we now clear `volumes_15min_mvt` prior to any insert, the anti join should also no longer be necessary, speeding up this function.

PR #822, opened 2024-01-04
746 refactor miovision dag

gabrielwol:
This PR replaces #760. 

## What this pull request accomplishes:
- New taskflow refactored Miovision DAG as `miovision_pull` 
  - Keeps old `pull_miovision` with end date of 2024-01-15 (will need to adjust)
- Updates to `intersection_tmc.py`:
  - Allows the aggregations to be called individually so they can be separated out into Airflow tasks
  - clear each of the miovision tables prior to inserting
    - `miovision_api.volumes`, `miovision_api.api_log` and `miovision_api.volumes_15min_mvt` have intersection parameter to clear individual / all intersections. 
    - Other tables can be cleared and re-run in their entirety even if just a single intersection is pulled, so I used more simple DELETE queries directly in `intersection_tmc.py` 
- Improved the single intersection option; 
  - Added an airflow parameter so it can be used in run config

## Issue(s) this solves:
- Closes #746 

## What, in particular, needs to reviewed:
- I chose to clear aggregate tables one at a time before inserting:
  - Potential downside is if someone inadvertently runs the `pull_miovision` task without the downstream aggregations, then there could be a mismatch between the raw `volumes` table and the downstream aggregate tables. Airflow clear "downstream" should take care of this!

## What needs to be done by a sysadmin after this PR is merged
New/modified sql functions to be updated:
- [ ] volumes/miovision/sql/function-clear-api_log.sql
  - [ ] Delete existing
- [ ] volumes/miovision/sql/function-clear-volumes_15min.sql
  - [ ] Delete existing
- [ ] volumes/miovision/sql/function-clear-volumes_15min_mvt.sql
  - [ ] Delete existing
- [ ] volumes/miovision/sql/function-clear-volumes.sql
  - [ ] Delete existing
- [ ] volumes/miovision/sql/function-api_log.sql
  - [ ] Delete existing

- [ ] Adjust the start/end dates of the two DAGs when this PR gets reviewed.

- [x] Assuming #738 gets merged before this one, we will need to manually add the data check tasks to the new `miovision_pull` DAG, as the file name has changed and it won't get merged cleanly. 
  - [ ] Completed on 2024-01-04, but may need to revisit once merged.

PR #829, opened 2024-01-15
827 add intersection param to miovision agg 15 min mvt

gabrielwol:
- A small follow up to the #822: simplifies the two volumes_15min_mvt agg functions into one by adding an optional intersections param as we did for other functions in #822.
- When we added the single intersection function in #669, we weren't running a clear function before each insert so we needed a special case with an anti-join to avoid inserting duplicate padding values. Now that we delete before inserts, we can simplify this.
 
## Issue(s) this solves:
- #827 

We should test this function by trying to aggregate a single intersection in Airflow which was being very slow before. I tested the "SELECT" part of the insert which ran in just 8s and returned the expected row count. 

Delete the two existing functions (`aggregate_15_min_mvt`, `aggregate_15_min_mvt_single_intersection`) and replace with new function definition.


gabrielwol:
Some merge weirdness happening with this one. will try to resolve Monday.

PR #831, opened 2024-01-15
Ecocounter discontinuities table

Nate-Wessel:

- creates a way of tracking changes / discontinuities in the ecocounter data

## Issue(s) this solves:

- discontinuities were introduced when it was discovered that the counts were not great at many sensors -- they subsequently received some adjustments to their sensitivity which will be in place going forward. 

- The table definition

@Nate-Wessel will create the table

### DIFF
--- a/volumes/ecocounter/tables/discontinuities.sql
+++ b/volumes/ecocounter/tables/discontinuities.sql
@@ -1,13 +1,13 @@
 CREATE TABLE ecocounter.discontinuities (
     -- a references to EITHER a site_id or flow_id
-    site_id REFERENCES ecocounter.sites (site_id),
-    flow_id REFERENCES ecocounter.flows (flow_id),
+    site_id numeric REFERENCES ecocounter.sites (site_id),
+    flow_id numeric REFERENCES ecocounter.flows (flow_id),
     -- moment the change takes place
-    break timestamp IS NOT NULL,
+    break timestamp NOT NULL,
     -- approximate bounds if the precise time is not known
     give_or_take interval,
     -- required description of what changed - be verbose!
-    notes text IS NOT NULL
+    notes text NOT NULL
     CHECK (
         -- only one or the other specified but not both
         (site_id IS NOT NULL OR flow_id IS NOT NULL)
--- a/volumes/ecocounter/tables/discontinuities.sql
+++ b/volumes/ecocounter/tables/discontinuities.sql
@@ -1,0 +1,20 @@
+CREATE TABLE ecocounter.discontinuities (
+    -- a references to EITHER a site_id or flow_id
+    site_id REFERENCES (sites.site_id),
+    flow_id REFERENCES (flows.flow_id),
+    -- moment the change takes place
+    break timestamp IS NOT NULL,
+    -- approximate bounds if the precise time is not known
+    give_or_take interval,
+    -- required description of what changed - be verbose!
+    notes text IS NOT NULL
+    CHECK (
+        -- only one or the other specified but not both
+        (site_id IS NOT NULL OR flow_id IS NOT NULL)
+        AND NOT (site_id IS NOT NULL AND flow_id IS NOT NULL)
+    )
+);
+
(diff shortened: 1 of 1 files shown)


Nate-Wessel:
@chmnata I'm not totally sold on this yet, but just putting it out there for conversation / feedback while I think some more about it. Some further questions I have:

* is there any way we can better describe changes other than through a free-form notes field?
* is it too complicated to have two potential external keys? What level are the changes actually implemented at? I feel like it's possible we could tweak one sensor at a site but not the other.  
* do these field names make sense? Are there better synonyms?

Nate-Wessel:
I've created this table and added the data for the seven recent site-level changes to it.

volumes/ecocounter/tables/discontinuities.sql
```diff
@@ -0,0 +1,21 @@
...
+    flow_id numeric REFERENCES ecocounter.flows (flow_id),
+    -- moment the change takes place
+    break timestamp NOT NULL,
+    -- approximate bounds if the precise time is not known
+    give_or_take interval,
+    -- required description of what changed - be verbose!
+    notes text NOT NULL
+    CHECK (
+        -- only one or the other specified but not both
+        (site_id IS NOT NULL OR flow_id IS NOT NULL)
```
chmnata, line 14: The first clause is good enough for the checks, e.g.
```
-- does not return row `uid = 3`
SELECT *
FROM (VALUES 
	  (1, null, 1), 
	  (2, 2, null), 
	  (3, null, null)) as dates(uid, col_a, col_b)
where col_a IS NOT NULL OR col_b IS NOT NULL
```

chmnata: we could add a constraint for the check as well 
```
alter table ecounter.discontinuities
  add constraint discontinuities_site_id_flow_id
    check (site_id is not null or flow_id is not null);		
    ```

chmnata (commented):
Yea im thinking on what other categories we can add to make using this table a little easier 🤔 but can't think of other distinct groupable reasoning that will lead to data discontinuities (and not anomalous ranges) at the movement. 

I think this is a good structure for now, we can circle back in the future if needed. :meow_salute:

volumes/ecocounter/tables/discontinuities.sql
Nate-Wessel: > we could add a constraint for the check as well

As far as I know, that's just a more verbose way of doing the same thing. A `CHECK` in the initial table definition just _is_ a constraint.

Nate-Wessel: > The first clause is good enough for the checks

I wanted to prevent both `site_id` and `flow_id` from being defined at the same time though. One and only one of those should be defined, or else I'm not sure what it would mean.

chmnata: yea I think the only difference is we can name the constraint ourselves and since it will show up as a constraint in the table after creation, might as well include it when creating the table.

chmnata: Maybe we should just make both not null then ? I think was thinking if the discontinuities was discovered at one site_id, it will just cascade to all relevant flow_id within the site. But maybe we should just be very explicit in which flow_ids should be included in an array. This could help if the query only uses flow_id.

Nate-Wessel: Hmmm. I'll give this a little more thought. I can see now how having both options in this table is going to end up causing some confusion. Maybe we just define either the site or the flow level and also have a view that translates to the other level based on that.

Nate-Wessel:
@chmnata This perhaps isn't perfect yet but I'd like to see it merged in, as it at least describes what I'm currently doing. My next step is to work on this general issue through the volumes index and I'd like to be able to reference these changes. 

Can you please approve and merge?

chmnata (approved):

chmnata:
We should chat soon to talk about the structure for both anomalous ranges and discontinuities, since we have to add a `discontinuities` table for miovision as well. 🤔  Would be best to figure out a standardized structure before creating one for miovision

Issue #833, opened 2024-01-17
Bug in `public.summarize_gaps_data_check` where id_col get cast to regtype


gabrielwol:
Example failure [here](https://10.162.65.124/airflow/dags/miovision_check/grid?dag_run_id=scheduled__2024-01-16T09%3A00%3A00%2B00%3A00&task_id=check_gaps&tab=logs) where intersection_uid `27` gets cast to `tid` (regtype). 

Can be traced back to this line which actually passed type `regtype` and not the intended `id_col_dtype` param. 
https://github.com/CityofToronto/bdit_data-sources/blob/27298508344f90c5090851dc5af525cfa8622f7e/dags/sql/create-function-summarize_gaps_data_check.sql#L34

PR #834, opened 2024-01-17
833 regtype bug in publicsummarize gaps data check

gabrielwol:
- removes `id_col_dtype` param from `public.summarize_gaps_data_check`. Replace with always passing `null::text` to `generic_find_gaps` when summarizing.  

## Issue(s) this solves:
- #833 

- Tested in gwolofs schema. Working! 

- DELETE FUNCTION public.summarize_gaps_data_check;
- replace with new function from: dags/sql/create-function-summarize_gaps_data_check.sql
- update `miovision_check` DAG


chmnata (approved):
Tested on airflow, function updated. Thanks @gabrielwol for the speeeeedy fix :meow_salute:

Issue #860, opened 2024-02-06
  Bug in anti join conditions of Miovision `identify_zero_counts


gabrielwol:
Anti join conditions don't account for case when there is an existing `null` classification a_r and a new, non-null a_r to be inserted. This line needs an `OR existing.classification_uid IS NULL` condition:
```sql
AND COALESCE(existing.classification_uid, 0) = COALESCE(new_gaps.classification_uid, 0)
```

> @gabrielwol Actually, these short little blips created for `classification_uid=10` take place during a larger anomaly, (suspect, questionable). 
> 
> These are a different problem-level, `do-not-use`, and investigation level though. I guess I don't see a way that I would have done it much differently in an automated way... Some manual fixing needed here though.
> 
> 

_Originally posted by @Nate-Wessel in https://github.com/Toronto-Big-Data-Innovation-Team/monitoring/pull/60#discussion_r1478904965_

gabrielwol:
Closed by https://github.com/CityofToronto/bdit_data-sources/pull/887

PR #861, opened 2024-02-06
#860 update join conditions for null classification case

gabrielwol:
- updates the anti-join with existing anomalous ranges: when `existing` record has a null classification_uid, we want to exclude any new records from being added for that intersection. 

## Issue(s) this solves:
- #860 

- Join conditions when there are nulls classification_uid's on either side

- update function in database. I don't think we should backfill as Nate has been cleaning these up manually.


gabrielwol:
@chmnata we need to resolve the question of if we want individual classification gaps to overlap with full intersection gaps (null classification). My thought is no, especially since we are soon going to add legs to the mix which ups the potential for many gaps overlapping with an existing manual entry.

Issue #867, opened 2024-02-08
Miovision: Add anomalous_range notes to volumes_daily view


gabrielwol:
This view includes anything not in ['do-not-use'::text, 'questionable'::text]. For these issues we should add the notes as an extra column.

PR #876, opened 2024-02-15
#867 add AR notes to volumes_daily

gabrielwol:
- Adds context from `anomalous_range` table to filtered `volumes_daily` (see sample in the comment below)
- new columns `anomalous_range_caveats` and `anomalous_range_uid` are arrays to support multiple issues per row (last row below)

## Issue(s) this solves:
- #867 

- What do you think of the formatting of the anomalous_ranges and the information contained in the extra columns? We can also consider using `string_agg` instead of `array_agg`. 
- You can see this in action at `gwolofs.volumes_daily`. Already confirmed that:
```sql
SELECT (SELECT COUNT(*) FROM gwolofs.volumes_daily) = (SELECT COUNT(*) FROM miovision_api.volumes_daily)
```

- DROP existing view and replace with new definition.
- After changes are approved I will update the readme.

### DIFF
--- a/volumes/miovision/sql/views/create-view-volumes_daily.sql
+++ b/volumes/miovision/sql/views/create-view-volumes_daily.sql
@@ -10,13 +10,14 @@ CREATE VIEW miovision_api.volumes_daily AS (
         v.unacceptable_gap_minutes,
         v.avg_historical_gap_vol,
         array_agg(ar.notes ORDER BY ar.range_start, ar.uid)
-            AS anomalous_range_caveats,
-        array_agg(ar.uid::text ORDER BY ar.range_start, ar.uid)
-            AS anomalous_range_uid
+        AS anomalous_range_caveats,
+        array_agg(ar.uid ORDER BY ar.range_start, ar.uid)
+        AS anomalous_range_uid
     FROM miovision_api.volumes_daily_unfiltered AS v
     --left join anomalous_ranges to get notes
     --exclude ['do-not-use', 'questionable'] in HAVING
-    LEFT JOIN miovision_api.anomalous_ranges AS ar ON
+    LEFT JOIN miovision_api.anomalous_ranges AS ar
+    ON
         (
             ar.intersection_uid = v.intersection_uid
             OR ar.intersection_uid IS NULL
@@ -39,7 +40,7 @@ CREATE VIEW miovision_api.volumes_daily AS (
         v.datetime_bins_missing,
         v.unacceptable_gap_minutes,
         v.avg_historical_gap_vol
-    HAVING NOT(array_agg(ar.problem_level) && ARRAY['do-not-use', 'questionable'])
+    HAVING NOT (array_agg(ar.problem_level) && ARRAY['do-not-use', 'questionable'])
     ORDER BY
         v.dt,
         v.intersection_uid,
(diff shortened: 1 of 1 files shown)


gabrielwol:
Preview formatted better here: 

| "dt"         | "intersection_uid" | "classification_uid" | "daily_volume" | "isodow" | "holiday" | "datetime_bins_missing" | "unacceptable_gap_minutes" | "avg_historical_gap_vol" | "anomalous_range_caveats"                                                                                                                                                                                                                                                             | "anomalous_range_uid" |
|--------------|--------------------|----------------------|----------------|----------|-----------|-------------------------|----------------------------|--------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------|
| "2022-01-27" | 65                 | 1                    | 5658           | 4        | false     | 342                     | 0                          |                          | "{NULL}"                                                                                                                                                                                                                                                                              | "{NULL}"              |
| "2022-01-27" | 65                 | 2                    | 4              | 4        | false     | 1436                    | 0                          |                          | "{NULL}"                                                                                                                                                                                                                                                                              | "{NULL}"              |
| "2022-01-27" | 65                 | 3                    | 18             | 4        | false     | 1422                    | 0                          |                          | "{""From this point forward, this category includes streetcars, where before it did not.""}"                                                                                                                                                                                          | "{122}"               |
| "2022-01-27" | 65                 | 4                    | 110            | 4        | false     | 1343                    | 0                          |                          | "{NULL}"                                                                                                                                                                                                                                                                              | "{NULL}"              |
| "2022-01-27" | 65                 | 9                    | 2              | 4        | false     | 1438                    | 0                          |                          | "{""From this point forward, this classification shoul no longer include streetcars where before it did""}"                                                                                                                                                                           | "{123}"               |
| "2022-01-27" | 65                 | 10                   | 9              | 4        | false     | 1431                    | 0                          |                          | "{""Volumes might be low because of an algorithm issue during this range. Volumes for bikes in crosswalks (classification_uid 7) were present in the API but excluded from the database. "",""Mioivision incorrectly idenitified bikes as pedestrians, those records are removed.""}" | "{893,897}"           |

volumes/miovision/sql/views/create-view-volumes_daily.sql
```diff
@@ -8,10 +8,16 @@ CREATE VIEW miovision_api.volumes_daily AS (
         v.holiday,
         v.datetime_bins_missing,
         v.unacceptable_gap_minutes,
-        v.avg_historical_gap_vol
+        v.avg_historical_gap_vol,
+        array_agg(ar.notes ORDER BY ar.range_start, ar.uid)
```
chmnata, line 12: seems like if there are no notes the value would look like `{NULL}`, should we make it actually null with 
`array_agg(ar.notes ORDER BY ar.range_start, ar.uid) filter(where notes is not null) ` instead?

gabrielwol: Good suggestion! Also added column comments for those columns.

chmnata (approved):

chmnata:
Merging! @gabrielwol you can make the changes in db :meow_ez:

gabrielwol:
✅ with help from Nate who had to drop some dependencies.

Issue #874, opened 2024-02-13
Miovision: action items on anomalous_ranges and discontinuities


gabrielwol:
Transferring over action items from [meeting notes on anomalous_ranges](https://www.notion.so/bditto/2024-02-08-AR-and-DISCO-bad5d29cf07645148c87d2a0da2e9b21) to github for more visibility.

**Ranges to add:** 
- [x]  Add schema comments as ARs: Modes are considered reliable enough to use as of... “June 2019 for pedestrians”, "September 2019 for bikes"
- [ ]  rethink zero-padding of classifications in backfilled data before those classifications came online
  - is this an AR to add?
 
**Rethink investigation_level x problem_level.** Nate already did some work on consolidating.  
- [x]  Nate to have a think about the investigation_level x problem_level combos  
- [ ]  Cross constraint in anomalous ranges on investigation levels and problem levels
- [x]  remove `investigating` investigation_level, replacing with `suspect`

**Create discontinuity table**: https://github.com/CityofToronto/bdit_data-sources/pull/877
- [x] @Nate-Wessel evaluate all the null `intersection_uid` entries and evaluate if they belong in AR or Disco table (or both). Action item moved to https://github.com/CityofToronto/bdit_data-sources/pull/877. 
- [x] Add issues from here: https://github.com/CityofToronto/bdit_data-sources/issues/846#issuecomment-1947241738

**Updates to anomalous_ranges structure:** 
- [x] add `leg` to anomalous_ranges (nulls allowed): https://github.com/CityofToronto/bdit_data-sources/tree/874-anomalous_ranges-and-discontinuities-updates
- [x] ~~`intersection_uid` column of anomalous_ranges should not be NULL (after disco table is created).~~ Item axed after [discussion](https://github.com/CityofToronto/bdit_data-sources/issues/874#issuecomment-1944105613). 
- [x] Any changes to anomalous_ranges will require updates to `identify_zero_counts`

**features to add**
- [x] A weekly reminder to investigate open issues with volumes. Added in https://github.com/CityofToronto/bdit_data-sources/pulls/854

**Documentation**
- [ ]  Prepare example cases/documentation on creating anomalous ranges

Nate-Wessel:
>  remove `investigating` investigation_level, replacing with `suspect`

This is partly done. Someone more powerful than I just needs to run 

```sql
DELETE
FROM miovision_api.anomaly_investigation_levels
WHERE uid = 'investigating'
```

[...]

Issue #846, opened 2024-01-29
New model being deployed to increase bike accuracy


radumas:
Being deployed to the following locations after noon today

- Bayview Avenue and River Street
- Bloor Street West and Bay Street
- Bloor Street West and Dufferin Street
- College Street and University Ave
- Danforth Avenue and Broadview Avenue
- Danforth Avenue and Logan Avenue
- Dundas Street West and Bloor Street West
- Greenwood Avenue and Danforth Avenue
- Harbord Street/St. George Street/Hoskin Street
- Queen Street West and University Avenue
- Runnymede Road and Bloor Street West
- Sheppard Avenue West and Keele Street
- Yonge Street / Davenport Road / Church St
-

radumas:
As mentioned elsewhere (in relation to ecocounter), we should probably log these model updates to a table (start_date, end_date, intersection, notes). Though there is likely going to be some squishiness to them.

Nate-Wessel:
See what we did / are doing for Ecocounter in #831. 

I'd love to make this as consistent as possible with whatever we end up doing or Miovision. 
(This issue is about Miovision locations, right?)

radumas:
Noting a change was deployed to the following locations on 2024-02-06 which **should not affect counting**

>Bay st. & Front
Bay St. & Wellington St
Bay St. & King St.
Bay st. & Adelaide:  
Bay St. & Richmond St. 
Bay St. & Queen St.

gabrielwol:
Drafting these discontinuities up as test cases for new table:

```sql
WITH model_changes AS (
    SELECT intersection_name, start_date, notes, classification_uid FROM (VALUES
        ('Bayview Avenue and River Street', '2024-01-29 12:00:00'::timestamp, 'New model being deployed to increase bike accuracy.'),
        ('Bloor / Bay', '2024-01-29 12:00:00'::timestamp, 'New model being deployed to increase bike accuracy.'),
        ('Bloor Street West and Dufferin Street', '2024-01-29 12:00:00'::timestamp, 'New model being deployed to increase bike accuracy.'),
        ('College / University', '2024-01-29 12:00:00'::timestamp, 'New model being deployed to increase bike accuracy.'),
        ('Danforth / Broadview', '2024-01-29 12:00:00'::timestamp, 'New model being deployed to increase bike accuracy.'),
        ('Danforth / Logan', '2024-01-29 12:00:00'::timestamp, 'New model being deployed to increase bike accuracy.'),
[...]

PR #877, opened 2024-02-15
#874 add miovision breaks (discontinuities) table

gabrielwol:
- create `miovision_api.breaks` table. 

## Issue(s) this solves:

- #874 

- Copied basic structure from ecocounter. 
- I think it would be best to create this table, start adding issues, and continue discussion from there. 
- Should we actually name it `discos`? 

- [x] create table!
- [x] add some starting entries: https://github.com/CityofToronto/bdit_data-sources/issues/846#issuecomment-1947241738
- [x] @Nate-Wessel evaluate all the null intersection_uid entries and evaluate if they belong in AR or Disco table (or both).


volumes/miovision/sql/README.md
```diff
@@ -317,6 +318,19 @@ Used to determine the maximum acceptable gap for use in `unacceptable_gaps` tabl
...
+The discontinuity table contains `Moments in time when data collection methods changed in such a way that we would expect clear pre- and post-change paradigms that may not be intercomparable`.
+
+| column_name | Comments | data_type | sample |
+| ------ | ----------- |
+| uid    | simple incrementing primary key | serial | 1 |
+| intersection_uid | Reference to intersections table. A null intersection_uid refers to all intersections. | integer
+| classification_uid | Reference to classifications table. A null classification_uid refers to all classifications. | integer
+| break | Moment the change takes place | timestamp
+| give_or_take | approximate bounds if the precise time is not known | interval
+| notes | required description of what changed - be verbose! | text
```
Nate-Wessel, line 331: I don't think this table is formatted properly. It's not rendering right for me.

gabrielwol: Whoops. Fixed now.

Nate-Wessel (approved):
> * I think it would be best to create this table, start adding issues, and continue discussion from there.

I think that's a good approach. I'd expect the structure to change somewhat, but I think it will take a little bit of experience with this to figure out exactly how. 

> * Should we actually name it `discos`?

I'd be open to calling it `breaks` if we think `discontinuities` is a bit too much of a five-dollar-word. (full disclosure: I actually misspelled it as I was typing it just now and had to go back and fix it...)

volumes/miovision/sql/table/create-table-discontinuities.sql
```diff
@@ -0,0 +1,42 @@
+CREATE TABLE miovision_api.discontinuities (
+    uid serial PRIMARY KEY,
+    intersection_uid integer,
+    classification_uid integer,
```
Nate-Wessel, line 4: Should add `REFERENCES` constraints on the appropriate tables.

gabrielwol: Does that support nulls? 🤷‍♂️

Nate-Wessel: It does! (We don't? We should...)

Nate-Wessel: Just double checked and `anomalous_ranges` actually does have these constraints already. 
`miovision_qc_intersection_uid_fkey` and
`miovision_qc_classification_uid_fkey`

gabrielwol: Whoops don't know how I missed those. I will add to this table too.

chmnata (approved):

Nate-Wessel:
> @Nate-Wessel can you see the one remaining action item before we close the PR?

I'll put that on my to-do list for Monday!

Nate-Wessel:
@gabrielwol I took a look and decided to just breakify the anomalies related to the streetcar reclassification. The rest look like legit anomalies to me. I removed those two from `anomalous_ranges` and added them to `breaks`.

Issue #884, opened 2024-02-22
Centreline: Create directional centreline table for routing and routing function


chmnata:
create centreline table for routing and a function to get centreline between two intersections (that always uses the latest version of centreline in gis_core)

PR #845, opened 2024-01-26
Document the centreline data

Nate-Wessel:
## What this pull request accomplishes:

- documents the centreline data

## Issue(s) this solves:

closes #801

## What, in particular, needs to reviewed:

- facts, clarity

## What needs to be done by a sysadmin after this PR is merged

NA

Issue #819, opened 2024-01-03
Many `centreline` intersections are not connected to `centreline` lines


Nate-Wessel:
## My understanding
The table `gis_core.centreline` should contain all the centreline lines
The table `gis_core.centreline_intersection_point` should contain all the intersections of those lines. 
(Or does `gis_core.intersection` contain the intersections?? I can't tell the diference between these tables)

A query like this should give me the complete, current centreline dataset:
```sql
SELECT
    centreline_id,
    geom -- linestrings
FROM gis_core.centreline
WHERE version_date = '2024-01-01'

UNION

SELECT
    intersection_id AS centreline_id,
    geom -- points
FROM gis_core.intersection
WHERE version_date = '2024-01-01'
```

## However
The intersections table (either of them) includes many hundreds of intersection points not connected at all to the lines. It seems many of these are intersections of footpaths, railroads, etc. I know some of that is ostensibly in the centreline data. But the `centreline` line table doesn't contain it?

![Capture](https://github.com/CityofToronto/bdit_data-sources/assets/7830465/5782f15f-36a4-4350-8596-ca79764fad99)
(a map of the results from the query above)

## What I expected
I would expect the intersections to consistently intersect with the lines. Either we're missing lines or we have too many intersections. Or the documentation doesn't specify what is otherwise going on here.

radumas:
Aye this documentation is a WIP. Apologies for the confusion.

The pipeline is filtering out centrelines which are not streets before inserting them into the table.

https://github.com/CityofToronto/bdit_data-sources/blob/5d4035e4d8ac30963d06ecc9eae4219ff78210a2/gis/gccview/gcc_puller_functions.py#L319

Unfortunately the challenge with the intersection layers is that there is no set of classifications that encompass “this is an intersection of two streets”. But we should make a transformation of the intersections which contains only street intersections to make life easier for us.

Nate-Wessel:
It looks like the locations are precise enough to do a simple spatial join, though it leaves a lot of midblock "intersections". 

```sql
SELECT
    centreline_id,
    geom -- linestrings
FROM gis_core.centreline
WHERE version_date = (SELECT MAX(version_date) FROM gis_core.centreline)

UNION

SELECT
    intersection_id AS centreline_id,
    intersection.geom -- points
FROM gis_core.intersection
JOIN gis_core.centreline ON ST_Intersects(
    intersection.geom,
    centreline.geom
)
WHERE
[...]

PR #885, opened 2024-02-22
Centreline mat invest

chmnata:

- Creates centreline_latest as a mat view and a trigger function that looks at gis_core.centreline updates
- Create routing table and function for directional routing with latest centreline

## Issue(s) this solves:

- #884 

- the code! and if the comments are clear

- Nothing, I already created them 🙈


radumas:
When #845 is merged you should add documentation for using these handy views and functions

gis/centreline/sql/create_view_routing_centreline_directional.sql
```diff
@@ -0,0 +1,33 @@
+CREATE OR REPLACE VIEW gis_core.routing_centreline_directional AS
+
+SELECT
+    centreline.centreline_id,
+    concat(centreline.centreline_id, 0)::integer AS id,
```
radumas, line 5: Unfortunately this isn't like the HERE `link_id`

image

If we need these to be unique then we could instead add 100,000,000 and 200,000,000 to the IDs

radumas (changes requested):
Very nice. You have triggered innovation.

One comment on the directional IDs.

gis/centreline/sql/create_view_routing_centreline_directional.sql
chmnata: gahhh maybe I will just row_number it 🤔 edge id changing through out version should be ok right?

gis/centreline/sql/create_matview_centreline_intersection_point_latest.sql
```diff
@@ -0,0 +1,19 @@
+CREATE MATERIALIZED VIEW gis_core.centreline_intersection_point_latest AS
+
+SELECT *
+FROM gis_core.centreline_intersection_point
```
radumas, line 4: We should filter this to only include intersections that match centrelines we're pulling, to solve #819

radumas (changes requested):

Issue #837, opened 2024-01-18
Centreline segments missing from High Park


Nate-Wessel:
The centreline data in `gis_core.centreline` has no segments in High Park. They are present in MOVE, why not in `bigdata`? This is the case for all version dates.

`gis_core.centreline_intersection_point` however shows many points within High Park. 

![Screenshot 2024-01-18 at 12-09-20 MOVE · View Map](https://github.com/CityofToronto/bdit_data-sources/assets/7830465/39ae2131-7799-4339-83db-513e2bb31916)

chmnata:
We are only pulling centrelines that are within these fcode_desc definition:
```
query = {"where": "\"FEATURE_CODE_DESC\" IN ('Collector','Collector Ramp','Expressway','Expressway Ramp','Local','Major Arterial','Major Arterial Ramp','Minor Arterial','Minor Arterial Ramp','Pending')",
```

The roads in High Park are either `Trail` or `Others` so they weren't being pulled.

Nate-Wessel:
I'd suggest that we change that to include `Others` then. While I generally think it's good that some of the data (political boundaries etc) is filtered out, I think we should at least be matching MOVE. The streets through High Park are pretty substantial infrastructure with foot, bike, car, and even regular TTC bus traffic.

chmnata:
All `others` in this version of centreline.
![image](https://github.com/CityofToronto/bdit_data-sources/assets/46324452/5477c33b-8036-486c-988c-3827b0eb1d24)

Nate-Wessel:
I see that some of these we probably wouldn't want to bring in, but possibly some others (like the ones in High Park) could be selectively included by centreline_id?

radumas:
Or some combination of `linear_name_type` and `jurisdiction`?

chmnata:
Summary of what those `OTHER` streets holds:
- They are mostly private owned with some unknowns and federal + city owned roads.

| jurisdiction| counts   |
|-------------------|------|
| "CITY OF TORONTO" | 38   |
| "FEDERAL"         | 24   |
| "PRIVATE"         | 1885 |
| "UNKNOWN"         | 163  |

*Unknown contains high park blvd* :meow_dio: 

Judging by the street name of most of these "OTHER", they seems like private residential streets. However there are some "walkways" that are peds only, e.g. Lawrence Walkway. 

[...]

PR #888, opened 2024-02-26
Add centreline OTHER and fix a bug for gcc puller

chmnata:

- Add OTHER in query parameter for centreline puller
- Fix one missing package bug in dags_function (missing json)

## Issue(s) this solves:

- #837 

- the few lines of changed code.

Check symlink to data_scripts in morbius.

### DIFF
--- a/dags/dag_functions.py
+++ b/dags/dag_functions.py
@@ -3,6 +3,7 @@
 """Common functions used in most of the DAGs."""
 import os
 import re
+import json
 from typing import Optional, Callable, Any, Union
 from airflow.models import Variable
 from airflow.hooks.base import BaseHook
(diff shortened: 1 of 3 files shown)


chmnata:
Did a test pull for centreline, and while it does now pull `OTHER`, it doesn't update the materialized view to the latest date. This is because the trigger only applies to the parent table and doesnt cascade to the child table and apparently inserting data to the child table != `INSERT, UPDATE` on the parent table..... When I was testing I was inserting directly to the parent table in my test table, but the gcc_puller actually insert directly to the child table so the trigger wasn't working. 
Will update the gcc puller to just insert to the parent table, or look into a way to apply the trigger to all child tables

gabrielwol (approved):

chmnata:
Easiest way is to insert the data to the parent table, instead of creating at trigger for every child table. I have made this changes in gcc puller script. Tested it on airflow and now the latest version lives in `gis_core.centreline_20240227`, the first version with `OTHER`. `gis_core.centreline_latest` also got triggered and updated automatically.

PR #896, opened 2024-03-05
Note that HERE clips speeds to 130 kmph

Nate-Wessel:

- Notes a quirk of the HERE dataset

## Issue(s) this solves:

- adds to docs

- facts

NA


Nate-Wessel:
As evidence: 

Capture

This won't return any results:
```sql
SELECT *
FROM here.ta_path
WHERE max_spd > 130
LIMIT 1
```

Nate-Wessel:
This will return results:
```sql
SELECT *
FROM here.ta_path
WHERE max_spd = 130
LIMIT 1
```

This won't:
```sql
SELECT *
FROM here.ta_path
WHERE max_spd > 130
LIMIT 1
```

radumas:
I don't think we should document this publicly

PR #898, opened 2024-03-05
Clarify HERE's `stddev` calculation

Nate-Wessel:

- clarifies some ambiguity in the `stddev` field for HERE data 

## Issue(s) this solves:

- It was ambiguous whether this was the _sample_ or _population_ metric because there are `0`'s, where there should be `NULL`s. 

- facts

NA


Nate-Wessel:
```sql
SELECT
    link_dir,
    tx,
    stddev,
    stddev_samp(obs.speed),
    stddev_pop(obs.speed)
FROM here.ta_path, 
LATERAL (SELECT unnest(ARRAY[min_spd,max_spd]) AS speed) AS obs
WHERE sample_size = 2
GROUP BY
    link_dir,
    tx,
    stddev
LIMIT 100
```

Issue #900, opened 2024-03-08
GCC Puller: Create primary key only if not exist


chmnata:
Ran into an error 
`ERROR:  multiple primary keys for table "loop_detector" are not allowed`

Not exactly sure why this wasn't an issue before.... 
And because there is no `add constraint if exist`, there are a couple of solutions:
- drop primary key and create again (NO CANNOT DO, BECAUSE IT WILL MESS WITH OUR AUDIT TABLE STRUCTURE)
- Look at information table and see if a primary key exist on the table
- Edit the create table line to do something like
```
Create table if not exist (
column1 data_type, 
column2 data_type, 
constraint pk_mail_app_recipients primary key (column1)
```
Then it will only add a primary key constraint when a new table is created.

https://github.com/CityofToronto/bdit_data-sources/blob/f79fd6eee455f2ca35a002ba081270f8f13012c1/gis/gccview/gcc_puller_functions.py#L191

gabrielwol:
@chmnata should this issue also be closed as not planned?

PR #910, opened 2024-03-27
GCC Puller Primary Key

MelinaGoula:

- modify `create_audited_table()` to only create a primary key when a new table is created
- fixes the issue of creating multiple primary keys for a table 

## Issue(s) this solves:

- #900 

- `gcc_puller_functions.py`

- testing of the new code


chmnata:
Thanks a lot for the update, upon further investigation, the issue seems to be having an unrenamed table in the database, e.g. `_loop_detector` from a previous failed action. So the duplicated primary key error was alerting about adding the same primary key to the previously unrenamed tabled. This issue was resolved after deleting the unrenamed table in the database.

PR #924, opened 2024-04-09
Eco counter dagify

### LINKED ISSUE #912: Ecocounter: automate data pulls with an Airflow DAG

gabrielwol:
- Builds on Nate's ecocounter code for a daily airflow DAG.
- Automatically inserts new sites/channels (marked as `validated = False`) and notifies in slack.

## Issue(s) this solves:
- Closes #912 

- 

- Update dta_scripts path
- Replace all references to `gwolofs` schema with `ecocounter`. Also add validated columns to ecocounter and backfill data since the last ecocounter pull.


dags/ecocounter_pull.py
```diff
@@ -0,0 +1,186 @@
...
+                    channel_id, channel_name = channel['id'], channel['name']
+                    if not flowIsKnownToUs(channel['id'], conn):
+                        insertFlow(conn, channel_id, site_id, channel_name, channel['interval'])
+                        new_flows.append({
+                            'site_id': site_id,
+                            'channel_id': channel_id,
+                            'flow_name': channel_name
+                        })
+
+        if new_sites != []:
```
Nate-Wessel, line 94: slightly clearer to do `if len(new_sites) > 0:`?

dags/ecocounter_pull.py
```diff
@@ -0,0 +1,186 @@
...
+                    })
+                for channel in site['channels']:
+                    channel_id, channel_name = channel['id'], channel['name']
+                    if not flowIsKnownToUs(channel['id'], conn):
+                        insertFlow(conn, channel_id, site_id, channel_name, channel['interval'])
+                        new_flows.append({
+                            'site_id': site_id,
+                            'channel_id': channel_id,
+                            'flow_name': channel_name
+                        })
```
Nate-Wessel, line 92: A "channel" _is_ a "flow", right? 

This block could be a little confusing later if we forget about the incongruous terminologies. Perhaps just stick to "flow" as much as possible with the var names.

dags/ecocounter_pull.py
```diff
@@ -0,0 +1,186 @@
...
+        
+        start_time = dateutil.parser.parse(str(ds))
+        end_time = dateutil.parser.parse(str(ds_add(ds, 1)))
+        LOGGER.info(f'Pulling data from {start_time} to {end_time}.')
+
+        with eco_postgres.get_conn() as conn:
+            for site_id in getKnownSites(conn):
+                LOGGER.debug(f'Starting on site {site_id}.')
+                #if site_id != 300028589:
+                #    continue
```
Nate-Wessel, line 129: remove old comment?

dags/ecocounter_pull.py
```diff
@@ -0,0 +1,186 @@
...
+                for channel_id in getKnownChannels(conn, site_id):
+                    LOGGER.debug(f'Starting on flow {channel_id} for site {site_id}.')
```
Nate-Wessel, line 131: Again, probably should stick to one consistent term here.

Nate-Wessel (commented):
Looks good, acknowledging that I don't really know how Airflow works and can't run this. 

The main thing I notice is the inconsistent use of "channel" vs "flow" - I'd suggest sticking with just one of these across the whole Ecocounter universe, and documenting the other term where necessary. E.g. let's just always call them "flows" but acknowledge where necessary that the API calls them "channels".

gabrielwol:
@Nate-Wessel addressed your comments + added a readme section for the DAG. Once you're OK with how we're handling the unvalidated flows, I can merge and adjust all the scripts to point to `ecocounter` schema instead of gwolofs. 
@chmnata can you quickly look over the DAG? It is working as expected: https://trans-bdit.intra.prod-toronto.ca/airflow/dags/ecocounter_pull/grid. Nate has reviewed the rest.

dags/ecocounter_pull.py
```diff
@@ -0,0 +1,190 @@
...
+        if len(new_flows) > 0:
+            LOGGER.info('New flows added: %s', new_flows)
+            extra_msg = '\n'.join([str(item) for item in new_flows])
+            send_slack_msg(
+                context=context,
+                msg=f"There were new flow_ids when pulling ecocounter data :eco-counter:",
+                attachments=[{"text": extra_msg}]
+            )
+        if new_flows == [] and new_sites == []:
+            #Mark skipped to visually identify which days added new sites or flows in the UI.
```
chmnata, line 117: cool!

dags/ecocounter_pull.py
```diff
@@ -0,0 +1,190 @@
...
+except:
+    raise ImportError("Cannot import DAG helper functions.")
+
+DAG_NAME = 'ecocounter_pull'
+DAG_OWNERS = Variable.get('dag_owners', deserialize_json=True).get(DAG_NAME, ["Unknown"])
+
+README_PATH = os.path.join(repo_path, 'volumes/ecocounter/readme.md')
+DOC_MD = get_readme_docmd(README_PATH, DAG_NAME)
+
+CONFIG_PATH = '/data/airflow/data_scripts/volumes/ecocounter/.api-credentials.config'
```
chmnata, line 44: wondering if we should put this as an airflow connection instead 🤔

dags/ecocounter_pull.py
```diff
@@ -0,0 +1,190 @@
...
+            raise AirflowSkipException('No new sites or flows to add. Marking task as skipped.')
+
+    @task(trigger_rule='none_failed')
+    def pull_ecocounter(ds):
+        config = ConfigParser()
+        config.read(CONFIG_PATH)
+        token = getToken(CONFIG_PATH)
+        eco_postgres = PostgresHook("ecocounter_bot")
+        
+        start_time = dateutil.parser.parse(str(ds))
```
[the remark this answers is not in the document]
gabrielwol: Agreed, changed to start_date and end_date! 
ds_format returns a string, not a datetime object, so I'll stick with dateutil.parser.

dags/ecocounter_pull.py
gabrielwol: Will do!

gabrielwol:
@Nate-Wessel , @chmnata 
Ready for re-review! The DAG is now inserting into `ecocounter.counts_unfiltered` and there is a new view `ecocounter.counts` which has only validated flows. Also added partitioning to `ecocounter.counts_unfiltered` on Natalie's request.

Nate-Wessel (changes requested):
I think we should do the same view-validation-filter-layer for the `sites` and `flows` tables, following the same pattern as for `counts`. Those now also have non-validated junk entries in them (sites at [0,0], etc). 

We should also update the docs to reflect that these are now views, not tables. 
https://github.com/CityofToronto/bdit_data-sources/blob/1d69df99f2f8b4bd760e60257213f57666062f38/volumes/ecocounter/readme.md?plain=1#L22-L26

Not sure where best to mention the unfiltered tables though... probably not in the high-level overview there.

chmnata (commented):
DAG looks good!

Nate-Wessel (approved):

Issue #927, opened 2024-04-12
Add a replicator table check


gabrielwol:
Add a notification to bigdata replicator DAG which identifies discrepancies between move_staging updated tables and airflow *_tables variables.

PR #928, opened 2024-04-12
927 add a replicator health check

gabrielwol:
- Creates a new DAG (`replicator_table_check`) to monitor the MOVE -> bigdata replication process 
- This will help notify us if the list of tables being replicated is out of sync (hint: it currently is)
- See latest slack alerts for review:
https://bditto.slack.com/archives/C021YMH9BK5/p1713196079217679
https://bditto.slack.com/archives/C021YMH9BK5/p1713196079227109

## Issue(s) this solves:
- #927 

- 

- For @peter-lyons: we should have the MOVE replicator trigger this DAG through the API at the same time as the counts and collisions replicators.

### DIFF
--- a/dags/replicator_table_check.py
+++ b/dags/replicator_table_check.py
@@ -49,6 +49,10 @@ def replicator_DAG():
 
     @task()
     def tables_to_copy():
+        """This task finds all the replicators from `replicators` airflow variable,
+        and then finds all the tables listed for replication by looking at the Airflow variable
+        listed in the `tables` key item in the replicator dictionaries."""
+
         #a list of the replicators
         replicators = Variable.get('replicators', deserialize_json=True)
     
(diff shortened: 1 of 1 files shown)


dags/replicator_table_check.py
```diff
@@ -0,0 +1,161 @@
...
+        SELECT pgn.nspname::text || '.' || pgc.relname::text, pgd.description
+        FROM pg_description AS pgd
+        JOIN pg_class AS pgc ON pgd.objoid = pgc.oid
+        JOIN pg_namespace pgn ON pgc.relnamespace = pgn.oid
+        WHERE
+            pgn.nspname = 'move_staging'
+            AND pgc.relkind = 'r'
+            AND pgd.description ILIKE %s"""
+
+        con = PostgresHook("collisions_bot").get_conn()
```
chmnata, line 85: is it all collision related tables? Should we use replicator_bot instead?

gabrielwol: It's both collisions and counts tables within `move_staging`, but from my testing, using this method works even if you don't have correct select permissions. 

Tested as `vds_bot`: 

Returns 62 rows:
```sql
SELECT pgn.nspname::text || '.' || pgc.relname::text, pgd.description
FROM pg_description AS pgd
JOIN pg_class AS pgc ON pgd.objoid = pgc.oid
JOIN pg_namespace pgn ON pgc.relnamespace = pgn.oid
WHERE
    pgn.nspname = 'move_staging'
    AND pgc.relkind = 'r'
```

Original method using `obj_description()` returns 0 rows:
```sql
WITH move_staging_comments AS (
    SELECT table_schema::text || '.' || table_name::text AS tbl_name
    FROM information_schema.tables
    WHERE table_schema = 'move_staging'
)

SELECT tbl_name, obj_description(tbl_name::regclass) 
FROM move_staging_comments
```

So it doesn't really matter which connection we use here.

chmnata (approved):
Thanks @gabrielwol for making this happen so quickly. Everything looks good!
Kinda hope we had an audit table on move_staging, so we can do something like the below instead of parsing from the table comment. 🤔 Would work on tables without comments automatically generated as well. Not sure if its worth implementing the whole audit table thingy however :meow_dunno: 
```
select table_name, max(action_tstamp_clk) from gis_core.logged_actions 
group by table_name 
```

gabrielwol:
I think the audit table might be overkill in this case since all these tables have comments from the upstream DAG. But maybe we could have something like this using audit table for GIS!

dags/replicator_table_check.py
chmnata: not really on the permission issue but more on using the collision bot for only collision related tasks, since they all have limited connection (granted it has 20 :meow_dio:) and we already have a replicator bot

gabrielwol:
@chmnata Some minor updates since your last review. Decided with Peter to trigger this locally (daily) since triggering via MOVE replicator would require more changes. 
Mostly waiting on the replicator_bot connection to merge!

chmnata (approved):
APPROVED

PR #950, opened 2024-05-10
949 missing leg in miovisionopen issues join

### LINKED ISSUE #949: Missing leg in miovision.open_issues join to anomalous_ranges

gabrielwol:
- The addition of leg to anomalous_ranges broke the open_issues view + data check. 
- Other minor improvements include: fixing the `last_week_volume` to only look at dates within the AR + formatting that number with commas

## Issue(s) this solves:

- Closes #949 

- 

- nothing

### DIFF
--- a/volumes/miovision/sql/views/create-view-open_issues.sql
+++ b/volumes/miovision/sql/views/create-view-open_issues.sql
@@ -13,6 +13,7 @@ SELECT
         WHEN ar.classification_uid IS NULL THEN 'All modes'
         ELSE c.classification
     END,
+    ar.leg,
     ar.range_start::date,
     (current_timestamp AT TIME ZONE 'EST5EDT')::date - ar.range_start::date AS num_days,
     ar.notes,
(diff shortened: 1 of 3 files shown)


volumes/miovision/sql/views/create-view-open_issues.sql
```diff
@@ -26,18 +44,19 @@ JOIN miovision_api.intersections AS i USING (intersection_uid)
 --find last week volume
 LEFT JOIN miovision_api.volumes AS v
     ON ar.intersection_uid = v.intersection_uid
+    --volume within the last 7 days and after AR started
+    AND v.datetime_bin >= ar.range_start
```
chmnata, line 48: so for example if we run this today, and the range was add yesterday, 
the sum volume for `last_week_volume` would be summing up only 1 day of data

gabrielwol: Yes, because we wouldn't want to know the volume from before `range_start`, which would either be zero, or non-zero from a time before the issue started. I could rename it/change it to average daily volume if you think that's better?

chmnata: yess, to reduce confusion if that ever happens

gabrielwol: Decided to explain it with a column comment and also in the `check_open_anomalous_ranges` airflow alert.

chmnata (approved):
:gabe-approved:

chmnata:
Updated `miovision_api.open_issues`

PR #957, opened 2024-05-21
953 reduce notification frequency to weekly

gabrielwol:
## What this pull request accomplishes:

- changes two daily "_check" DAGs to weekly.
  - adjust the schedule interval, external task sensors, and the sql params
- removed a non critical `check_gaps` check for miovision gaps > 4 hours, < 1 day

## Issue(s) this solves:
- #953 

## What, in particular, needs to reviewed:
- My method of linking weekly dag to daily: set task sensor to look 1 week ahead of schedule interval.

## What needs to be done by a sysadmin after this PR is merged
- nothing

PR #961, opened 2024-05-23
#954 Add Citywide TTI DAG 

chmnata:

- Restart the citywide TTI aggregation task in this new DAG and get rid of all the other tasks that are not relevant anymore
- Taskflow it - ish 

## Issue(s) this solves:

- #954 

- 

Move to data script


dags/citywide_tti_aggregate.py
```diff
@@ -0,0 +1,86 @@
...
+
+"""
+DAG_NAME = 'citywide_tti_aggregate'
+
+DAG_OWNERS = Variable.get('dag_owners', deserialize_json=True).get(DAG_NAME, ["Unknown"]) 
+
+# Slack alert
+SLACK_CONN_ID = 'slack_data_pipeline'
+
+default_args = {'owner':'natalie',
```
gabrielwol, line 45: You can use the DAG_OWNERS variable for owner and remove your email!

dags/citywide_tti_aggregate.py
```diff
@@ -0,0 +1,86 @@
...
+    catchup=False
+)
+
+## Tasks ##
+## ExternalTaskSensor to wait for pull_here
+def citywide_tti_aggregate():
+    
+    wait_for_here = ExternalTaskSensor(task_id='wait_for_here',
+                                       external_dag_id='pull_here',
+                                       external_task_id='pull_here',
```
gabrielwol, line 71: This did make me wonder why we use ExternalTaskMarker normally and [the reason is](https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/external_task_sensor.html): 

> ExternalTaskSensor can be used to establish such dependencies across different DAGs. When it is used together with ExternalTaskMarker, clearing dependent tasks can also happen across different DAGs.

dags/citywide_tti_aggregate.py
```diff
@@ -0,0 +1,86 @@
...
+    doc_md = doc_md,
+    tags=["HERE"],
+    catchup=False
+)
+
+## Tasks ##
+## ExternalTaskSensor to wait for pull_here
+def citywide_tti_aggregate():
+    
+    wait_for_here = ExternalTaskSensor(task_id='wait_for_here',
```
gabrielwol, line 69: Add a timeout?

dags/citywide_tti_aggregate.py
```diff
@@ -0,0 +1,86 @@
...
+                'email_on_success': False,
+                'retries': 0,
+                'retry_delay': timedelta(minutes=5),
+                'on_failure_callback': task_fail_slack_alert
+                }
+
+@dag(
+    DAG_NAME, 
+    default_args=default_args, 
+    schedule_interval='30 16 * * * ', # same as pull_here task
```
gabrielwol, line 59: `pull_here` normally takes 20 minutes, so if you schedule this DAG later (+adjust `execution_delta`), the ExternalTaskSensor won't have to retry.

dags/citywide_tti_aggregate.py
```diff
@@ -0,0 +1,86 @@
...
+    
+    wait_for_here = ExternalTaskSensor(task_id='wait_for_here',
+                                       external_dag_id='pull_here',
+                                       external_task_id='pull_here',
+                                       start_date=datetime(2024, 5, 16)
+                                       )
+
+## Postgres Tasks
+# Task to aggregate citwyide tti
+    aggregate_daily = PostgresOperator(sql='''SELECT covid.generate_citywide_tti('{{ yesterday_ds }}')''',
```
gabrielwol, line 77: `yesterday_ds` is [deprecated](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html)! I think instead use `ds_add(ds, -1)`

dags/citywide_tti_aggregate.py
```diff
@@ -0,0 +1,86 @@
...
+                                       task_id='aggregate_daily',
+                                       postgres_conn_id='natalie',
```
gabrielwol, line 79: `natalie`? 👀

gabrielwol (changes requested):

dags/citywide_tti_aggregate.py
chmnata: gah so true. Created all the wait_for_here cross dependencies DAG before this feature was out and just never changed it :simpson-hide:

dags/citywide_tti_aggregate.py
chmnata: It doesn't retry and just starts poking at 16:30 and wait to it to successfully finishes. Does ExternalTaskSensor fails and retry?

dags/citywide_tti_aggregate.py
chmnata: shouldn't have copy 🍝

dags/citywide_tti_aggregate.py
chmnata: :simpson-hide:

dags/citywide_tti_aggregate.py
gabrielwol: actually you need to use `macros.ds_add(...` as I just found out in #957.

dags/citywide_tti_aggregate.py
gabrielwol: We normally use [mode="reschedule"](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/sensors.html)+`poke_interval` which doesn't take up a worker slot.

dags/citywide_tti_aggregate.py
chmnata: `from airflow.macros import ds_add, ds_format` I used ds_add in here_path DAG with this import!

gabrielwol (approved):

Issue #966, opened 2024-05-28
Add centreline_id field to ecocounter sites


gabrielwol:
Add centreline_id field to ecocounter sites for dashboarding needs.

PR #968, opened 2024-05-28
#966 add centreline_id to ecocounter.sites and sites_unfiltered

gabrielwol:

- adds `centreline_id` to `ecocounter.sites` and `ecocounter.sites_unfiltered`. 
- a sql script used to identify new matches

## Issue(s) this solves:

- #966 

- Decision to add centreline_id to sites and not to flows. I figured for flows it would be duplicated between the flows at a given site which would make it very difficult to visualize. 
- Should I also add cl_geom to `ecocounter.sites` view?

Nothing.

### DIFF
--- a/volumes/ecocounter/readme.md
+++ b/volumes/ecocounter/readme.md
@@ -146,6 +146,7 @@ Key tables `ecocounter.sites_unfiltered`, `ecocounter.flows_unfiltered`, `ecocou
 
 ### `ecocounter.sites_unfiltered`
 CAUTION: Use VIEW `ecocounter.sites` which includes only sites verified by a human. Sites or "locations" of separate ecocounter installations. Each site may have one or more flows.
+When you want to update new rows with missing `centreline_id`s, use [this script](./updates/ecocounter_centreline_updates.sql).  
 
 | column_name | data_type | sample | comments |
 |:------------|:----------|:-------|:---------|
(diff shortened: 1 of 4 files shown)


gabrielwol:
I'm not sure about these two locations. The rest are on arterials and are pretty straightforward.  
site_ids: 300026125, 300026120
image

Nate-Wessel:
> I'm not sure about these two locations. The rest are on arterials and are pretty straightforward. site_ids: 300026125, 300026120 image

I think the site to the east here should definitely not have a centreline id. The one to the west, maybe? But probably not?

It really boils down to a question of what the centreline is meant to represent. Does it implicitly include the sidewalks and other peripheral paths associated with a street? Does it depend how far away they are from the main car lanes, or that they're facing the exact same direction? Does it matter if they're on both sides, or only on one side?

There's an eternal debate in OSM about how and when to map things as separate ways. I tend to [map bike lanes on both sides as part of the central way](https://www.openstreetmap.org/way/15818587), but it's almost always mapped as a separate way if the cycleway is a [two-lane path off on one side of the road](https://www.openstreetmap.org/way/660002977), which some of these Ecocounter locations are, not just the two above. (the way sidewalks are mapped in OSM is pure chaos, by the way)

I'd want to avoid any casual mistakes where someone using this data forgets that the Ecocounter sensors actually _only_ record what's directly above them, in the bike facilities, not the whole street. 

Maybe I'm over-thinking it?

gabrielwol:
@Nate-Wessel I set those two centreline_ids to null. As per @radumas's comment in slack, we could revisit if we add trails to gis_core.centreline. Ready for review.

radumas:
I think the one to the West could be tied to the street centreline (but primarily because the trail centreline doesn't exist 🤷🏻‍♂️ )

Nate-Wessel (changes requested):
I think I would just suggest changing `centreline_id` to `nearest_centreline_id` or `associated_centreline_id` or something like that to make clear the the sensor isn't counting the centreline segment _per se_ bit just the specific bike facilities.

Or at least be clearer in the documentation for this field what the relation is. 

Looks good otherwise!

gabrielwol:
@Nate-Wessel I updated the documentation and the column comment to be more specific about only counting bike ROW!

Nate-Wessel (approved):

Issue #969, opened 2024-05-28
Add id field to beginning of gis_core mat views


gabrielwol:
Update the gis_core.*_latest layers to have id field first. Otherwise they don't work in QGIS:
![image](https://github.com/CityofToronto/bdit_data-sources/assets/80077912/0cc7c02c-91f2-4b21-aa8d-b28d951510e0)

PR #970, opened 2024-05-28
#969 add row number to gis_core latest mat views

gabrielwol:
- adds a unique id field to the beginning of these 3 mat views so they can be added to qgis projects without error.

## Issue(s) this solves:
- #969

- 

- run these 3 sqls


gabrielwol:
Good suggestion. Removing the row number column and adding a unique index on centreline_id worked! I'll do that instead.

chmnata (approved):

gabrielwol:
Success! 
image

Issue #745, opened 2023-11-08
Create Miovision Reporting View(s)


radumas:
Non-exhaustive list of requirements:

- [ ] aggregates from the atr data
- [ ] because we are only aggregating cycling approach counts, this should be entrance volumes only
- [ ] daily agg: checks if there's 23 hours of data for the day
- [ ] anti-join on the anomalous ranges table and filters for....?
- [ ] aggregates vehicle volumes
- [ ] handles the transition from bicycle TMC to approach volumes (see #695 )

gabrielwol:
I began working on this issue with an `hourly_vehicle_volumes` view. need to regroup with @radumas to see if on the right track.

radumas:
Phase 1 can:
* exclude bike volumes
* include unacceptable gaps (filter out the hour with a missing 15-min bin)
* include anomalous gaps (filter out do-not-use, questionable, include valid-*)

Phase 2: figure out cycling volumes

Phase 3: incorporate alerts...?

PR #972, opened 2024-05-30
971 miovision 15 minute views that filter out anomalous ranges

gabrielwol:
- adds views which filter out anomalous ranges from volumes_15min and volumes_15min_mvt aggregate tables. 
- new comments on 15min tables directing to new views.

## Issue(s) this solves:

- #971

- table comments

- create views

### DIFF
--- a/volumes/miovision/sql/views/create-view-volumes_15min_filtered.sql
+++ b/volumes/miovision/sql/views/create-view-volumes_15min_filtered.sql
@@ -1,0 +1,32 @@
+CREATE VIEW miovision_api.volumes_15min_filtered AS (
+    SELECT
+        v15.volume_15min_uid,
+        v15.intersection_uid,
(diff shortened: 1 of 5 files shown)


volumes/miovision/sql/views/create-view-volumes_15min_filtered.sql
radumas, line 1: Would it be possible to include the centreline here so it's like (ish) other ATR data?

volumes/miovision/sql/views/create-view-volumes_15min_mvt_filtered.sql
radumas, line 1: I think this one could include textual names (and associated centreline intersection).

radumas (approved):
Approving, comments on including centreline information can be applied towards #745

Issue #937, opened 2024-04-19
Create Traffic pattern layers for HERE


chmnata:
Create traffic patterns layer for network generation for routing.

PR #974, opened 2024-05-31
update readme and add traffic pattern sql

chmnata:

- 

## Issue(s) this solves:

- #937 

- Traffic pattern sql, does the mean of 5 min bin make sense? Or should I change it to one hour?


here/traffic_patterns/README.md
```diff
@@ -128,5 +128,9 @@ LATERAL (
     ) )
 );
 ```
-## Sad news
-For top secret contract reasons, we are no longer receiving traffic patterns data. As of writing this (in August 2022) the latest (and last) traffic patterns dataset that we received is from 2019. As time goes by, and the street network evolves, traffic patterns will cease to be useful...
\ No newline at end of file
+## Generate Quarterly traffic patterns
+For top secret contract reasons, we are no longer receiving traffic patterns data. The latest (and last) traffic patterns dataset that we received is from 2019. As time goes by, and the street network evolves, traffic patterns will cease to be useful. We wil be creating quaterly traffic patterns for each version for routing purposes. 
+
+Depending on the date you are routing, you should use different quaters, years, and map version of traffic patterns. We will only include weekday (Tue - Thurs), and exclude holiday seasons to create a more [generalized traffic pattern layer](/sql/create_quaterly_traffic_patterns.sql).
```
radumas, line 134: quaters -> quarters

here/traffic_patterns/sql/create_quaterly_traffic_patterns.sql
```diff
@@ -0,0 +1,24 @@
...
+        dt >= _start_date AND dt < _end_date
+        AND EXTRACT(ISODOW FROM dt) IN (2, 3, 4) -- only include tues-thurs traffic
+        AND holiday.dt IS NULL -- excluding holidays
+    GROUP BY routing_streets.link_dir, dt, hr
+)
+
+SELECT
+    link_dir,
+    hr,
+    harmean(daily_cost)::int AS cost
```
radumas, line 22: this isn't a cost it's a speed, should be renamed to speed

radumas (changes requested):
This look pretty good. One pretty important change in terms of explicitly naming columns in the traffic patterns table.

Otherwise I would also suggest specifying what need to be changed in the two queries to make traffic pattern for a particular quarter and revision. Should also make explicit the assumptions in terms of if no data then speed limit and if no speed limit then 50 km/hr

chmnata:
Ready for re-review @radumas, made them into functions instead. :meow_dio:

here/traffic_patterns/sql/create_routing_speed_limit.sql
```diff
@@ -0,0 +1,44 @@
...
+
+    EXECUTE format($$CREATE TABLE here.%I AS
+                    SELECT
+                        routing_streets.link_dir,
+                        COALESCE(
+                            CASE
+                                WHEN RIGHT(routing_streets.link_dir, 1) = 'T' AND att.to_spd_lim > 0 THEN att.to_spd_lim
+                                WHEN RIGHT(routing_streets.link_dir, 1) = 'F' AND att.fr_spd_lim > 0 THEN att.fr_spd_lim
+                                ELSE NULL
+                            END, 50
```
radumas, line 26: `COALESCE(CASE [...]  ELSE NULL END, 50)`, the last part feels redundant

radumas (changes requested):
Smol fix requested

here/traffic_patterns/sql/create_routing_speed_limit.sql
chmnata: got rid of COALESCE

radumas (approved):
:party-porg:

Issue #975, opened 2024-06-03
VDS notification failure


gabrielwol:
- [`vds_pull.check_rows_vdsdata_div2`](https://morbius.corp.toronto.ca:8080/airflow/dags/vds_pull_vdsdata/grid?base_date=2024-06-02T23%3A59%3A59.999Z&dag_run_id=scheduled__2024-05-30T08%3A00%3A00%2B00%3A00&task_id=data_checks.check_rows_vdsdata_div2&tab=logs) slack notification didn't work because of a None value returned from the SQLCheckOperatorWithReturnValue: 
> File "/data/home/gwolofs/bdit_data-sources/dags/dag_functions.py", line 103, in task_fail_slack_alert
    extra_msg_str = '\n> '.join(
TypeError: sequence item 3: expected str instance, NoneType found
- The slack alert shouldn't fail when a None is returned from SQLCheckOperatorWithReturnValue.
- The None was from the weather summary function `weather.airport_weather_summary(date)`, because the weather pipeline runs at 10:30AM and the VDS pipeline runs earlier. We should move up this pipeline so all the volume pipelines can reference it.

radumas:
There might be a reason the weather pipeline was scheduled for 10:30 AM

gabrielwol:
There might be, but it's not mentioned in the docs and there is no readme page for scraping weather.gc.ca. The [package](https://pypi.org/project/env-canada/) we use for predictions also doesn't mention anything. We can check results tomorrow.
Likely it was just to make testing easier.

PR #976, opened 2024-06-03
975 weather DAG task flow plus schedule change

gabrielwol:

- replace pull_weather with task flowified weather_pull + adjust start and end dates
  - adjusted ds-1 to ds to pull historical weather for yesterday and not day before.
  - adds a data check to see if historical weather is all nulls and if so alert. 
- converted extra_msg_str elements to string to prevent NoneType failure.
- remove unused imports in weather modules
- updates the weather summary sql function to have fallback "unknown"
- add a `wait_for_weather` common task to have other data checks wait a few hours for weather data. I think this is better than task sensors/triggers because it should be a soft link without possibility of downstream failures.
  - added to 4 other DAGs

## Issue(s) this solves:
- #975 

- taskflow

- already implemented in data_scripts for testing ☑️


chmnata (changes requested):
Can you update the new scheduled time in the readme as well? https://github.com/CityofToronto/bdit_data-sources/tree/master/weather#data-pipeline

gabrielwol:
Updated the readme and set the time to 830pm using a TimeSensor operator.

gabrielwol:
Not quite ready to merge yet, need to keep trying different start times to find when historical weather becomes available and then push the other data checks back to match.

chmnata:
https://climate.weather.gc.ca/FAQ_e.html#Q17 !!!!

gabrielwol:
I miscalculated 1000UTC, so we need to wait one more day for this to run at the correct time.
- Success: the check tasks still ran and reported unknown weather so the soft link worked as intended.  
- Fail: nulls were silently inserted into the historical weather tables (will add a sql check).

dags/weather_pull.py
```diff
@@ -0,0 +1,105 @@
...
+def weather_pull_dag():
+
+    no_backfill = LatestOnlyOperator(task_id="no_backfill")
+    no_backfill.doc_md = "Pull predicted weather data - can ONLY pull 5 days ahead of run date - no backfill."
+
+    wait_till_1030am = TimeSensor(
+        task_id="wait_till_1030am",
+        timeout=5*3600,
+        mode="reschedule",
+        poke_interval=3600,
```
chmnata, line 65: Looking at the log, it seems like its poking to check whether its 10:30am a few times starting when the dag is active (so 8am, 9am, and 10am). Would we need poke_interval here?

chmnata: @gabrielwol pointed out that the timesensor (no async) needs to be poked to figure out if its at target time yet, so we can't get rid of poke_interval :okfine::okfine::okfine:

chmnata (approved):
:okfine:

PR #994, opened 2024-06-25
645 replication processes check if data

### LINKED ISSUE #645: Replication processes should check there's data in the replicated drop zone
Because these replication processes operate by `TRUNCATE` & `INSERT`, we should check that the replicator actually finished inserting data before we run our own `TRUNCATE` & `INSERT` to our schemas.

Alternatively return the count of inserted rows in the function and have the task fail if that's 0.

gabrielwol:

- Check for non-zero row count in source table before and destination table after truncate/insert.
- The dynamic task mapping is now done over task_group instead of task so that `check_source -> copy_table -> check_dest` happens in sequence for each map index (copy shouldn't proceed if source check failed)
- I tested successfully with a dummy table variable and then tested successfully with the actual table variables. I caught one empty table in move_staging which I notified MOVE about: `move_staging.collision_factors_cyccond` (failure msg [here](https://bditto.slack.com/archives/C04BWRWVB3L/p1719331402031249))

image

## Issue(s) this solves:

- Closes #645

- 

Update in data_scripts

### DIFF
--- a/dags/common_tasks.py
+++ b/dags/common_tasks.py
@@ -40,6 +40,31 @@ def get_variable(var_name:str) -> list:
     return Variable.get(var_name, deserialize_json=True)
 
 @task()
+def check_not_empty(conn_id:str, table:str, **context) -> None:
+    con = PostgresHook(conn_id).get_conn()
+    sch, tbl = table.split(".")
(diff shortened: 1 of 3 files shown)


chmnata (approved):
Looks good!

PR #997, opened 2024-06-25
Filter anomalous ranges from Ecocounter `counts`

### LINKED ISSUE #995: Ecocounter `counts` view should filter out anomalous_ranges
The `counts` [view](https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/ecocounter/views/create-view-counts.sql) currently only checks that flows are validated. It should also filter out `do-not-use` anomalous ranges at either the flow or site level. These data would still be available in the raw tables.

Nate-Wessel:

- removes anomalous ranges from the `ecocounter.counts` view  

(they are still in the unfiltered tables)

## Issue(s) this solves:

- closes #995 

- make sure the query is still nice and snappy

- replace existing view with new definition

### DIFF
--- a/volumes/ecocounter/views/create-view-counts.sql
+++ b/volumes/ecocounter/views/create-view-counts.sql
@@ -7,8 +7,10 @@ FROM ecocounter.counts_unfiltered
 JOIN ecocounter.flows_unfiltered USING (flow_id)
 JOIN ecocounter.sites_unfiltered USING (site_id)
 WHERE
-    flows_unfiltered.validated --is true
+    -- must be validated at the level of both site and flow
+    flows_unfiltered.validated
     AND sites_unfiltered.validated
+    -- must not intersect with any do-not-use level anomalous ranges
     AND NOT EXISTS (
         SELECT 1
         FROM ecocounter.anomalous_ranges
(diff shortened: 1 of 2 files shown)


volumes/ecocounter/views/create-view-counts.sql
```diff
@@ -5,7 +5,15 @@ CREATE VIEW ecocounter.counts AS (
         counts_unfiltered.volume
     FROM ecocounter.counts_unfiltered
     JOIN ecocounter.flows_unfiltered USING (flow_id)
-    WHERE flows_unfiltered.validated --is true
+    WHERE
+        flows_unfiltered.validated --is true
+        AND NOT EXISTS (
```
gabrielwol, line 10: I think an anti-join is a little less verbose: 
```sql
    --anti join anomalous_ranges
    LEFT JOIN ecocounter.anomalous_ranges USING (flow_id)
    WHERE
        flows_unfiltered.validated --is true
        AND anomalous_ranges.time_range IS NULL
        OR NOT(anomalous_ranges.time_range @> counts_unfiltered.datetime_bin)
```

gabrielwol (changes requested):
Thanks for proposing this! Can you please update the comment on the view as well as the readme mentioning this?
Also RE: speed, SELECTS on the view are still able to use indices on datetime_bin & flow_id ✅

gabrielwol:
Just noticed that `ecocounter.anomalous_ranges` also has both null flow_ids and null site_ids which you'll need to account for in the join (see Miovision) - or you could make `flow_id not null` in anomalous_ranges.

volumes/ecocounter/views/create-view-counts.sql
Nate-Wessel: Perhaps less verbose, but I've always found those harder to read. I like the `NOT EXISTS` because, though large, it's a single `WHERE` condition which is conceptually a lot tidier for me.

Nate-Wessel:
> Just noticed that `ecocounter.anomalous_ranges` also has both null flow_ids and null site_ids which you'll need to account for in the join (see Miovision) - or you could make `flow_id not null` in anomalous_ranges.

I was getting there. Sorry, I didn't think requesting your review on a draft PR would actually notify you until I marked it as ready for review 😬. Almost done!

Nate-Wessel:
@gabrielwol ready for re-viewing!

PR #1008, opened 2024-07-05
1007 miovision one updated alerts schema

### LINKED ISSUE #1007: Implement new Miovision Alerts Schema

gabrielwol:
- Thanks to our feedback, the API now returns alerts active over a range of time, rather than alerts active at an instant in time. 
  - Update the alerts schema to match new API (new unique `alert_id` column)
- minor updates to readme, sql scripts
- Biggest changes are in the `volumes/miovision/api/pull_alert.py` puller script
  - Add CLI option
  - Consolidate the connections / api key from DAG into just this script + use Airflow connections rather than cfg file.
  - Deal with timezones: need to pass TZs with time zones to API and then convert the response in UTC back to local TZ. 
  - move the alerts insert script to a file. 

## Issue(s) this solves:

- Closes #1007 

This change: 
  - Consolidate the connections / api key from DAG into just this script + use Airflow connections rather than cfg file.
Only issue I can think of is it will make it more difficult for a future user without airflow access to run locally.

- replace current `miovision_api.alerts` table with `miovision_api.alerts_new`


volumes/miovision/api/readme.md
```diff
@@ -137,6 +141,18 @@ In command prompt, navigate to the folder where the python file is [located](../
 
 The `--pull` and `--agg` commands allow us to run data pulling and aggregation together or independently, which is useful for when we want to check out the data before doing any processing. For example, when we are [finding valid intersection movements for new intersections](../update_intersections/readme.md#update-miovision_apiintersection_movements).  
 
+### Alerts
+
+Although it it typically run daily through the Airflow DAG [miovision_pull](../../../dags/miovision_pull.py) `pull_alerts` task, you can also pull from the Alerts API using the command line. This is  helpful for backfilling multiple dates at once. An example command is: 
+`python3 pull_alert.py run-alerts-api-cli --start_date=2024-06-01 --end_date=2024-07-01`
```
chmnata, line 147: because the connection grabbing is in the pull_alert.py instead of the dag, we should mention this command can only be run in the airflow venv.

chmnata: there is an extra space between `this is` and `helpful`

volumes/miovision/api/pull_alert.py
```diff
@@ -133,56 +131,37 @@ def pull_alerts(conn: any, start_date: datetime, end_date: datetime, key: str):
...
-
-    #pull alerts at each timestamps and append to list
+    #pull alerts from each page and append to list
     dfs = []
-    for dt in dt_list: 
-        miovpull = MiovAlertPuller(dt, key)
-        test = miovpull.get_response()
-        response = miovpull.process_response(test)
-        df = pd.DataFrame(response).drop_duplicates()
+    pageSize = 100
```
chmnata, line 141: Would pagesize change in the future?

chmnata (changes requested):

volumes/miovision/api/pull_alert.py
gabrielwol: We probably won't ever change it. The API default is 1 with maximum being 1000. I found it was much faster when set to a higher number.

chmnata: so pagesize is like a "how many alerts is shown in one page" instead of "look through 100 pages of alerts" ?

chmnata: mainly just wondering if we might miss some alerts on another page if its the latter

gabrielwol: ohhh no, the block below is iterating through the pages. See lines 143, 146.
The API returns the link to the next page so you just keep following the next link. 

Here's the schema of the response from the API docs: 

```
{
  "alerts": [
    {
      "alertId": "string",
      "alertStartDateTime": "2024-07-15T15:37:11.809Z",
      "alertEndDateTime": "2024-07-15T15:37:11.809Z",
      "intersectionId": "string",
      "type": "string"
    }
  ],
  "links": {
    "next": "string"
  }
}
```

chmnata: :ayeaye:

chmnata (approved):
:gabe-approved:

gabrielwol:
Merged and backfilled alerts to 2019-01-01.

PR #1017, opened 2024-07-18
Fix miovision example sql

chmnata:

- update sql that filters for count approaches only

## Issue(s) this solves:

- #

- that one line chnage

nothin


gabrielwol (approved):
good catch!

PR #1024, opened 2024-07-24
Miovision Aggregation Improvements (ATR view, intersection_movements AFTER INSERT trigger, unacceptable_gaps)

gabrielwol:
## What this pull request accomplishes:

1. Remove `miovision_api.volumes_15min` (table) in place of view (`volumes_15min_atr_filtered`). Reduces duplicate storage (~75GB) and complexity in aggregation. 
2. Add an `AFTER INSERT` trigger to `intersection_movements` to help aggregate and pad new movements into volumes_15min_mvt.  
    - volumes/miovision/sql/function/function-add_intersection_movement_padding_values.sql: based on existing inserts into volumes_15min_mvt, except using `new_rows` (transition relation) instead of `intersection_movements` and without date limits. 
3. Move `unacceptable_gap` removal from table inserts to filtered view: less complexity in adding and removing unacceptable_gaps. 

**Summary of files changed:**
Many files have minor changes including renamed references, removal of ATR aggregation, big fluff (intersection_movements, volumes_15min_tmc).

Files with major changes:
- [volumes/miovision/sql/function/function-add_intersection_movement_padding_values.sql](https://github.com/CityofToronto/bdit_data-sources/pull/1024/files#diff-a18eb22409f0b923d050598f95788f37425c53d02664af98a4e20d2f0b85ad57): trigger function to add values to 15min aggregate table based on additions to intersection_movements. Based closely on [existing aggregation](https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/miovision/sql/function/function-aggregate-volumes_15min_mvt.sql) but only for new intersection_movements. 
- [volumes/miovision/sql/table/create-table-movement_map_new.sql](https://github.com/CityofToronto/bdit_data-sources/pull/1024/files#diff-436002ac1093f53236f22451598bb06ff4c7eb1edf1b622c49677f6fa6f5cd48): a wide version of `movement_map` with entry/exit leg columns and directions more clearly labelled. Helps with ATR view joins. 
- [volumes/miovision/sql/views/create-view-volumes_15min_atr_filtered.sql](https://github.com/CityofToronto/bdit_data-sources/pull/1024/files#diff-a25a0ab27d992960339211265d569955314948c953008335fcd37c969f0f74de): ATR view which replaces table. 
[...]

PR #1025, opened 2024-07-26
855 monitor intersection movements for missing high volume movements

### LINKED ISSUE #855: Need to monitor intersection_movements for missing high volume movements
> None of the new intersections have volumes for classification_uid 10 (bike approaches). 
> We'll need to confirm whether this is an algorithm change or a configuration issue.
> 
> 4 of the new intersection are missing any volumes for classification_uid 6 (pedestrians): `intersection_uids IN (69, 71, 72, 90)`

_Originally posted by @gabrielwol in https://github.com/CityofToronto/bdit_data-sources/issues/850#issuecomment-1919838271_
            
Because there was no data for these classifications when we added the intersections, the movements didn't get added to `miovision_api.intersection_movements` (common intersection movements which get zero-padded in aggregate `volumes_15min_mvt` table). If these classifications eventually get added, we will need to go back and add the movements and backfill the aggregate tables.

gabrielwol:

- adds a view to find common movements in the last 100 days which are not in intersection_movements table
  - add a weekly notification to check this view (see example in `#airflow_pipelines-dev`)
- Adds a denylist: `intersection_movements_denylist` with same structure as base ("allowlist")
  - Didn't want to add this as column to existing table because it would require a lot more join changes
- Adds `BEFORE INSERT` triggers on both `intersection_movements` and `intersection_movements_denylist` to restrict inserting overlapping rows between the two. See it in action below. 
- Note there is a new `after insert` trigger (to insert padding values based on new inserts to intersection_movements) being developed in #1024 which will make it easier to make additions to intersection_movements. 

Test new triggers using: 
```sql
INSERT INTO miovision_api.intersection_movements
SELECT * FROM miovision_api.intersection_movements_denylist LIMIT 1;

INSERT INTO miovision_api.intersection_movements_denylist
SELECT * FROM miovision_api.intersection_movements LIMIT 1;
```

## Issue(s) this solves:

- Closes #855 

- ~~The two functions are basically the same, maybe I should make them take a table parameter...~~ ✅ 


volumes/miovision/sql/views/create-view-monitor_intersection_movements.sql
```diff
@@ -0,0 +1,59 @@
+CREATE OR REPLACE VIEW miovision_api.monitor_intersection_movements AS (
```
chmnata, line 1: when I select from the view, there was one row where sum_volume > intersection_classification_total which might be sus.
image

volumes/miovision/sql/function/function-exclude_denylist_from_intersection_movements.sql
```diff
@@ -0,0 +1,39 @@
+CREATE OR REPLACE FUNCTION miovision_api.exclude_denylist_from_intersection_movements()
+RETURNS trigger
+LANGUAGE plpgsql
+AS $BODY$
+
+DECLARE row_in_im numeric = (
+    SELECT COUNT(*)
+    FROM miovision_api.intersection_movements_denylist AS im
```
chmnata, line 8: is it possible to make the table name a param?

chmnata (changes requested):
just some questions on the view, possibility of using 1 function for two tables 🤔

volumes/miovision/sql/views/create-view-monitor_intersection_movements.sql
gabrielwol: Not what I exepcted! It's because `SELECT now()::date AT TIME ZONE 'Canada/Eastern' - interval ' 100 days'` results in `2024-05-21 20:00:00` so the larger value had an extra 4 hrs of data.

volumes/miovision/sql/function/function-exclude_denylist_from_intersection_movements.sql
gabrielwol: Done!

chmnata (approved):
:gabe-approved:

PR #1035, opened 2024-08-09
1032 move gcc puller tables to airflow variable

### LINKED ISSUE #1032: Move gcc puller tables to Airflow variable

gabrielwol:

- move GCC layers and Primary key dictionaries to Airflow variables
- Dynamic DAG generation for bigdata/ptc GCC layers using Airflow variable which stores credentials, deployment
- added mat view refreshes to Airflow variable: easier addition of new aggregations!

## Issue(s) this solves:

- Closes #1032 

- I couldn't think of an obvious way to make mat view aggregations their own Airflow tasks in a dynamic way so I just added them to the pull_layers task. Thoughts?

- change data_scripts link in Morbius

### DIFF
--- a/dags/gcc_layers_pull.py
+++ b/dags/gcc_layers_pull.py
@@ -19,7 +19,7 @@ except:
 
 
 # the DAG runs at 7 am on the first day of January, April, July, and October
-def create_gcc_puller_dag(dag_id, default_args, name, conn_id): 
+def create_gcc_puller_dag(dag_id, default_args, name, conn_id):
     @dag(
         dag_id=dag_id,
         default_args=default_args,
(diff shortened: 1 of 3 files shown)


chmnata (changes requested):
Can we make the mapped task's name as the layer's name? for refreshing the materialized view, I'm thinking we could use a trigger function on the table instead 🤔

dags/gcc_layers_pull.py
```diff
@@ -17,115 +17,83 @@
...
-    'start_date': pendulum.datetime(2022, 11, 3, tz="America/Toronto"),
-    'email_on_failure': False, 
-    'retries': 0,
-    'on_failure_callback': partial(task_fail_slack_alert, use_proxy=True)
-}
+            #refresh mat views as necessary
+            agg_sql = layer[1].get("agg")
+            if agg_sql is not None:
+                with conn.cursor() as cur:
+                    print(agg_sql)
```
chmnata, line 59: this for logging purpose?

chmnata (approved):
Thanks for reminding me that morbius's airflow version doesn't have mapped tasks naming yet :meow_headache: This looks good! refreshing mat view as a tasks as good as well.

dags/gcc_layers_pull.py
gabrielwol: Oh I'll remove this, it was for testing (instead of execute).

PR #1046, opened 2024-08-27
#1045 restructure to use "last_active" columns

gabrielwol:

- rewrite ongoing outage query for Ecocounter to use `last_active` columns and remove lookback limit. 

## Issue(s) this solves:

- #1045 

- 

Pull into data_scripts

### DIFF
--- a/volumes/ecocounter/data_checks/select-ongoing_outages.sql
+++ b/volumes/ecocounter/data_checks/select-ongoing_outages.sql
@@ -1,54 +1,45 @@
-WITH out_of_order AS (
+WITH ooo_sites AS (
+    SELECT
+        site_id,
+        site_description,
+        last_active
+    FROM ecocounter.sites
+    WHERE
+        last_active < CURRENT_DATE - 1
+        AND date_decommissioned IS NULL
+),
+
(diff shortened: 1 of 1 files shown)


volumes/ecocounter/data_checks/select-ongoing_outages.sql
```diff
@@ -1,54 +1,45 @@
-WITH out_of_order AS (
+WITH ooo_sites AS (
+    SELECT
+        site_id,
+        site_description,
+        last_active
+    FROM ecocounter.sites
+    WHERE
+        last_active < CURRENT_DATE - 1
```
gabrielwol, line 8: I changed to `CURRENT_DATE - 1` (recommended by postgres docs 👀) instead of basing it off of `ds`, because this way it will be more accurate if task is rerun on a different date.

chmnata (approved):
:gabe-approved:

volumes/ecocounter/data_checks/select-ongoing_outages.sql
chmnata: makes sense!

Issue #1060, opened 2024-09-04
Upgrade to Airflow 2.10


gabrielwol:

gabrielwol:
Completed by #1062

PR #1062, opened 2024-09-06
#1060 upgrade airflow 2.10.0

gabrielwol:

- Update constraints.txt to Airflow 2.10.0 and use `no-providers` version which reduces the dependencies by 500 🆒

## Issue(s) this solves:

- #1060

- 

Nothing


radumas:
Glad this helps resolve some of our dependabot issues. Wish there was a way to ensure no dependency conflict with the few providers we do use. But given that we use only a few... should be fine.

I wish there was a way of limiting the google provider to just the connections we need, looks like it's installing a loooooot of extra stuff that we aren't using.

radumas (approved):
Looks good!

PR #1071, opened 2024-09-10
#1070 use `pre_execute` to reduce # of tasks in partition create taskgroups

gabrielwol:

- switch from using `@task.short_circuit` operators to `pre_execute` = reduces # of tasks for partition creates. 
  - Originally intended to use `@task.run_if` which would have had the same impact (fewer tasks), but the partition creates are currently PostgresOperators which would have required refactoring them to @task operators. 

Tested by re-running today's `miovision_pull` DAG ✅ 

## Issue(s) this solves:

- #1070

- `test_dags` action is failing on deprecated `pull_wys` and `pull_miovision`... Should we delete these DAGs once and for all?


chmnata (approved):
:gabe-approved: :gabe-approved: :gabe-approved:

Issue #1053, opened 2024-08-29
Transform Miovision Open Issues to table


gabrielwol:
This is to allow Chris to mark them as logged.
It will be less confusing to change the view into a table than to have both.

PR #1075, opened 2024-09-17
1053 transform miovision open issues to table

gabrielwol:

- Updates `open_issues` view to be a table. 
  - Updated daily via Airflow after inserting into `anomalous_ranges`. 
  - Only updates/keeps issues which are in anomalous_ranges, otherwise deletes. 
  - Has a `logged` column which Chris can use to keep track of which ones he has seen + potential for other notes column. 
  - Adds reference to this table in one of the weekly notifications

## Issue(s) this solves:

- #1053

- 

To do: 
- [x] switch `open_issues_review` to `open_issues` and delete existing view.
- [x] Add table to readme.


volumes/miovision/sql/function/function-update-open_issues.sql
```diff
@@ -0,0 +1,130 @@
...
+        ORDER BY
+            ar.intersection_uid,
+            ar.range_start,
+            ar.classification_uid
+    ),
+
+    closed AS (
+        DELETE FROM miovision_api.open_issues_review
+        WHERE uid NOT IN (SELECT uid FROM open_issues)
+    )
```
leo-oxu, line 91: CTE `closed` seems not to be used after declaration? Or it is collaborating with the MERGE INTO lines below to do something since it's a DELETE CTE? :meow_dizzy_puffy:

gabrielwol: This `MERGE INTO` is not much different than a `INSERT ... UPDATE`. Maybe I should use that instead for readability...

leo-oxu (approved):

Issue #1064, opened 2024-09-06
Short term count useful views need to be updated


radumas:
![image](https://github.com/user-attachments/assets/5c3bd9e2-55ce-4cca-836f-a85924cabaf2)

They have been renamed to `_DO_NOT_USE` because they are using the old regular expression process and not the newer conflation process.

[This documentation](https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/short_term_counting_program/#useful-views) will also have to be updated

Issue #774, opened 2023-11-21
Useful views depending on `artery_px` are a trap


radumas:
They use regex on `arterydata` and hooo boy are there a lot of PXs missing from the `location` column in `arterydata`

Also maybe need to update the documentation when this is finished.

https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/short_term_counting_program#5-useful-views

PR #1078, opened 2024-09-30
774 useful views depending on artery px are a trap

gabrielwol:

- Fix trap view joining arterydata and PX. 

## Issue(s) this solves:

- #774, #1064

- 

- [x] To keep dependencies we should `CREATE VIEW traffic.artery_locations_px_DO_NOT_USE AS ...` and then `ALTER VIEW traffic.artery_locations_px_DO_NOT_USE RENAME TO traffic.artery_locations_px`. 
- [x] Also `DROP traffic.artery_traffic_signals_DO_NOT_USE`. 
- [x] Comment on two data requests that the view has changed.


radumas:
> To keep dependencies we should CREATE VIEW traffic.artery_locations_px_DO_NOT_USE AS ... and then ALTER VIEW traffic.artery_locations_px_DO_NOT_USE RENAME TO traffic.artery_locations_px.

Unfortunately I think postgres actually stores VIEW dependencies as references to the OIDs. So the relationship will persist whatever the name (Except for objects referenced in functions, I think).

We'll have to use the dependency saver instead.

radumas:
Code looks good. I agree with the fluff though 🐑  please fix the SQL

gabrielwol:
> > To keep dependencies we should CREATE VIEW traffic.artery_locations_px_DO_NOT_USE AS ... and then ALTER VIEW traffic.artery_locations_px_DO_NOT_USE RENAME TO traffic.artery_locations_px.
> 
> Unfortunately I think postgres actually stores VIEW dependencies as references to the OIDs. So the relationship will persist whatever the name (Except for objects referenced in functions, I think).
> 
> We'll have to use the dependency saver instead.

I was able to use `CREATE OR REPLACE VIEW ...` and `ALTER VIEW ... RENAME TO` and have the dependencies remain linked!

gabrielwol:
Implemented and ready to merge with your approval @radumas

radumas:
> > > To keep dependencies we should CREATE VIEW traffic.artery_locations_px_DO_NOT_USE AS ... and then ALTER VIEW traffic.artery_locations_px_DO_NOT_USE RENAME TO traffic.artery_locations_px.
> > 
> > 
> > Unfortunately I think postgres actually stores VIEW dependencies as references to the OIDs. So the relationship will persist whatever the name (Except for objects referenced in functions, I think).
> > We'll have to use the dependency saver instead.
> 
> I was able to use `CREATE OR REPLACE VIEW ...` and `ALTER VIEW ... RENAME TO` and have the dependencies remain linked!

Right. I didn't follow the bit about replacing the bad view. That works in line with my understanding of how things work.

radumas (approved):
Looks great! People are going to have such efficient lives now.

PR #1089, opened 2024-10-29
#1086 switch alerts to use custom operator

chmnata:

- Make silent alerts not silent 

## Issue(s) this solves:

- #1087 

- the switching to custom operator 

### DIFF
--- a/dags/bluetooth_check_readers_temp.py
+++ b/dags/bluetooth_check_readers_temp.py
@@ -60,7 +60,7 @@ def blip_pipeline():
     pipeline_check = SQLCheckOperatorWithReturnValue(
         task_id = 'pipeline_check',
         conn_id = 'bt_bot',
-        sql = '''SELECT  * 
+        sql = '''SELECT  (COUNT(*) > 0) AS "_check"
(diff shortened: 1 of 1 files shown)


dags/bluetooth_check_readers_temp.py
```diff
@@ -60,7 +60,7 @@ def blip_pipeline():
     pipeline_check = SQLCheckOperatorWithReturnValue(
         task_id = 'pipeline_check',
         conn_id = 'bt_bot',
-        sql = '''SELECT  * 
+        sql = '''SELECT  (COUNT(*) > 0) AS "_check"
```
gabrielwol, line 63: Similar to `broken_readers` sql below, can you please add a second column with a message? 
Otherwise we get an alert without any details and we have to check the logs, ie. 
`temp_bluetooth_check_readers.pipeline_check (20241113T145010.485091+0000) FAILED.
@Natalie @Gabe, please, check the logs`

gabrielwol (approved):
One tiny change and then you can merge! 
Lots of alerts now 😿

Issue #1101, opened 2024-11-22
Capture late arriving Ecocouter Data


gabrielwol:
Follow up to https://github.com/CityofToronto/bdit_data-sources/issues/984; discovered that for some Ecocounter sites (Bloor and Old Mill), not all 96-15minute bins arrive in the first data pull. This looks like because some of the uploads at this site are 1pm - 1pm. We should try repulling these partial days with an update to: https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/ecocounter/functions/create-function-identify-outages.sql

PR #1102, opened 2024-11-22
#1101 Repull days without all bins present

gabrielwol:

- Repull days without all bins present
- Tested this on the last 60 days and 100% (108 days) of partial data were filled 🥳 Will proceed with backfilling all data.

## Issue(s) this solves:

- #1101

- 


A-DUYVESTYN (approved):
looks good to me

PR #1105, opened 2024-12-02
1104 `eoy_table_create` bug fixes

gabrielwol:

- eoy_table_create DAG was targeting wrong year
  - Add a task to calculate `yr` in a more pythonic way
- Made bluetooth task more consistent; receive year instead of ds
- Add here.ta_path yearly table task

## Issue(s) this solves:

- #1104 

- 


chmnata (approved):
@gabrielwol your changes, you can check my task as well!
Hopefully next year we will only have one here task.

PR #1127, opened 2025-01-06
Define order of returned routed centreline links

### LINKED ISSUE #1126: `gis_core.get_centreline_btwn_intersections` should return links in routed order
The centreline routing function `gis_core.get_centreline_btwn_intersections` seems to return links as an array sorted by link centreline_id in ascending order. 

This is easy to calculate. It would be more helpful if the function preserved the sequence of the links in the route from start to end - something that's quite hard to determine after the fact.

Nate-Wessel:

- Defines an order for the results returned by the centreline routing function. 
- Orders aggregated results by sequence in the routed path

## Issue(s) this solves:
- Closes #1126

- That this works and nothing breaks. I'm not used to working on SQL within functions.

Since the order was previously undefined, it shouldn't break anything to specify it now.

- replace function with the new definition

### DIFF
--- a/gis/centreline/sql/create_function_get_centreline_btwn_intersections.sql
+++ b/gis/centreline/sql/create_function_get_centreline_btwn_intersections.sql
@@ -30,8 +30,8 @@ WITH results AS (
 SELECT
     get_centreline_btwn_intersections._node_start,
     get_centreline_btwn_intersections._node_end,
-    array_agg(routing_centreline_directional.centreline_id),
-    st_union(st_linemerge(routing_centreline_directional.geom)) AS geom 
+    array_agg(routing_centreline_directional.centreline_id ORDER BY path_seq),
+    st_union(st_linemerge(routing_centreline_directional.geom ORDER BY path_seq)) AS geom 
 FROM results
 INNER JOIN gis_core.routing_centreline_directional ON edge = id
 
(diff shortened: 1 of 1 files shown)


gis/centreline/sql/create_function_get_centreline_btwn_intersections.sql
```diff
@@ -30,8 +30,8 @@ WITH results AS (
 SELECT
     get_centreline_btwn_intersections._node_start,
     get_centreline_btwn_intersections._node_end,
-    array_agg(routing_centreline_directional.centreline_id),
-    st_union(st_linemerge(routing_centreline_directional.geom)) AS geom 
+    array_agg(routing_centreline_directional.centreline_id ORDER BY path_seq),
+    st_union(st_linemerge(routing_centreline_directional.geom ORDER BY path_seq)) AS geom
```
gabrielwol, line 34: I get this error while testing: 
`ERROR:  ORDER BY specified, but st_linemerge is not an aggregate function`
Need to remove the order by in this line.

gabrielwol (changes requested):
Thanks for suggesting! One issue to resolve.

Nate-Wessel:
Fixed! From a quick test, it does appear that order is preserved within the multi-geometries, making the link IDs and geometries accessible as pairs with an unnest/st_dump combo.

gabrielwol (approved):

PR #1150, opened 2025-02-14
Summarize all SVC types as daily total counts per segment

Nate-Wessel:
Just offering this here as a suggestion, but would love to totally hand this off to Data Ops to talk through how to do this properly. 

- Sketches out a view of the various SVC types (speed, volume, classification) that summarizes them as daily totals per centreline segment

## Issue(s) this solves:

- have to look through three tables to get consistent volume data
- not clear that most of the data is in speed, not volume

- sums are correct (are any of the classification categories not mutually exclusive?)
- can probably find a more efficient/scaleable way of doing this

* create the view, etc


Nate-Wessel:
The query as it is is breaking down in more recent years because the `count_date` doesn't align with the dates in the time bins. 

```sql
SELECT
    EXTRACT(YEAR FROM count_date) AS yr,
    COUNT(*) FILTER (WHERE count_date = time_start::date) AS date_matches,
    COUNT(*) FILTER (WHERE count_date != time_start::date) AS not_matches
FROM traffic.svc_study_speed
GROUP BY yr
ORDER BY yr
```

volumes/short_term_counting_program/svc_daily_totals.sql
```diff
@@ -0,0 +1,79 @@
...
+            + vol_66_70_kph
+            + vol_71_75_kph
+            + vol_76_80_kph
+            + vol_81_160_kph
+        ) AS daily_volume
+    FROM traffic.svc_study_speed
+    GROUP BY
+        study_id,
+        count_date
+    HAVING COUNT(*) = 2 * 4 * 24
```
gabrielwol, line 28: There appear to be a lot of counts with one direction only. I think we should add direction as a group by field and remove the `2 *` from Having.

volumes/short_term_counting_program/svc_daily_totals.sql
```diff
@@ -0,0 +1,79 @@
+CREATE OR REPLACE VIEW traffic.svc_daily_totals AS
+
+WITH daily_totals AS (
+    -- speed studies
+    SELECT
+        study_id,
+        count_date,
```
gabrielwol, line 7: As you discovered, we should use `time_start::date` for svc_study_speed, at least until it gets patched by MOVE

volumes/short_term_counting_program/svc_daily_totals.sql
```diff
@@ -0,0 +1,79 @@
...
+    HAVING COUNT(*) = 2 * 4 * 24
+
+    UNION
+
+    -- classification studies
+    SELECT
+        study_id,
+        count_date,
+        SUM(
+            -- check that these are mutually exclusive
```
gabrielwol, line 50: I think you're safe there. Each of these columns refers to an FHWA class: https://move-etladmin.intra.prod-toronto.ca/docs/database_schema.html#atr.table.study-class-human

volumes/short_term_counting_program/svc_daily_totals.sql
gabrielwol: A lot of (mostly) old FlowLoad counts are 1 directional: 
```sql
SELECT
    svc_metadata.study_id,
    study_source,
    svc_metadata.study_start_date,
    svc_metadata.centreline_id, 
    COUNT(DISTINCT svc_study_speed.direction) AS count_directions,
    ARRAY_AGG(DISTINCT svc_study_speed.direction)
FROM traffic.svc_study_speed
JOIN traffic.svc_metadata USING (study_id)
GROUP BY 1, 2, 3, 4
HAVING COUNT(DISTINCT svc_study_speed.direction) < 2
ORDER BY svc_metadata.study_start_date DESC
```

gabrielwol (commented):
Seems useful. I wonder if it would be better to separate out your CTE into a view that simply unions the 3 study types 15 minute volumes, and then another view that aggregates up to daily.

Nate-Wessel:
> Seems useful. I wonder if it would be better to separate out your CTE into a view that simply unions the 3 study types 15 minute volumes, and then another view that aggregates up to daily.

That makes sense to me. Just converting them to volumes from their various categories would definitely be a helpful first step.

volumes/short_term_counting_program/svc_daily_totals.sql
Nate-Wessel: Makes sense. Would be easy enough to aggregate up from there.

gabrielwol: Added direction as a group by.

gabrielwol:
I made the requested changes and also added indices on the relevant date columns to speed up queries with date filters, eg: 

```sql
--0.5s for selecting studies last month now
SELECT * FROM gwolofs.traffic_svc_daily_totals WHERE count_date >= '2025-01-01' AND count_date < '2025-02-01'
```

Nate-Wessel:
This looks great! Thanks @gabrielwol!

I'll make a note to refactor some of my code to rely on this.

gabrielwol (approved):
Merging with Nate's approval!

Issue #1149, opened 2025-02-13
Add geojson export to Ecocounter Open Data


gabrielwol:

gabrielwol:
We need to install `ogr2ogr` on Morbius (requires root). 
Could also use a python approach like `geopandas` recommended by Peter, which we would need to add to vfh dev requirements.

PR #1151, opened 2025-02-18
1149 add geojson export to ecocounter open data

gabrielwol:

- Create a view to format open data locations as geojson in postgres
- Add new task to export the view

## Issue(s) this solves:

- #1149 

- 


chmnata (commented):
does the geojson need a `properties` field? Can't get the geojson to dispplay on geojson.io

gabrielwol:
> does the geojson need a `properties` field? Can't get the geojson to dispplay on geojson.io

Weird, it opened fine in QGIS and [https://geojson.tools](https://geojson.tools), but looks like some tools require `properties` where I used `attributes`. I'll change that.

gabrielwol:
I made the `attributes` -> `properties` change and it works in all 4 tools I tried. I created a Jira ticket mentioning we want the map displayed as a preview on the landing page.

chmnata (approved):

PR #1155, opened 2025-02-21
#1154 Centreline legs cardinal directions

### LINKED ISSUE #1154: Automate determining legs & direction for intersections
Relevant to collision safety work, count conflation, many applications...

Could be used as a building block for #959 linear volume view and #1117 

Should live in `gis_core` and be built on the relevant `*_latest` Mat Views. We can extend this approach to the partitioned versions of the centreline and intersections where/if relevant.

Nate-Wessel:

- Identifies intersection legs by cardinal direction to make them conflatable with TMC and collision data, etc which identify e.g. a "north leg" etc. This maps "north" to a centreline_id and geometry.

## Issue(s) this solves:

- Resolves #1154 

- That the code makes sense and it actually does what the comments describe
- That it provides intuitively correct results in the vast majority of cases

Create the materialized view


Nate-Wessel:
@radumas this is ready; who should review?

radumas:
Thanks Nate!

I skimmed the code. Great work with the comments!

Does this currently live in the database in your schema? Did you say it took on the order of a minute to create?

I've let @peter-lyons know about this, as there's synergy here. Is this in line with what you all had been thinking? Doesn't seem like it would be too difficult to extend to 8 cardinal directions.

We can create a new issue for dealing with the 60-odd 4+ legged intersections.

Also flagging for @chmnata to add a TRIGGER to refresh the mat view on implementation.

Nate-Wessel:
> Does this currently live in the database in your schema? Did you say it took on the order of a minute to create?

I just created it in my schema if you want to take a look. It ran in a bit under 6 seconds. 
`nwessel.centreline_leg_directions`

mkewins:
@lowsound42 your review is requested on this PR. Can you please review Nate's algorithm and compare with the MOVE algorithm?

From your investigations in the MOVE-1498 bug, here's where the MOVE algorithm [starts](https://github.com/CityofToronto/bdit_flashcrow/blob/dbedc86ceac215337395af01a12c0237556d5afe/lib/reports/ReportBaseFlowDirectional.js#L328) and [selects cardinal directions for intersection legs](https://github.com/CityofToronto/bdit_flashcrow/blob/1a28972424243c80226a3c1166bc43da44877af8/lib/geo/GeometryUtils.js#L181).

gis/centreline/sql/legs.sql
```diff
@@ -0,0 +1,280 @@
...
+    FROM node_edges
+    GROUP BY node_id
+    HAVING COUNT(DISTINCT edge_id) > 2
+),
+
+legs AS (
+    /*
+    In this step we find the legs of all the intersections
+    (either inbound or outbound) and measure their bearing
+    or compass direction of travel *away* from the reference
```
lowsound42, line 43: @Nate-Wessel is this the same as meaning the clockwise direction?

Nate-Wessel: Outbund from the intersection, rather than inbound which would be 180 degrees off from that. The direction of measurement, and the start/0 point for the angle calculations are handled consistently by `st_azimuth` throughout.

lowsound42: That makes sense, I totally misunderstood 😆

lowsound42:
@Nate-Wessel [this](https://github.com/CityofToronto/bdit_flashcrow/blob/master/lib/geo/GeometryUtils.js) is the file in which we do our direction assignments. I think [here](https://github.com/CityofToronto/bdit_flashcrow/blob/dbedc86ceac215337395af01a12c0237556d5afe/lib/geo/GeometryUtils.js#L69) is where we do a bearing calculation comparable to your azimuth bearing bit and [here](https://github.com/CityofToronto/bdit_flashcrow/blob/dbedc86ceac215337395af01a12c0237556d5afe/lib/geo/GeometryUtils.js#L52) is where we do the bit that corresponds to that cross join you are doing.

I think for the most part our algorithm is doing the same general steps as yours but I think that maybe this part where we initially get the bearings (first link) might slightly differ because it's using that great bearing method vs yours which is using postgis' azimuth function (maybe those do the same thing??).

Would you be able to compare these and see if we're doing a different thing altogether? If so, I wonder if that coupled with the fact that you're doing the 17 degree rotation may end up getting different results for some intersections.

Nate-Wessel:
> ...think that maybe this part where we initially get the bearings (first link) might slightly differ because it's using that great bearing method vs yours which is using postgis' azimuth function (maybe those do the same thing??).

In theory, the `st_azimuth` method should be getting something equivalent to a great circle method. It's doing spherical/spheroidal angles because it's taking the geography type rather than geometry. 

The bigger difference I see is in (the 17 degree offset, yes, but also) how the legs are assigned. Just from reading the code comments, the MOVE method is described as 

>  To calculate this, we compute the bearing at which each `LineString` leaves `point`. If this bearing is within 45 degrees of a cardinal direction, we say that it lies in that direction.  Of all features that lie in a cardinal direction, the one that has an bearing closest to that direction is chosen as the "best directional candidate".

The method proposed here doesn't have that 45% constraint. It assigns the legs iteratively, one at a time starting with the best match. 

So yeah, I'd say it's kind of a different approach altogether. I'd definitely expect it to get different results in some cases. Is there any way of seeing the results on a map (or elsewhere) for the MOVE algorithm? How could we find the differences to compare?

Nate-Wessel:
@radumas can this get merged and implemented? I'd like to start building on this and it would be great to point to something not in my personal schema for that work.

radumas:
I'm OK with the approach proposed. We'll find out what counts & intersections don't follow the pattern once it's implemented.

What needs to be done for this to be merged:

- [ ] Add a comment on the VIEW to report any issues in https://github.com/CityofToronto/bdit_data-sources/issues/1190
- [ ] Add a bit of description of the view under [`master/gis/centreline`](https://github.com/CityofToronto/bdit_data-sources/tree/master/gis/centreline) to document its existence. 
- [ ] Add a bullet under [Short Term Counting Program Useful Views](https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/short_term_counting_program#useful-views) where this might be most relevant for converting TMC data to linear volumes on adjacent streets
- [ ] @chmnata to add a TRIGGER when implementing to refresh the mat view automatically.

Nate-Wessel:
Thanks @radumas. I'll add the documentation. 

I'd like to also make one last minor tweak to the geoms output from this view. I want to orient the full centreline geometries relative to the intersection, same as for the stub geometries. I'm realizing that it's going to be very tricky to do that downstream of the current view as it stands now.

gabrielwol:
> @chmnata to add a TRIGGER when implementing to refresh the mat view automatically.

I can contribute the update trigger to this branch

Nate-Wessel:
All asks have been accomplished. This should be ready to merge.

radumas (approved):
Awesome! Thanks everyone!

radumas:
Squashing because none of the commit messages have an issue number. In the future will require history with issue numbers.

PR #1169, opened 2025-03-05
#1168 Check for missing miovision centreline_ids

### LINKED ISSUE #1168: 13 centrelines referenced from `miovision_api.centreline_miovision` are no longer in `centreline_latest`
They exist in the history of the centreline table (2023-04-01) so they were probably correct originally, but should now be updated. 

```sql
WITH unmatched_centrelines AS (
    SELECT centreline_id
    FROM miovision_api.centreline_miovision
    LEFT JOIN gis_core.centreline_latest USING (centreline_id)
    WHERE centreline_latest.version_date IS NULL
)

SELECT DISTINCT ON (centreline_id)
    centreline.*
FROM gis_core.centreline
JOIN unmatched_centrelines USING (centreline_id)
ORDER BY
    centreline_id,
    version_date ASC
```

gabrielwol:

- add notification for missing / outdated miovision centreline ids
- See successful test: https://bditto.slack.com/archives/C04BWRWVB3L/p1741213251358929

## Issue(s) this solves:

- #1168 

- 


radumas (changes requested):
Looking pretty good. I like the combination of automatic conflation and alerting. I have one comment on the DAG itself.

I'm not entirely following the manual `NULL` exception for intersections that won't automatically conflate, and feeling like it should be more explicitly spelled out in the docs. My understanding of the troubleshooting is that it's pointing at the original script to assign centreline legs. Are the manual `NULL`s explained there?

gabrielwol:
This is ready to re-review and merge.

radumas (approved):
Awesome! Thanks Gabe!

PR #1171, opened 2025-03-10
#1159 basic data quality checks HERE

### LINKED ISSUE #1159: Basic Data Quality Checks: HERE
- [ ] check probe amount
- [ ] check tti variation (snow day is gonna be there)
- [ ] num rows

### LINKED ISSUE #1206: Expand use of `AFTER INSERT... FOR EACH STATEMENT` trigger functions to reduce data checks query time
https://github.com/CityofToronto/bdit_data-sources/blob/38a9c1d238ff0920f304114aa8113f75e9dbe9ee/volumes/ecocounter/tables/counts_unfiltered.sql#L32-L36

Similarly to the [trigger function used by Ecocounter](https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/ecocounter/functions/function-fn_update_last_active.sql) to update `last_active` column in a summary table with each insert into raw data table (`counts_unfiltered`), we should use this methodology for other datasets to reduce the query time for data checks.
For example, `pull_here.data_checks.check_distinct_link_dirs` is taking 24 minutes to find distinct link_dirs from the last 30 days. This information could be summarized on insert into here.ta using a trigger, so we wouldn't have to scan a large portion of the raw table each day. 

This also applies to Miovision (>2 minutes to check row counts for 60 days), WYS (60 seconds for row count check).

gabrielwol:

- adds (existing) row count and link_dir (sensor count) checks to `here.ta` and `here.ta_path`
  -  Needed to add a `ds_offset` variable to all the DAGs which use this script, since HERE runs with for `ds-1` and other pipelines which use this sql run for `ds`. 

## Issue(s) this solves:

- #1159 

- 0.7 threshold, 30 day lookback OK?
- Natalie had the idea we should monitor for row count/sample going up as well... one issue I have with that is we would get the notification many days running since it's based on a lookback. As opposed to an outage, which hopefully we'd only get when it's relevant.


dags/pull_here.py
```diff
@@ -104,7 +106,48 @@ def trigger_dags(**kwargs):
...
+            retries=0,
+            params=data_check_params | {
+                    "id_col": "link_dir"
+                },
+        )
+        check_distinct_link_dirs.doc_md = '''
+        Compare the count of link_dirs appearing in today's pull vs the lookback period.
+        '''
+
+        wait_for_weather_timesensor() >> [
```
chmnata, line 146: weather timesensor?

gabrielwol: This is for the early AM pipelines so that when they publish their data quality checks it includes the weather (which we only get at 10am I think). But it doesn't need to be here because this pipeline runs 2 days late. Will remove.

gabrielwol:
Paused this DAG (reverted to version from `master`) because it is unapproved and `check_distinct_link_dirs` was running for an average of 30+ minutes per day with a sequential scan. `check_row_counts` was a more reasonable ~4 minutes.
image

Issue #1162, opened 2025-02-27
Data quality check: OpenData


gabrielwol:
- [ ] freshness
- [ ] row counts
- [ ] page views from: https://open.toronto.ca/dataset/open-data-web-analytics/

PR #1181, opened 2025-03-14
#1162 Open Data - data checks DAG

gabrielwol:

- New slack alert for Open Data freshness: 🎉
 {9579B36E-6955-4F22-9C57-6F727E29BEE8}

## Issue(s) this solves:

- #1162

- I have defaults for freshness, should I instead have that in the Airflow variable?

- update data_scripts


chmnata:
Yea I think the only thing that we would want to monitor is freshness, looks good like this

chmnata (approved):
✅✅✅

Issue #1140, opened 2025-01-29
Create Centreline intersection categories based on the connected links categories


radumas:
E.g. `Major / Minor`, possibly include separate column(s) for "highest category" based on the hierarchy of functional classes.

Some exploration required. 

Some discussion with people who might find this useful is warranted. Should we include number of legs too? Automate identifying whether it is an offset intersection?

radumas:
Some thinking around identifying multi-level "pseudo" intersections, in particular for highways, where any arterial is going to create an "intersection" (of two lines) but the roads do not connect. Using elevations might be the solution.

PR #1182, opened 2025-03-14
Add new intersection view and update readme

chmnata:

- Add a new intersection view that has information on connected centreline's road classes, its number of elevation, etc. 

## Issue(s) this solves:

- #1140 

- logic

Create the view in bigdata

### DIFF
--- a/gis/centreline/readme.md
+++ b/gis/centreline/readme.md
@@ -80,3 +80,15 @@ Intersections are stored in either of two tables, each of which is copied from a
     - **Not unique** on `intersection_id`: appears to be 1 row to describe every relationship between edges at a node.  
     - contains additional elevation information such as elevation level, elevation unit, height restriction, etc
     - does not include cul-de-sacs, overpass/underpass
+
+* `gis_core.intersection_degree`
+    - A view that provide information on intersection's related road classes, and connectivity degree.
+    | Column Name            | Description  |
+    |------------------------|-------------|
+    | `intersection_id`      | Unique identifier for each intersection |
+    | `feature_desc_list`    | A list of all road classes associated with the intersection, ordered by road order |
+    | `highest_order_feature` | The highest order of road class associated with the intersection. |
+    | `number_of_elevations` | Number of elevations this intersection has |
+    | `degree`              | The number of centreline segments connected to this intersection. |
+    | `centreline_list`     | An array of connected centreline_id. |
+    | `geom`            | Geometry of the intersection. |
--- a/gis/centreline/sql/create_view_intersection_degree.sql
+++ b/gis/centreline/sql/create_view_intersection_degree.sql
@@ -1,0 +1,63 @@
+CREATE OR REPLACE VIEW gis_core.intersection_degree AS
+WITH all_intersections AS (
+    SELECT 
+        intersections.intersection_id,
+        intersections.classification_desc,
+        cent.feature_code_desc,
+        intersections.number_of_elevations,
+        cent.centreline_id,
+        cent.linear_name_full,
+        cent.geom,
+        intersections.geom,
+        cent.feature_code
+    FROM gis_core.centreline_latest cent
+    JOIN gis_core.centreline_intersection_point_latest intersections 
+        ON cent.from_intersection_id = intersections.intersection_id
+    UNION
+    SELECT 
+        intersections.intersection_id,
+        intersections.classification_desc,
+        cent.feature_code_desc,
+        intersections.number_of_elevations,
+        cent.centreline_id,
+        cent.linear_name_full,
+        cent.geom,
+        intersections.geom,
+        cent.feature_code
+    FROM gis_core.centreline_latest cent
+    JOIN gis_core.centreline_intersection_point_latest intersections 
+        ON cent.to_intersection_id = intersections.intersection_id
+)
+
+SELECT 
+    dist_feature.intersection_id,
+    array_agg(dist_feature.feature_code_desc ORDER BY dist_feature.feature_code) AS feature_desc_list,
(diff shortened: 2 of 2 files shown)


gis/centreline/sql/create_view_intersection_degree.sql
```diff
@@ -0,0 +1,74 @@
...
+        intersections.number_of_elevations,
+        cent.centreline_id,
+        cent.linear_name_full,
+        cent.geom AS centreline_geom,
+        intersections.geom AS intersection_geom,
+        cent.feature_code
+    FROM gis_core.centreline_latest AS cent
+    JOIN gis_core.centreline_intersection_point_latest AS intersections
+        ON cent.from_intersection_id = intersections.intersection_id
+    UNION
```
gabrielwol, line 16: I was thinking about union vs union all here and I found that the duplicate records are indeed (100) centrelines that legitimately start and end at the same node 🤯 
{A015EA9A-FE24-4488-A7B5-B711382E36B1}
We should use `UNION ALL` to capture them, since they represent 2 distinct legs despite being the same centreline.

gis/centreline/sql/create_view_intersection_degree.sql
```diff
@@ -0,0 +1,74 @@
...
+        ORDER BY dist_feature.feature_code
+    ))[
+        1
+    ] AS highest_order_feature,
+    dist_feature.number_of_elevations,
+    array_length(centreline_agg.centreline_list, 1) AS degree,
+    centreline_agg.centreline_list,
+    dist_feature.geom
+FROM (
+    SELECT DISTINCT
```
gabrielwol, line 57: I think we want to keep duplicates here so that the `feature_desc_list` includes duplicate values, ie. `{Local, Local, Minor Arterial, Minor Arterial}`, that way we could make decisions based on how many of each type of leg there are if we wanted to, not just the highest order leg.

gis/centreline/sql/create_view_intersection_degree.sql
```diff
@@ -0,0 +1,74 @@
...
+    array_agg(
+        dist_feature.feature_code_desc
+        ORDER BY dist_feature.feature_code
+    ) AS feature_desc_list,
+    (array_agg(
+        dist_feature.feature_code_desc
+        ORDER BY dist_feature.feature_code
+    ))[
+        1
+    ] AS highest_order_feature,
```
gabrielwol, line 51: Clever! I thought you were over-complicating this at first but I couldn't find an alternative

gabrielwol (changes requested):

mkewins:
I can't speak to the specific queries, but I think the format of the table makes good sense.

How is elevation (z-axis) considered? Will an underpass/overpass have all segments adjoining the intersection, even though the segments might be at different elevations?

gis/centreline/sql/create_view_intersection_degree.sql
chmnata: Yea I see how that can be helpful.

chmnata:
> I can't speak to the specific queries, but I think the format of the table makes good sense.
> 
> How is elevation (z-axis) considered? Will an underpass/overpass have all segments adjoining the intersection, even though the segments might be at different elevations?

I grabbed the number of elevation info from the intersection, but I now realize it also counts non road like centrelines... like river, which I don't think we would want to count that as 2 elevations.  e.g. if a road intersects the river, the number of elevation should still be 1, if a road intersects with a bridge (which is drive able) then it should be 2.

chmnata:
> > I can't speak to the specific queries, but I think the format of the table makes good sense.
> > How is elevation (z-axis) considered? Will an underpass/overpass have all segments adjoining the intersection, even though the segments might be at different elevations?
> 
> I grabbed the number of elevation info from the intersection, but I now realize it also counts non road like centrelines... like river, which I don't think we would want to count that as 2 elevations. e.g. if a road intersects the river, the number of elevation should still be 1, if a road intersects with a bridge (which is drive able) then it should be 2.
> 
> ```
> select * from gis_core.centreline_intersection_point_latest 
> where intersection_id = 13441583
> ```
> 
> ```
> select * from gis_core.centreline_latest 
> where from_intersection_id = 13441583 or to_intersection_id = 13441583
> ```

Can't really think of a way to do it unless we also pull stuff like river and then exclude that after.... hmmmmmmmmmmmmm

radumas:
Some notes of caveats from the presentation. Because Steeles is the northern border of the City and streets north of Steeles don't exist in the centreline:

* Steeles degrees wrong
* Steeles 3-way intersections with a street to the North aren’t centreline intersections

gis/centreline/readme.md
```diff
@@ -78,5 +78,24 @@ Intersections are stored in either of two tables, each of which is copied from a
...
+    | `distinct_feature_desc_list`| Disintct list of unique road class descriptions associated with the intersection. |
+    | `highest_order_feature`     | The highest-order road class associated with the intersection. |
+    | `all_feature_code_list`     | Full list of all road class descriptions (including duplicates). |
+    | `road_names`                | List of distinct road names connected at the intersection. |
+    | `degree`                    | Number of connected centreline segments. |
+    | `centreline_ids`            | Array of `centreline_id`s connected to the intersection. |
+    | `geom`                      | Intersection geometry. |
+    | `cent_geom`                 | Combined geometry of all associated centreline segments. |
+
+    - Known Caveats:
```
gabrielwol, line 99: Can you add:
- Intersections where centrelines intersect with themselves (e.g. North Hills Terrace) are not included, since only the number of unique centreline_ids is considered for the degree.

gabrielwol (approved):
Approved! with one comment to add as a caveat

Issue #1183, opened 2025-03-17
Regex pattern `'\(.*\)'` in `gis.abbr_street` incorrectly matches everything after a bracket


gabrielwol:
```sql
SELECT
val.input,
regexp_REPLACE(val.input, '\(.*\)', '', 'g') AS incorrect,
regexp_REPLACE(val.input, '\(([^\(]+)\)', '', 'g') AS correct
FROM (values('Between Scotswood Road (East Intersection) And Scotswood Road (West Intersection)')) AS val(input)
```

PR #1184, opened 2025-03-17
#1183 fix brackets regex in function-abbr_street.sql

gabrielwol:

Tested: 
```sql
SELECT
val.input,
regexp_REPLACE(val.input, '\(.*\)', '', 'g') AS incorrect,
regexp_REPLACE(val.input, '\(([^\(]+)\)', '', 'g') AS correct
FROM (values('Between Scotswood Road (East Intersection) And Scotswood Road (West Intersection)')) AS val(input)
```

## Issue(s) this solves:

- #1183

- 


chmnata (approved):
works!

PR #1191, opened 2025-03-27
#1172 adjust volumes_15min_mvt_filtered for null ar.intersection_uid case

### LINKED ISSUE #1172: Miovision `*_filtered` views miss anomalies effecting all intersections
In the anomalies table, it's possible for `intersection_uid` to be null, meaning the anomaly should be interpreted as applying to _all_ intersections. The joins [here](https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/miovision/sql/views/create-view-volumes_15min_filtered.sql#L13) and [here](https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/miovision/sql/views/create-view-volumes_15min_mvt_filtered.sql#L14) miss this an should also join where the `intersection_uid` may be null, as is done for the other join conditions. 

vis
```sql
SELECT * 
FROM miovision_api.anomalous_ranges
WHERE intersection_uid IS NULL
```

gabrielwol:
- Adjusts `volumes_15min_mvt_filtered` for null intersection_uid case in anomalous_ranges. Right now we only use this to caution against using data from before 2019 for certain modes. 

## Issue(s) this solves:

- Closes #1172

- 


Nate-Wessel (approved):
Looks good to me! Thanks

Issue #1158, opened 2025-02-25
gcc_puller fails to pull new layer: `toinview_program_line` with IndexError


gabrielwol:
Input: `{'toinview_program_line': {'is_audited': True, 'layer_id': 47, 'mapserver': 12, 'pk': 'objectid', 'schema_name': 'gis', 'include_additional_feature': False}}`

Error:
```
  File "/data/home/nchan6/bdit_data-sources/gis/gccview/gcc_puller_functions.py", line 253, in line
    return 'SRID=4326;LineString('+','.join(' '.join(str(x) for x in tup) for tup in geom['paths'][0]) +')'
IndexError: list index out of range
```

gabrielwol:
I suspect it is failing on the null geometries in that partition: 
https://gis.toronto.ca/arcgis/rest/services/cot_geospatial12/FeatureServer/47/query?where=1%3D1&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&distance=&units=esriSRUnit_Foot&relationParam=&outFields=objectid%2C+PROG_NAME&returnGeometry=true&maxAllowableOffset=&geometryPrecision=&outSR=&having=&gdbVersion=&historicMoment=&returnDistinctValues=false&returnIdsOnly=false&returnCountOnly=false&returnExtentOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&multipatchOption=xyFootprint&resultOffset=30000&resultRecordCount=&returnTrueCurves=false&returnExceededLimitFeatures=false&quantizationParameters=&returnCentroid=false&sqlFormat=none&resultType=&f=html
![Image](https://github.com/user-attachments/assets/6734f88f-eb34-49de-8742-6a0185202274)

gabrielwol:
Also failed on April 1 with: `psycopg2.errors.InvalidTableDefinition: multiple primary keys for table "_toinview_program_line" are not allowed`

chmnata:
Thats probably cause the previously temporary table still exists, I dropped it

chmnata:
Do you know what this layer is for?

gabrielwol:
I just added it last month. I am investigating it as a source of construction/closure data!

PR #1193, opened 2025-04-01
#1158 allow null geometries in gcc puller

gabrielwol:

- allow null geometries in gcc puller

## Issue(s) this solves:

- #1158 

- I couldn't figure out an arcserver WHERE clause to filter out null geometries, but adding a try..except worked. 


chmnata (approved):

gabrielwol:
Pulled changes on Bancroft and Morbius ✅

PR #1212, opened 2025-05-06
Add from/to "intersection" node IDs to SVC directions view

Nate-Wessel:

- Adds from/to centreline node IDs to the directed SVC view to make it easier to join back to a fuller representation of the centreline network. 

## Issue(s) this solves:

- Would have had to join on a geom field, which seems sketchy

- the code

- recreate the matview (can drop AADT dependents)

### DIFF
--- a/volumes/short_term_counting_program/sql/create-view-svc_centreline_directions.sql
+++ b/volumes/short_term_counting_program/sql/create-view-svc_centreline_directions.sql
@@ -28,13 +28,21 @@ WITH to_cardinal (bearing, direction) AS (
 
 SELECT
     cl.centreline_id,
+    CASE
+        WHEN sq.angular_distance > radians(90) THEN to_intersection_id -- reversed
+        ELSE cl.from_intersection_id
+    END AS from_node,
+    CASE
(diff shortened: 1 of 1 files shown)


volumes/short_term_counting_program/sql/create-view-svc_centreline_directions.sql
```diff
@@ -28,13 +28,21 @@ WITH to_cardinal (bearing, direction) AS (
...
+    END AS to_node,
     sq.direction,
     CASE
-        WHEN sq.angular_distance > radians(90) THEN ST_Reverse(cl.geom)
+        WHEN sq.angular_distance > radians(90) THEN ST_Reverse(cl.geom) -- reversed
         ELSE cl.geom
     END AS centreline_geom_directed,
     CASE
-        WHEN degrees(sq.angular_distance) > 90 THEN (180 - degrees(sq.angular_distance))::real
+        WHEN sq.angular_distance > radians(90) THEN (180 - degrees(sq.angular_distance))::real -- reversed
```
Nate-Wessel, line 45: Changed for consistency with the other CASE blocks

PR #1214, opened 2025-05-09
1080 automagically add miovision intersections

### LINKED ISSUE #1080: Miovision intersections - non counting
We have some new Miovision cameras on the Queensway that are being used for detection and not counts. DC would like them to be brought into `miovision_api.intersections`, however we will need to do so in such a way that they don't trigger outage alerts and don't add any unnecessary bloat (via cross joins to intersections table).

gabrielwol:

- Automatically adds to `intersections` and notifies via slack about new Miovision intersections (borrowed code from similar Ecocounter feature 🎉 )
- `miovision_api.intersections.intersection_uid` was changed into a serial so that new values could be added automatically when inserting
- Update readme for DAG and adding new intersections
- Commit existing `active_intersections` view + modify to exclude not yet installed intersections
 
## Issue(s) this solves:

- #1080 

- 


chmnata (approved):
:gabe-approved:

Issue #1190, opened 2025-03-26
Issues with `gis_core.centreline_leg_directions`


radumas:
We know, because of the complexity of Toronto's streets, that the automated identification of cardinal directions of legs by intersection for 3 & 4-legged intersections isn't going to be 100%. Additionally, there may be other discrepancies due to how the count contractor conceptualized of cardinality for particular counts/intersections.

Use this issue to document any mismatches between the view and our turning movement count data. Please be specific. 

From this feedback we may identify manual exceptions, counts to rectify, or possible tweaks to the algorithm.

gabrielwol:
We found there are some intersections with same from and to intersection_ids that are legitimately different legs. I think that means the UNION here should be a UNION ALL so we capture duplicates. 

https://github.com/CityofToronto/bdit_data-sources/blob/f24c5873de851a2d8d3961da149637dd4da0a088/gis/centreline/sql/create_matview_centreline_leg_directions.sql#L17-L27

```sql
SELECT * From gis_core.centreline_latest WHERE to_intersection_id = from_intersection_id
```

![Image](https://github.com/user-attachments/assets/153ade82-95a3-41bd-be8c-d7d1708847fe)

Nate-Wessel:
The view classifies the legs for intersections that don't actually intersect in reality.

![Image](https://github.com/user-attachments/assets/a2fa00fe-cbf7-4346-80fc-13379ab7994d)

This is a problem for all highway crossings and bridges as far as I can tell.

radumas:
@Nate-Wessel could `gis_core.intersection_classification` help with the filtering?

radumas:
> We found there are some intersections with same from and to intersection_ids that are legitimately different legs. I think that means the UNION here should be a UNION ALL so we capture duplicates. **Edit:** Actually I guess these are the same centreline_ids so there'd be no way to differentiate them...
> 
> [bdit_data-sources/gis/centreline/sql/create_matview_centreline_leg_directions.sql](https://github.com/CityofToronto/bdit_data-sources/blob/f24c5873de851a2d8d3961da149637dd4da0a088/gis/centreline/sql/create_matview_centreline_leg_directions.sql#L17-L27)
> 
> Lines 17 to 27 in [f24c587](/CityofToronto/bdit_data-sources/commit/f24c5873de851a2d8d3961da149637dd4da0a088)
>  SELECT 
>      centreline_id AS edge_id, 
>      from_intersection_id AS node_id 
[...]

PR #1216, opened 2025-05-14
Filter some intersections from leg direction assignment

Nate-Wessel:

- in reference to https://github.com/CityofToronto/bdit_data-sources/issues/1190#issuecomment-2881526992

## Issue(s) this solves:

- Directions are assigned to intersections that are either 
    - not real (overpasses)
    - highway ramp connections with very tight/wide angles

In neither case would we do a TMC or assign some value to a leg. 

- 

Recreate the `gis_core.centreline_leg_directions` view.

### DIFF
--- a/gis/centreline/sql/create_matview_centreline_leg_directions.sql
+++ b/gis/centreline/sql/create_matview_centreline_leg_directions.sql
@@ -32,6 +32,12 @@ nodes AS (
     -- i.e. a legit intersection with three or more legs
     SELECT node_id
     FROM node_edges
+    JOIN gis_core.centreline_intersection_point_latest AS p
+        ON node_edges.node_id = p.intersection_id
+    WHERE p.classification NOT IN (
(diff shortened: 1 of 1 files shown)


gabrielwol (commented):
From my memory I think the centreline_intersection_point classifications are a bit inconsistent - you may instead want to filter out pseudo/highway instead by removing certain centreline types in the previous step? 
But if you've looked at the results and it's OK then this approach is fine too.

Nate-Wessel:
Not yet perfect, but I think a definite improvement. Here are some examples with the current view in red, this one in black

Screenshot 2025-05-28 095813

Screenshot 2025-05-28 095943

Not possible yet to distinguish between expressway ramps that connect with streets and those that don't:

Screenshot 2025-05-28 100201

Nate-Wessel:
@gabrielwol there's a version of this view in my schema if you want to take a look at the differences.

gabrielwol (approved):
Good progress. Results look much better!

PR #1225, opened 2025-05-30
Transform short-term TMCs into mid-block counts

Nate-Wessel:

- Transforms short-term TMCs into mid-block counts (entries to and exits from the intersection per approach)

## Issue(s) this solves:

- We have this transformation for Miovision, but not for short-term counts. It's useful! And a little tedious to type out.

- that all the turns are correctly identified with their corresponding entries/exits

- create the view in the `traffic` schema

### DIFF
--- a/volumes/short_term_counting_program/sql/create-view-tmc2atr.sql
+++ b/volumes/short_term_counting_program/sql/create-view-tmc2atr.sql
@@ -10,28 +10,37 @@ SELECT
     time_end,
     -- North leg traffic
     n_bike AS n_bike_sb,
-    NULL::int AS n_bike_nb,
+    CASE -- no where else for bikes to come from
+        WHEN w_bike + e_bike + s_bike = 0 THEN 0
+        -- but if there is a non-zero bike count, we can't say where the bike exits
+    END AS n_bike_nb,
     n_cars_l + n_cars_t + n_cars_r AS n_cars_sb,
     s_cars_t + e_cars_r + w_cars_l AS n_cars_nb,
     n_truck_l + n_truck_t + n_truck_r AS n_truck_sb,
     s_truck_t + e_truck_r + w_truck_l AS n_truck_nb,
     -- South leg traffic
     s_bike AS s_bike_nb,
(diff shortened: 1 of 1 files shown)


Nate-Wessel:
@gabrielwol I made some last minute changes but this is ready for review now

volumes/short_term_counting_program/sql/create-view-tmc2atr.sql
```diff
@@ -0,0 +1,137 @@
...
+        id,
+        count_id,
+        time_start,
+        time_end,
+        -- three-tier nested JSON object which will be unpacked in the next
+        -- step into a long-format view
+        json_build_object( -- classification level
+            'bike', json_build_object( -- approach level
+                'N', json_build_object( -- direction level
+                    'NB', CASE WHEN w_bike + e_bike + s_bike = 0 THEN 0 END,
```
gabrielwol, line 23: What is the logic of these CASE statements in the bike data?
NULL if there is data for other approaches, zero if not?

gabrielwol (commented):
Nice! You asked for name recommendations; `tmc_to_atr` fits our scheme better than `tmc2atr`. Or `tmc_atr_format`.

volumes/short_term_counting_program/sql/create-view-tmc2atr.sql
Nate-Wessel: It's explained in a comment up at the top. Basically, the only case where we actually know the exit counts at the moment is where it necessarily has to be zero. Otherwise we know bikes entered the intersection but can't say where they exited.

It's a bit pedantic at the moment but I also wanted to hold space for actual bike exit counts once full bike TMCs are available in the database.

Nate-Wessel:
I'm down with `tmc_to_atr` 👍

volumes/short_term_counting_program/sql/create-view-tmc2atr.sql
gabrielwol: Thanks for the explanation!

gabrielwol (approved):

Issue #1226, opened 2025-06-02
gcc puller mat view refresh timing


gabrielwol:
We refresh downstream mat views on trigger; since in our gcc pipeline inserts in batches, the mat view is being refreshed in between every batch. In addition to making the pull slower, this means the `_latest` mat views are only partially complete for each refresh until the final batch is pulled. 
 
We can transition to a new task to run the refreshes rather than using a trigger.

PR #1227, opened 2025-06-02
1226 gcc puller mat view refresh timing

gabrielwol:

- there were three duplicates in the latest centreline_intersection_point pull, so I added a `distinct on` in the mat view to prevent failure with the unique key.
- mat view refreshes were happening after each batch insert, so I moved them to a new taskgroup after pulling is complete.
- mat view refreshes also had potential to occur in wrong order; new method is an dictionary of tasks to run -> enforces order

## Issue(s) this solves:

- #1226

- 


gabrielwol:
Reverting to draft.. discovered one other issue with mat view refreshes. There is an ordering issue because centreline_latest is used in the other gis_core views:
`centreline_latest -> [centreline_intersection_point_latest, intersection_latest]`

The agg method won't work because we can't control order of mat view refreshes.

chmnata (approved):

PR #1261, opened 2025-08-29
#1260 Alter `vds.detector_inventory` from view to table

### LINKED ISSUE #1260: VDS Blue City sensors missing from detector_inventory CASE

gabrielwol:

- Alter `vds.detector_inventory` from view to table
  - means we can make one-off changes to detector inventory without a PR
  - Only new rows get inserted (ON CONFLICT DO NOTHING)
  - Should also improve speed (may need to add indices)
- To do: 
  - [ ] Airflow task (weekly?) to refresh centreline_latest geoms and delete ones which are no longer current

## Issue(s) this solves:

- #1260 

- Overall approach

--change all references from gwolofs to vds.

```
--drop all dependencies
DROP VIEW vds.detector_inventory;
ALTER TABLE gwolofs.detector_inventory SET SCHEMA vds;
--recreate all dependencies 
```


gabrielwol:
Idea from review sesh: Turn `detector_inventory` into a table, and add a ~trigger~ function that ~updates~ inserts new rows with these calculated fields.

Issue #1271, opened 2025-09-24
Update `gcc_layers_pull` to use functions instead of selects for mat view refreshes


gabrielwol:
..so mat views can be owned by `gis_admins` instead of `gcc_bot`.

PR #1281, opened 2025-11-06
#1271 GCC mat view refreshes to use functions

gabrielwol:

- Use functions (with `SECURITY DEFINER`) to refresh mat views, in order to allow `gcc_bot` to be removed from `gis_admins`.

## Issue(s) this solves:

- #1271

- 

- `Revoke gis_admins from gcc_bot` -> test to make sure gcc pipeline still runs.


chmnata:
should the bdit_dag_utils stuff be here?

PR #1285, opened 2025-11-13
Distinguish between Ecocounter flows that count either scooters or bikes

Nate-Wessel:

- Recognizes that some of the Ecocounter flows actually count scooters rather than bikes
- Allows us to distinguish between these (where possible, currently)

## Issue(s) this solves:

- We've been assuming all flows are bike counts

- veracity of statements

* None. Making the changes in the DB as I go.


gabrielwol (changes requested):
Going to request that the column is substantially filled out before this PR is approved, otherwise a column with mostly null values could be potentially misleading.
Also would like a `check` constraint on the new column to verify values (unless you want to have fun with a custom data type ;) ).

Nate-Wessel:
I believe this field has now been fully populated for validated flows.

PR #1333, opened 2026-01-19
#1292 updated flows at pape and millwood to represent the actual dire…

### LINKED ISSUE #1292: Eco-counter - Fill in missing site and flow information for new and existing counters
Update existing and newly installed Eco-counters with flow and site information. This is required to validate Eco-counter volumes against manual counts. Many flows have null `direction_main` and/or `includes_contraflow`:

```
SELECT * 
FROM ecocounter.sites_unfiltered AS s 
JOIN ecocounter.flows_unfiltered USING (site_id)
WHERE direction_main IS NULL
```

Several newly installed Eco-counters are also missing these fields:
- `sites_unfiltered`
  - `facility_description`
  - gis info (`centreline_id`, `linear_name_full`, `side_street`)
  - `technology`
  - `notes`
- `flows_unfiltered`
  - `flow_direction` (value of "unknown"), 
  - `direction_main`
  - `includes_contraflow`
  - `flow_geom`
  - `notes`

Updates to Docs: [readme for ecocounter.flows_unfiltered](https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/ecocounter#ecocounterflows_unfiltered) is missing some columns, most notably `direction_main` and `mode_counted`

wael6954:
the flow goems at millwood and pape have an incorrect directionality for which the southbound flows are drawn pointing towards the north

### DIFF
--- a/volumes/ecocounter/updates/flow_directionality_updates.sql
+++ b/volumes/ecocounter/updates/flow_directionality_updates.sql
@@ -1,0 +1,3 @@
+UPDATE ecocounter.flows
+SET flow_geom = ST_Reverse(flow_geom)
+WHERE flow_id IN (353534018, 353534020,353534008,353534010)


wael6954:
image

this is another instance of which I am unsure if I need to switch the flow geoms because it is possible that the flow measures for bikes going the wrong way.

Nate-Wessel (changes requested):
I think we've strayed from this a bit, but my original intention for the flow geoms is that they would have their first node over the actual sensor and the second node pointing in the direction of travel measured. Simply reversing these would swap the nodes and put the first node in the wrong location.

This is documented [here](https://github.com/CityofToronto/bdit_data-sources/tree/master/volumes/ecocounter#ecocounterflows_unfiltered).

IMO, if these flows are reversed, they should just be manually modified via QGIS's node tool.

gabrielwol (approved):

gabrielwol:
I reversed the flows that Wael found at Pape/Donlands that were pointing the wrong way. The lines are quite short so I don't think it's misleading to reverse rather than withdraw these 4 geoms. 

Merging is blocked by @Nate-Wessel 's change request.

Nate-Wessel (approved):

PR #1372, opened 2026-03-26
#1340 filter out all vehicles if lights absent

### LINKED ISSUE #1340: Miovision Anomalous Range - Partial vehicle outage
Shahrzad gave feedback that it can be very misleading when there is a Miovision `anomalous_range` for `classification_uid = 1` and not for other vehicular modes, as in the example below. This case requires you to first do a pre-screen to determine that all modes are available on each day before using it. 

Consider:
- filtering out all vehicle modes if one has an anomalous range
- not allowing `investigation_level = 'auto_flagged'` to be used with custom notes (made this one harder to find) 

(these 2 anomalous ranges have since been deleted after car levels seemed consistent with the following months)
```json
{
"leg": null, 
"uid": 43, 
"notes": "low car count - investigate", 
"range_end": "2020-10-05T00:00:00", 
"range_start": "2020-09-28T00:00:00", 
"problem_level": "questionable", 
"intersection_uid": 53, 
"classification_uid": 1, 
"investigation_level": "auto_flagged"
}

{
"leg": null, 
"uid": 83, 
"notes": "low car count - investigate", 
"range_end": "2020-10-05T00:00:00", 
"range_start": "2020-09-28T00:00:00", 
"problem_level": "questionable", 
"intersection_uid": 50, 
"classification_uid": 1, 
"investigation_level": "auto_flagged"
}
```

gabrielwol:

- adjust filtered views to filter out all vehicles if lights absent
- add constraint so auto-flagged ranges are not edited

## Issue(s) this solves:

- #1340 

- logic, consensus this is the right approach

- [x] update views


gabrielwol:
Testing:

```sql
CREATE OR REPLACE VIEW gwolofs.volumes_15min_atr_filtered AS

SELECT
    v15.intersection_uid,
    v15.datetime_bin,
    v15.classification_uid,
    v15.leg,
    v15.dir,
    v15.volume
FROM miovision_api.volumes_15min_atr_unfiltered AS v15
LEFT JOIN miovision_api.unacceptable_gaps USING (datetime_bin, intersection_uid)
WHERE
    unacceptable_gaps.datetime_bin IS NULL
    AND NOT EXISTS ( --anti join anomalous_ranges
        SELECT 1
        FROM miovision_api.anomalous_ranges AS ar
        WHERE
            ar.problem_level IN ('do-not-use', 'questionable')
            AND (
                ar.intersection_uid = v15.intersection_uid
                OR ar.intersection_uid IS NULL
            ) AND (
                ar.classification_uid = v15.classification_uid
                --don't use any vehicle modes if lights are anomalous
                OR (ar.classification_uid = 1 AND v15.classification_uid IN (1,3,4,5,9))
                --issue affects all modes
                OR ar.classification_uid IS NULL
            )
            -- leg is ignored here
            -- any anomalousness on a leg removes all ATR counts
            AND v15.datetime_bin >= ar.range_start
            AND (
                v15.datetime_bin <= ar.range_end
                OR ar.range_end IS NULL
            )
    );

--Total rows: 497062 Query complete 00:00:42.874
SELECT * FROM miovision_api.volumes_15min_atr_filtered
WHERE intersection_uid = 2
AND datetime_bin >= '2025-01-01'
AND datetime_bin < '2026-01-01'
AND classification_uid NOT IN (2, 6, 7, 10) --just vehicles

--Total rows: 487781 Query complete 00:00:35.216
SELECT * FROM gwolofs.volumes_15min_atr_filtered
WHERE intersection_uid = 2
AND datetime_bin >= '2025-01-01'
AND datetime_bin < '2026-01-01'
AND classification_uid NOT IN (2, 6, 7, 10) --just vehicles

--there are three anonalous ranges for cars in this period
SELECT * FROM miovision_api.anomalous_ranges WHERE
AND classification_uid = 1 AND intersection_uid = 2
uid IN (2945, 3857, 5976)
```

Nate-Wessel:
If we're getting no data for any vehicles (which we expect to exist), do we actually expect peds and cyclists to be counted correctly? Are there any real examples where that is the case?

gabrielwol:
>If we're getting no data for any vehicles (which we expect to exist), do we actually expect peds and cyclists to be counted correctly? Are there any real examples where that is the case?

I can't think of examples of that, but there are examples of vehicles being counted correctly and peds/bikes not being counted correctly (they are configured using separate detection zones) - so we definitely want to retain the ability to assign ARs to individual modes.

I feel like 
```sql
--don't use any vehicle modes if lights are anomalous
OR (ar.classification_uid = 1 AND v15.classification_uid IN (1,3,4,5,9))
```
And 
```sql
--don't use any modes if lights are anomalous
OR (ar.classification_uid = 1)
```
are pretty much synonymous - I could be open to either.

Nate-Wessel:
One possible real exception there is for street festivals, parades, ActiveTO, etc where vehicles are excluded (and may be counted as zero or near-zero) while the other modes proliferate. Currently I think these would have to be manually identified as `valid-irregular` anomalies, which I've done for a few.

A related but tangential idea: I've been wondering for a while how we could think about some meta-classifications like for "motor vehicles" or "active modes" etc. If we could somehow identify a broader classification category on the anomaly directly that would seem to be a better solution than a generalization over all anomalies.

E.g. this anomaly applies to "all active modes", which is taken to mean any variant of bikes + peds when the filter is applied.

aharpalaniTO (approved):
looks good, thanks!

PR #1435, opened 2026-06-23
#1431 Ecocounter API V2 Migration

### LINKED ISSUE #1431: Migrate to Eco-counter API v2
- in addition to changes required to pull reconstructed data.

gabrielwol:

- Swith to Ecocounter API v2
  - password instead of token authentication + more securely pass password callable around instead of text password (prevent from logging)
  - Instead of iterating through flows, need to iterate through sites now
    - Instead of daily attempt to backfill flow-level outages, now backfilling only site level outages. No evidence that single-flow outages occur. 
  - V2 is more strictly rate limited, add sleep on 429 status code
- Remove `counter` field, which had a non 1:1 relationship with sites under new api = difficult to integrate and not used.

## Issue(s) this solves:

- Closes #1431

- 

### DIFF
--- a/volumes/ecocounter/pull_data_from_api.py
+++ b/volumes/ecocounter/pull_data_from_api.py
@@ -1,5 +1,6 @@
 import requests
 import logging
+from typing import Callable
 from configparser import ConfigParser
 from psycopg2 import connect
 from psycopg2.extras import execute_values
@@ -14,7 +15,8 @@ default_end = datetime.now().replace(hour = 0, minute = 0, second = 0, microseco
 URL = 'https://api.eco-counter.com/api/v2/'
 
 # get a list of all sites from the API
-def getSites(pw: str, sites: any = ()):
+def getSites(password_getter: Callable[[], str], sites: any = ()):
+    pw = password_getter() # Retrieve only when needed
     response = requests.get(
         f'{URL}/sites',
         headers={'X-API-KEY': f'{pw}'}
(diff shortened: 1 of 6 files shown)


volumes/ecocounter/pull_data_from_api.py
```diff
@@ -104,95 +96,89 @@ def flowIsKnownToUs(flow_id: int, conn: any):
...
         %s::numeric,
         %s::text,
-        %s::text,
         ST_SetSRID(ST_MakePoint(%s, %s), 4326),
         null::boolean --not validated by default
     )
     """
     with conn.cursor() as cur:
-        cur.execute(insert_query, (site_id, site_name, counter, lon, lat))
+        cur.execute(insert_query, (site_id, site_name, lon, lat))
```
A-DUYVESTYN, line 132: Does the removal of `counter` mean we have to populate this identifier  manually for new sites?

A-DUYVESTYN (approved):
Looks good to my eye (which does not have much experience with airflow). One inline comment.

Please document this in the readme:
> - we can no longer query for specific flows, only sites. One implication of this is that instead of trying to backfill missing flow-level ranges, I am now looking for site-level outages.
> - ... never seen a single flow_id outage via that notification
> - We would still get notified through this weekly outage script: https://github.com/CityofToronto/bdit_data-sources/blob/master/volumes/ecocounter/data_checks/select-ongoing_outages.sql

volumes/ecocounter/pull_data_from_api.py
gabrielwol: To do: Found you can still query via /sites `include: counters`. Also realized this column needs to live in the `flows` table.

gabrielwol: Edit: decided to remove `counter` field from sites table. Not currently used and some ambiguity as to which is the correct counter at some sites.

PR #1469, opened 2026-08-17
#1276 Extend anomalous ranges to cover partial days before/after

### LINKED ISSUE #1276: Automated Miovision gap detection could roll forward, bacward to capture full gap
We currently flag, automatically, gaps in Miovision counts where we have zeros for a full day, at least for some modes. These `anomalous_ranges` are delimited by date boundaries, but the sensors almost never actually stop and start working precisely at midnight. 

This leaves partial-days of count data on either side of the automatically identified gaps where the run of zero-counts continue beyond the identified range. If we could just automatically extend the ranges in either direction until they encountered any non-zero count we'd be able to capture more of the problem data without being too overzealous.

Alternatively, we could be more conservative and exclude a full extra day on either side of these cases to ensure we fully capture the problem.

gabrielwol:

- This SQL function extends anomalous ranges both backwards and forwards based on available data, updating the start and end ranges accordingly.

## To do
- [ ] Implement into miovision_pull

## Issue(s) this solves:

- #1276 

- 

### DIFF
--- a/volumes/miovision/sql/function/function-extend_anomalous_ranges.sql
+++ b/volumes/miovision/sql/function/function-extend_anomalous_ranges.sql
@@ -1,0 +1,65 @@
+CREATE OR REPLACE FUNCTION miovision_api.extend_anomalous_ranges(ds date)
+RETURNS void AS $$
+
+WITH ranges_to_extend_backwards AS (
+    SELECT
+        ar.uid,
+        ar.range_start,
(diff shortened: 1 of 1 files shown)


DavidBoucaud (approved):
Approved without changes.

We could make fields to distinguish `range_start` as it is input against how it is extended, but  I see this as a non issue.

PR #1471, opened 2026-08-17
#1239 Use midblock model instead of centreline where appropriate

### LINKED ISSUE #1239: Adjust views to use new centreline2 (midblock_id / intersection_id) naming

gabrielwol:

- Use new midblock model in place of centreline where relevant.

## Issue(s) this solves:

- Closes #1239

- Within the `traffic` schema I think this one view was the extent of changes necessary, but it would be worth discussing other dependents of `gis_core.centreline_*` objects to see if they would be better off pointing to midblock model.

### DIFF
--- a/volumes/short_term_counting_program/sql/create-view-svc_daily_totals.sql
+++ b/volumes/short_term_counting_program/sql/create-view-svc_daily_totals.sql
@@ -4,17 +4,17 @@ SELECT
     suv.study_id,
     suv.count_date,
     suv.direction,
-    meta.centreline_id,
+    meta.midblock_id,
     cl.geom AS centreline_geom,
     SUM(suv.volume) AS daily_volume
 FROM traffic.svc_unified_volumes AS suv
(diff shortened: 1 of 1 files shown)


volumes/short_term_counting_program/sql/create-view-svc_daily_totals.sql
```diff
@@ -4,23 +4,24 @@ SELECT
     suv.study_id,
     suv.count_date,
     suv.direction,
-    meta.centreline_id,
+    meta.midblock_id,
     cl.geom AS centreline_geom,
```
chmnata: add the centreline_id_array[] as well.

chmnata: midblock_geom

chmnata (changes requested):
great change!

chmnata (approved):

Issue #1492, opened 2026-09-14
Investigate and fix miovision_pull.pull_data.pull_alerts (scheduled__2026-09-14T07:00:00+00:00) FAILED


DavidBoucaud:
- Recreate error and check response

Issue #1449, opened 2026-07-21
Miovision v1 API deprecation


gabrielwol:
Discovered at https://docs.api.miovision.one/#/

> This is the v2 replacement for GET /v1/intersections/{intersectionId}/tmc/crosswalk. The only difference is that response timestamps are returned in the Organization's local timezone (matching the Miovision One dashboard) rather than UTC.

> This is the v2 replacement for GET /v1/intersections/{intersectionId}/tmc. The only difference is that response timestamps are returned in the Organization's local timezone (matching the Miovision One dashboard) rather than UTC.

Note: need to use a mix of v1/v2 endpoints.

DavidBoucaud:
1. Map v2 replacements
2. Check/compare outputs
3. Plan implementation

DavidBoucaud:
1. Map v2 replacements.

The v2 APIs only occur in one script (volumes.miovision.api.intersection_tmc):

API calls | scripts | v2
- /intersections | volumes.miovision.api.intersection_tmc | no
- **/intersections/{intersection_id}/tmc | volumes.miovision.api.intersection_tmc | yes**
- **/intersections/{intersection_id}/tmc/crosswalk/ | volumes.miovision.api.intersection_tmc | yes**
- /alerts | volumes.miovision.api.pull_alert | no
- /intersections/{intersection_id}/cameras | volumes.miovision.api.configuration_info.py | no
- /intersections/{intersection_id}/hardware/detectionConfiguration | volumes.miovision.api.configuration_info.py | no

2. Yep, its only the timestamp return which changes. We now receive local time instead of UTC
3.  We need to update the `process_timestamp` script, and switch from v1 to v2.

New process_timestamp:

```
def process_timestamp(row):
    """Parse and remove TZ information from local timestamp"""
    local_timestamp  = row['timestamp']
    if local_timestamp  is None:
        return None
    local_timestamp = dateutil.parser.parse(str(local_timestamp))
    # Remove TZ info and return
    return local_timestamp.replace(tzinfo=None)
```

This change is nonurgent and will be addressed by [PR: #1494](https://github.com/CityofToronto/bdit_data-sources/pull/1494)

PR #1493, opened 2026-09-15
Fix pull_alerts api call failure

DavidBoucaud:

- The api call for alerts was breaking because the url returned from miovisions 'next_page' was not functional
- We now return the next page number from miovision, and otherwise maintain our original url.

## Issue(s) this solves:

-  #1492 

- Check the API call and while loop logic

- move change to 1409-miovision-cordons branch, pull to Gabes directory, point the symlink of pull_miovision towards Gabe's directory


radumas (approved):
Looks good as a fix. 

I wonder if ultimately the page handling logic has broader application across the various miovision APIs

I am noting some repeat code between this file and intersection_tmc, including the class structure and am wondering if, maybe as part of https://github.com/CityofToronto/bdit_data-sources/issues/1449 we think about DRYing this code and having a parent class for Miovision API handling.

